PR #1, opened 2026-04-26
Gracefully handle missing MTP tensors in index

JDWarner:
Handle cases where no MTP tensors are found gracefully.

Some Qwen3.5/3.6 finetunes lack MTP, and on such models PrismaQuant reaches this point then the error raised causes a hard stop. This trivial patch allows the pipeline to continue. I tested this two-line patch with https://huggingface.co/AEON-7/Qwen3.6-27B-AEON-Ultimate-Uncensored (which lacks MTP) and it succeeded, passing on 0 MTP tensors.

### DIFF
--- a/prismaquant/incremental_probe.py
+++ b/prismaquant/incremental_probe.py
@@ -1842,6 +1842,32 @@ def _run_mtp_streaming_shard(
     mtp_wrapper.eval()
 
     raw = _load_mtp_state_dict(model_path)
+    if not raw:
+        # No MTP weights in source — write empty pickle to satisfy the
+        # schedule and return. Mirrors the text-only visual fallback.
+        Path(output_path).parent.mkdir(parents=True, exist_ok=True)
+        with open(output_path, "wb") as f:
+            pickle.dump({
+                "stats": {},
+                "router_counts": {},
+                "router_totals": {},
+                "expert_info": {},
+                "expert_saliency": {},
+                "meta": {
+                    "model": model_path,
+                    "dataset": dataset_name,
+                    "nsamples": int(calib.size(0)),
+                    "seqlen": seqlen,
+                    "dtype": dtype_name,
+                    "execution_device": str(device),
+                    "linear_include": linear_include,
+                    "linear_exclude": linear_exclude,
+                    "skipped_reason": "no MTP weights in source",
+                },
+            }, f)
+        print(f"[incremental/mtp] no MTP weights; wrote empty shard "
+              f"pickle to {output_path}", flush=True)
+        return
     missing, extra = _load_into_mtp(inner_mtp, raw)
     loaded = len(raw) - len(missing)
     print(f"[incremental/mtp] loaded {loaded}/{len(raw)} mtp weights "
--- a/prismaquant/incremental_probe.py
+++ b/prismaquant/incremental_probe.py
@@ -290,14 +290,25 @@ def build_extended_shard_regexes(
             n_body, layers_per_shard, layer_prefix="model.layers"))
 
     if include_mtp:
-        n_mtp = int(
+        n_mtp_config = int(
             text_cfg.get("num_nextn_predict_layers")
             or cfg.get("num_nextn_predict_layers")
             or text_cfg.get("num_mtp_layers")
             or cfg.get("num_mtp_layers")
-            or _count_mtp_layers_from_safetensors(model_path)
             or 0
         )
+        n_mtp_actual = _count_mtp_layers_from_safetensors(model_path)
+        # Empirical safetensors count is ground truth: a config may
+        # declare MTP layers (inherited from a base) when the finetune
+        # actually stripped the weights. Take the min so we never
+        # schedule MTP shards the runner can't fulfill.
+        n_mtp = min(n_mtp_config, n_mtp_actual) if n_mtp_actual > 0 else 0
+        if n_mtp_config > 0 and n_mtp_actual == 0:
+            print(f"[shard-schedule] config declares "
+                  f"{n_mtp_config} MTP layer(s) but safetensors index "
+                  f"has no `mtp.*` keys; skipping MTP shards "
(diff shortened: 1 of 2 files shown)


RobTand:
Thanks for the patch — confirmed the failure mode and reproduced the path.

The PR fixes the symptom cleanly. Worth flagging for context (and in case you want to expand the patch): there's a broader root cause one layer up. The MTP shard scheduler at `incremental_probe.py:372-379` uses an `or` chain that prefers the config-declared MTP count over an empirical check of the safetensors:

```python
n_mtp = int(
    text_cfg.get(\"num_nextn_predict_layers\")    # base-config says \"1\"
    or cfg.get(\"num_nextn_predict_layers\")
    or text_cfg.get(\"num_mtp_layers\")
    or cfg.get(\"num_mtp_layers\")
    or _count_mtp_layers_from_safetensors(model_path)  # never reached
    or 0
)
```

So Qwen3.6 finetunes that inherit `num_nextn_predict_layers=1` from the base but strip the MTP weights at finetune time end up scheduling an MTP shard the runner can't fulfill — that's exactly your hit.

Your patch lets `_load_mtp_state_dict` return empty, which avoids the hard error. The thing it doesn't catch: `_run_mtp_streaming_shard` then runs forward+backward on the random-init MTP weights and writes a Fisher shard pickle with effectively-garbage stats. Harmless if nothing reads it (the merged probe still works), but technically incorrect.

Three layers of defense that would close the loop end-to-end:

1. **Schedule-level (root cause)**: `n_mtp = min(config_value, safetensors_count)` so a config that lies about MTP doesn't propagate. Keeps your `or` fallback for cases where safetensors don't yet exist.
2. **Loader (your patch)**: keep `_load_mtp_state_dict` permissive — return `{}` with a warning instead of `RuntimeError`. Defense-in-depth.
3. **Runner**: detect empty `raw` in `_run_mtp_streaming_shard`, write an empty pickle and return early (mirrors the visual-shard fallback at `incremental_probe.py:3036`).

~15 LoC across the three sites if you want to expand the PR. Otherwise happy to merge this as-is and file the broader fix as a follow-up — your call. Either way, thanks for the report and the test against AEON-Ultimate-Uncensored.

RobTand:
Concrete diff for the three-layer fix, in case it's useful:

### Layer 1 — `incremental_probe.py:371-382` (schedule-level root cause)

```diff
     if include_mtp:
-        n_mtp = int(
+        n_mtp_config = int(
             text_cfg.get("num_nextn_predict_layers")
             or cfg.get("num_nextn_predict_layers")
             or text_cfg.get("num_mtp_layers")
             or cfg.get("num_mtp_layers")
-            or _count_mtp_layers_from_safetensors(model_path)
             or 0
         )
+        n_mtp_actual = _count_mtp_layers_from_safetensors(model_path)
+        # Empirical safetensors count is ground truth: a config may
+        # declare MTP layers (inherited from a base) when the finetune
+        # actually stripped the weights. Take the min so we never
+        # schedule MTP shards the runner can't fulfill.
+        n_mtp = min(n_mtp_config, n_mtp_actual) if n_mtp_actual > 0 else 0
+        if n_mtp_config > 0 and n_mtp_actual == 0:
+            print(f"[shard-schedule] config declares "
+                  f"{n_mtp_config} MTP layer(s) but safetensors index "
+                  f"has no `mtp.*` keys; skipping MTP shards "
+                  f"(common on Qwen3.5/3.6 finetunes that strip MTP)",
+                  flush=True)
         if n_mtp > 0:
             regexes.extend(build_layer_shard_regexes(
                 n_mtp, layers_per_shard, layer_prefix="mtp.layers"))
```

### Layer 2 — `mtp_module.py:131-135` (your patch, slightly extended)

```diff
     mtp_files = sorted({wm[k] for k in wm if k.startswith("mtp.")})
     if not mtp_files:
-        raise RuntimeError("no mtp.* weights in safetensors index")
+        # Some Qwen3.5/3.6 finetunes inherit num_nextn_predict_layers
+        # from the base config but strip the MTP weights. Return empty
+        # rather than raising — the runner detects empty and writes an
+        # empty shard pickle. See PR #1.
+        print("[mtp] no mtp.* weights in safetensors index; "
+              "returning empty state dict", flush=True)
+        return {}
```

### Layer 3 — `incremental_probe.py:_run_mtp_streaming_shard` after line 2387

Add an early return on empty raw, mirroring the visual fallback at `:3036`:

```diff
     raw = _load_mtp_state_dict(model_path)
+    if not raw:
+        # No MTP weights in source — write empty pickle to satisfy the
+        # schedule and return. Mirrors the text-only visual fallback.
+        Path(output_path).parent.mkdir(parents=True, exist_ok=True)
+        with open(output_path, "wb") as f:
+            pickle.dump({
+                "stats": {},
+                "router_counts": {},
+                "router_totals": {},
+                "expert_info": {},
+                "expert_saliency": {},
+                "meta": {
+                    "model": model_path,
+                    "dataset": dataset_name,
+                    "nsamples": int(calib.size(0)),
+                    "seqlen": seqlen,
+                    "dtype": dtype_name,
+                    "execution_device": str(device),
+                    "linear_include": linear_include,
+                    "linear_exclude": linear_exclude,
+                    "skipped_reason": "no MTP weights in source",
+                },
+            }, f)
+        print(f"[incremental/mtp] no MTP weights; wrote empty shard "
+              f"pickle to {output_path}", flush=True)
+        return
     missing, extra = _load_into_mtp(inner_mtp, raw)
```

The schema validator already accepts empty `stats` / `expert_saliency` (it's the same shape as the visual fallback), so the merged probe will load cleanly without an MTP shard.

If you'd rather keep the PR minimal and have these land separately, totally fine — say the word and I'll merge as-is and file a follow-up issue.

JDWarner:
Sorry I have been busy at a conference. Thanks for the breadcrumbs! I believe the complete 3-layer fix is now included.

PR #13, opened 2026-07-17
DSv4-Flash probe enablement

tylerwagler:
**Commits:** `layer_streaming: dequant DSv4-Flash MXFP4 routed experts` ·
`layer_streaming: DSv4 rope-axis dict falls back to 'main' in _call_layer` ·
`layer_streaming: CSA layers take the sliding mask in DSv4 probe mode` ·
`incremental_probe: stream phase-1 activations to host per layer` ·
`cost: measure DSv4 per-expert Linears; stratified dense-path expert subsample`

### Summary
Five self-contained fixes that let the incremental probe and the cost stage
run on DeepSeek-V4-Flash (per-expert-Linear MoE, MXFP4-packed routed experts,
compressed-sparse-attention layer types):

1. **MXFP4 dequant in layer streaming.** The Flash checkpoint stores routed
   experts as OCP MX FP4 (E2M1 nibble pairs packed int8, per-row E8M0 scales
   over 32-element groups), not block-FP8; the fp8_scale_pairs sweep routed
   them into the FP8 path where the scale-grid guard (correctly) refused
   them. `_apply_fp8_dequant_inplace` now detects the
   int8+E8M0(out, packed_in/16) signature and dequants on a dedicated
   vectorized path, validated bit-exact against an independent scalar
   reference decode on real checkpoint tensors.
2. **Rope-axis fallback.** DSv4's rotary exposes
   `layer_types=("main","compress")` — rope axes, not attention-schedule
   types — so the Gemma-style per-layer-type selection finds no entry for a
   decoder layer type like `sliding_attention`. `_call_layer` mirrors the
   vendored probe forward: `"main"` rope for every layer.
3. **CSA mask aliasing.** DSv4's `compressed_sparse_attention` /
   `heavily_compressed_attention` layer types run sliding-window-only
   attention in probe mode; `_compute_attention_mask` aliases exactly those
   two types to the sliding mask. Any other unknown type still fails loudly.
4. **Per-layer host capture of phase-1 activations.** This deliberately
   reverts the v22 "Fix E1" batched device→host transfer: stacking all L+1
   device-resident activations before one `.cpu()` doubles the peak device
   memory of the activation working set at the exact phase-1/2 transition
   where the probe's high-water mark already sits. The per-layer copy is off
   the hot path (the layer forward dominates). The list now holds host
   tensors end to end (renamed `device_acts` → `host_acts`, including the
   embedding output at index 0 which E1's stack used to transfer).
5. **Per-expert cost rows + dense-path expert subsample.**
   `canonical_linear_name()` remaps DSv4 per-expert names to packed-style
   names that are not in `target_names` (the probe keys them RAW), so all
   three cost loops silently skipped every routed expert. A shared helper
   (`resolve_cost_target_name`) falls back to the raw name. And because
   256 experts × 3 projections × 43 layers through the exhaustive IQ grid
   one tensor at a time is not tractable, the packed-path
   `PRISMAQUANT_EXPERT_COST_SAMPLE` lever is ported to the dense path:
   measure a deterministic evenly spaced expert sample per
   (experts-prefix, projection) group, extrapolate the rest with the
   sampled mean so every expert keeps a cost row (the allocator drops
   row-less names, which would shrink the DP's accounting and serving-unit
   membership). Export is unaffected — it always quantizes every expert.

Nothing on non-DSv4 models: the MXFP4 path triggers only on its tensor
signature, the rope fallback only when `"main"` is a rope-axis key, the mask
aliasing only for the two DSv4 layer types, and the subsample only when
`PRISMAQUANT_EXPERT_COST_SAMPLE` is set. The E1 revert applies to all models:
phase-1 device residency drops from L+1 activations to 1, in exchange for
per-layer (off-hot-path) copies.

### How to verify
- `pytest tests/test_layer_streaming.py tests/test_layer_streaming_fp8.py
  tests/test_dsv4_layer_streaming_rename.py tests/test_incremental_probe.py
  tests/test_expert_cost_subsample.py -q`
- MXFP4 dequant can be spot-checked against any MXFP4-packed checkpoint by
  comparing the streamed bf16 weights to a scalar nibble+E8M0 decode.

### Tests
New: `tests/test_expert_cost_subsample.py` (deterministic evenly spaced
sample, per-projection grouping, non-expert rows untouched, small groups
exempt; extrapolation = sampled mean per format, error entries excluded,
unmeasured-output marker propagated). Existing layer-streaming and probe
suites cover the rest.

https://claude.ai/code/session_01QXh7ricvRaq4pMg4RKPDf5


RobTand:
Reviewed, and tested against the new CI (#18). Thank you for this — the reasoning quality across the whole stack is high, and the tests assert contracts rather than restate implementations, which is not the norm.

**Verdict: needs changes — one blocking, the rest follow-ups.**

**Blocking — 4× cache-size undercount for `I8`.** `streaming_model.py:236` and `autoscale.py:94-106`: `_safetensors_cache_dtype_bytes("I8", bf16) → 1` and `_safetensors_source_float_bytes("I8") → None`. After this PR the resident tensor is **4 bytes per on-disk `I8` byte** (1 packed byte → 2 logical elements × 2 B bf16). `F8` is handled correctly (1→2); `I8` was inert before, and this PR is what makes it a lie.

On DSv4 MoE layers — where nearly all the bytes live — `estimated_layer_bytes` comes out 4× small, so `prepare_for_load()` under-evicts, the post-load eviction loop has to catch up *after* the layer has landed, and prefetch (`put(..., force=False)`) **silently refuses** any layer over `effective_max` and discards the dequant. That is the thrash mode, on a 121 GB unified pool. `LayerCache.total_bytes` stays accurate — only the two pre-load estimators are wrong. One line each.

**Also worth addressing before merge:**

- **E8M0 `0xFF` must be NaN**, not `+inf` (`layer_streaming.py:576`). A `0xFF` block currently yields 29 `inf` + 3 `NaN` instead of 32 `NaN`.
- **MXFP4 detection is a shape heuristic** (`:506-510`): any 2-D int8 tensor whose scale grid satisfies `scale.shape[1]*16 == w.shape[1]`. An INT8-with-group-16-scales checkpoint false-positives and gets decoded as nibble pairs — silent corruption. The checkpoint declares its quant method in `config.json`; house rule is no heuristic where an explicit exists. Related: the `continue` at `:511` bypasses `_check_fp8_scale_grid`, the transposed-grid assertion the 2026-07-02 audit added.
- **The bit-exactness claim has no CI coverage.** I reproduced it — an independent scalar reference matches the vectorized path exactly on synthetic tensors — but nothing in `tests/` constructs an MXFP4 tensor. Worth pinning, since it is CPU-only and cheap.
- **`:568` loops per tensor** (~768 expert tensors/layer), which is what the function's own docstring at `:466-471` says the batched design exists to avoid.
- **The v22 per-layer `.cpu()` change** (`incremental_probe.py:1235-1241`) reverts a documented fix for *every* model with no measurement and no env gate. The memory win is real and probably necessary; the perf claim is unmeasured, and the `t_h2h` log now always prints `0.0s` because the timer starts after the transfers.
- **The subsample is wired into `measure_unbatched`/`measure_batched_gpu` only** — `_measure_production_render_dense` got the `resolve_cost_target_name` fix but not the subsample, so under the default `COST_MODE=production-render-score` the lever does nothing.

**What I could not verify:** the nibble-order convention and the dequant against the real DSv4-Flash checkpoint, which needs the 284 GB source. A single-layer decode compared against the reference `transformers` forward would settle both.

**Done well:** `resolve_cost_target_name` is strictly behaviour-preserving. The subsample is genuinely deterministic (sort → linspace → unique, no RNG), and `tests/test_expert_cost_subsample.py` pins the exact sampled ids rather than just the count. The mean-fill is also the *right* imputation given union-find forces per-layer expert uniformity anyway.

tylerwagler:
Implemented everything, one focused commit per item (pushed on top of the reviewed head, no rebase). All seven claims verified against the code first — none disputed.

- **I8 4× undercount** — `d8f073a`. Both pre-load estimators now price nibble-packed I8 experts at 2 logical elements × execution dtype per disk byte. Rather than unconditional one-liners I gated them on the checkpoint's explicit `expert_dtype: "fp4"` declaration (same gate as the decode path), so a genuine int8 tensor still sizes verbatim. Tests pin 4× declared / 1× undeclared.
- **MXFP4 detection** — `3081a40`. Decode now triggers only on the config declaration, carried on `Fp8ScaleInvMap.mxfp4_names`; the packed grid is a hard assertion (`_check_mxfp4_packed_grid`), never the trigger, and non-MXFP4 2-D tensors flow through `_check_fp8_scale_grid` again — your INT8-with-group-16-scales false positive now raises loudly (test included).
- **E8M0 `0xFF` → NaN** — `aa89964`, with a test pinning a `0xFF` block → 32 NaN, zero inf.
- **Bit-exactness CI coverage** — `tests/test_layer_streaming_mxfp4.py` (10 CPU-only tests): loop-based scalar reference that shares no code with the vectorized path, compared bitwise (`view(torch.int16)`), all 16 FP4 codes, NaN positions, synthetic DSv4-style checkpoints built through the real `fp8_scale_pairs`/profile path.
- **Per-tensor loop** — `f945722`. Step 3b batches per shape-group like step 3 via a (256,2) byte→element-pair LUT, chunked at 32 tensors/launch to bound the long-index gather transient (~768 → ~24 launches/layer); chunk-boundary test stays bit-exact vs the reference.
- **v22 `.cpu()` revert** — `10654ce`. Kept the per-layer default (the device high-water saving is the load-bearing claim; the comment now says plainly the transfer-time tradeoff is unmeasured), `PRISMAQUANT_PROBE_BATCHED_ACT_TRANSFER=1` restores Fix E1 (read once per probe run), and `host transfer` now times the actual copies on both paths. Flag documented in `docs/runtime_flags.md`.
- **Dense-path subsample** — `79964de`. `_measure_production_render_dense` applies the same split/extrapolate pair as the other two sites, so the lever works under the default `COST_MODE=production-render-score`; extrapolated rows carry production provenance and the log reports measured/extrapolated counts.

On what you couldn't verify: agreed the single-layer decode against the reference `transformers` forward is the right settle for the nibble-order convention — that needs the box with the 284 GB source, and it's queued as the next lane task before any DSv4 measurement gets trusted.

Full CPU suite: 964 passed; the 6 remaining failures reproduce byte-for-byte at the PR base and are the ones your `1cc7b90` fixed on main — they'll clear on merge/rebase, so CI here will show them until then.

RobTand:
**Merged.** All seven items verified in the code, not just in the prose — and the two structural ones are right: the MXFP4 trigger is declaration-driven at its only site, and I confirmed the estimator↔`LayerCache.total_bytes` parity end-to-end on the real name spaces (both the DSv4 checkpoint spelling and the live per-expert spelling), not only through the tests.

**I landed three follow-on fixes on top rather than filing them,** because two partly re-opened the memory-budget hole item 1 closed (see `4bf27ba`):

1. **U8 packed experts were still undercounted 4×.** Both estimators gated on `dtype_name == "I8"` while the decode accepts int8 *and* uint8 (`layer_streaming.py:386`). Measured: a declared-fp4 U8 tensor decodes to 512 resident bytes while `_estimate_layer_cache_bytes` returned 128. Since U8 is the common OCP/GPT-OSS nibble-pack spelling, that is the same `prepare_for_load` under-eviction and `put(force=False)` refusal you fixed for I8. Now one shared `_PACKED_BYTE_DTYPES`.
2. **The scale plane was unvalidated.** `_check_mxfp4_packed_grid` checked the weight dtype but not the `.scale` sibling that step 3b reinterprets as E8M0 via `exp2(byte-127)`. A width check wouldn't help — `float8_e4m3fn` is also one byte — so it's an explicit dtype allow-list, plus validation of `quantization_config.scale_fmt` when declared (absent stays non-fatal, since real checkpoints ship `expert_dtype` with no scale-format field).
3. **The decode chunk's transient was invisible to the cache budget** (`prepare_for_load` reserves only the resident layer size). int32 gather indices plus an in-place fp32 scale multiply: **−47% alloc volume, −38% peak live set** per packed byte, bit-exact by `int16`-view equality over the whole chunk, and your chunk-boundary test passes untouched.

Also fixed `declared_fp4_expert_dtype` raising `AttributeError` on a config whose root or `text_config` isn't a dict, against its own "False on anything unexpected" contract, and corrected the two runtime-flag doc entries (`PRISMAQUANT_PROBE_BATCHED_ACT_TRANSFER` had no real row; `PRISMAQUANT_EXPERT_COST_SAMPLE` still said "stacked-expert (3D)" after it started driving the per-expert 2-D dense paths).

**Two accuracy notes, in the house spirit of not overselling a screen.** The new MXFP4 test file is **8 tests, not 10**, and with scale exponents 100–200 every product is exactly representable in bf16 — so it pins nibble order, grouping and NaN handling, but not rounding/overflow numerics. And the whole decode now hinges on a config key (`expert_dtype: "fp4"`) that has no fixture and no real-config evidence anywhere in the tree, with DSv4's `mlp.shared_experts.*` structurally excluded from `mxfp4_names` by the `\.experts\.\d+\.` pattern. The failure mode if the real key differs is a loud `ValueError`, not silent corruption, so this isn't blocking — but it should be confirmed against the actual checkpoint header before the first DSv4 load is trusted. Filed as a follow-up.

One live non-DSv4 behaviour change your notes don't call out: the dense-path subsample newly activates for anyone with the legacy `PRISMAQUANT_GGUF_EXPERT_COST_SAMPLE` set under the default `COST_MODE=production-render-score`, so GGUF-lane cost estimates change. Intended, but worth stating.

PR #14, opened 2026-07-17
Fisher: normalize h_trace by the global calib token count

tylerwagler:
> Stacked on #13 — only the last 3 commits are new here (1 original + review round 1); I'll rebase as predecessors merge.

**Commit:** `probe/alloc: normalize Fisher h_trace by the global calib token count`

### Summary
The incremental probe finalized `h_trace = h_trace_raw / n_tokens_seen` PER
ROW. For per-expert-`nn.Linear` layouts, `n_tokens_seen` only counts the
tokens ROUTED to that expert — inflating a rarely-routed expert's Fisher by
(global/routed), i.e. exactly inverted importance weighting: the least-used
experts look the most sensitive, and the allocator spends bits on them while
attention/shared rows starve at the floor. Typical inflation is
~n_experts/top_k (≈32x on a 256-expert top-8 model); ~33,000x only in the
degenerate case of an expert routed 1 token of a 32,768-token calibration.

Scope correction from review: DSv4-Flash itself probes packed-3D under the
current toolchain (transformers 5.6.0 instantiates packed `*Experts`
containers), so this change is near-inert on today's shipped artifact
classes. It bites where experts genuinely probe as per-expert `nn.Linear`:
older transformers pins (MiniMax pins 4.57.5), vendored modeling files, and
`--model-override` runs on older images.

Tokens never routed to an expert contribute zero gradient, so the empirical
Fisher over the calibration set divides by the GLOBAL token count
(nsamples × seqlen) for every row (`finalize_fisher_stats`). Honest scope:

- **Dense trunk rows: unchanged** (they accumulate exactly the global count).
- **Packed-3D expert entries: unchanged** (their `n_tokens_seen` already
  incremented by the full batch token count).
- **Per-expert-Linear rows: fixed** (the headline case).
- **Batched MoE-block flush rows: fixed as a bonus** — they never incremented
  `n_tokens_seen`, so the old finalize divided them by 1.

Existing probe pickles are corrected at allocator load time from the stored
raw accumulators (`h_trace_raw`, `h_w2_sum_raw`, `h_trace_per_expert_raw`,
meta nsamples/seqlen) without re-probing; the recompute is idempotent for
probes written by the fixed finalize (stamped `meta.fisher_norm_tokens`).

### Review round 1
- `95b3993` extends the same global denominator to the h-detail blobs and
  the sensitivity backend (the review's mixed-units relocation): all writers
  funnel `global_tokens`, blobs carry `h_detail_version: 4` + `norm_tokens`,
  every pickle stamps `meta.fisher_norm_tokens`. It also reverses the
  documented M4 convention ON THE RECORD — the M4 comment block,
  `finalize_fisher_stats`'s docstring, CLAUDE.md §3, and
  `docs/prismaquant_design.md` §2.2 updated in the same commit.
- `612fc38` hard-fails probes that carry raw accumulators but no token meta
  (`renormalize_probe_fisher()`); `--allow-legacy-fisher-norm` restores the
  warning path. Detection widened to the flush-path rows
  (`n_tokens_seen == 0`) the old `> 0` gate missed.

Allocations on per-expert-Linear MoE models can change substantially (that
is the fix). Dense-model probes and packed-3D probes produce identical
h_trace values before and after. Probes lacking nsamples/seqlen meta now
hard-fail at load (`--allow-legacy-fisher-norm` opts back into the warning).

### How to verify
`pytest tests/test_fisher_normalization.py -v` — the synthetic reproducer:
two identical Linears with identical gradient mass over the same calibration
run (the "dense" one sees every token, most carrying zero gradient; the
"expert" one sees only its routed tokens). Equal h_trace after the fix; the
old per-row normalization, computed inline for contrast, inflates the expert
by exactly (global/routed).

### Tests
`tests/test_fisher_normalization.py` now 8 tests: the reproducer, the
divided-by-1 flush-path row, scalar/detail denominator agreement
(`sum(h_diag) == h_trace` at writer level and end-to-end), renorm-from-meta,
idempotence, hard fail, escape hatch, no-raw-rows no-op.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -104,8 +104,15 @@ Under Tradeoff; an earlier internal name for the L3-polish piece was
 ### The cost cascade (cheap→faithful, each level gates the next)
 - **L1 — additive Fisher.** `predicted_dloss = ½·H_trace·MSE` per `(Linear,
   format)`, from the diagonal-Fisher 2nd-order loss expansion. `H_trace` is the
-  empirical Fisher diagonal trace (one calibration backward pass; MoE experts
-  divided by routing probability). Solve the multi-choice knapsack DP. Seconds.
+  empirical Fisher diagonal trace (one calibration backward pass; **every row —
+  dense or MoE expert — is normalized by the same global calib token count**,
+  since tokens never routed to an expert contribute zero gradient to the
+  mean-Δloss objective). Superseded conventions, both reversed on purpose: an
+  explicit ÷route_prob (removed in audit M4) and the per-routed-token division
+  that M4 kept as "the implicit ÷token-fraction" (removed in PR #14 — it
+  inflated rarely-routed unpacked-expert rows by global/routed, i.e. inverted
+  importance weighting; see `finalize_fisher_stats`). Solve the multi-choice
+  knapsack DP. Seconds.
 - **L2 — perturbed-X fixed point.** Re-measure each Linear's *output* MSE under
   the activation distribution induced by the current assignment (upstream quant
   noise shifts downstream inputs), re-solve, iterate to weighted-Hamming
--- a/docs/prismaquant_design.md
+++ b/docs/prismaquant_design.md
@@ -142,15 +142,19 @@ and (optionally) a per-row/per-output-channel diagonal `h_full`.
 
 For MoE packed experts, the same principle applies — squared
 gradient per weight — but the storage layout is `[E, M, N]` instead
-of `[out, in]`. The implementation in
-`sensitivity_probe._GradNormCapture` (line 413) chunks along the
-expert dimension to bound peak memory and uses the
-`RouterTracker` (line 920+) to **divide each expert's Fisher by its
-observed routing probability**. Without that scaling, sparse experts
-look artificially less sensitive than densely routed ones. With it,
-expert Fisher values are calibrated as "loss change *per actually
-used token*" and become comparable across the expert bank and across
-dense Linears.
+of `[out, in]`. Every accumulator — dense trunk, packed expert, or
+per-expert `nn.Linear` — is normalized by the **same global
+calibration token count** (`finalize_fisher_stats`): tokens never
+routed to an expert contribute zero gradient, so under the mean-Δloss
+objective the empirical Fisher divides by the global count for every
+row, and expert values are directly comparable across the expert bank
+and against dense Linears. Two earlier conventions were deliberately
+reversed: an explicit ÷route_prob via the `RouterTracker` (removed in
+audit M4; `route_prob` is metadata only now), and the per-routed-token
+division M4 retained as the "implicit ÷token-fraction" (removed in
+PR #14 — it is the same 1/p_e scaling, and it *inflated* rarely-routed
+unpacked-expert rows by (global/routed), inverting importance
+weighting in the knapsack).
 
 ### 2.3 Why empirical Fisher and not true Hessian
 
--- a/prismaquant/allocator.py
+++ b/prismaquant/allocator.py
@@ -1268,6 +1268,50 @@ def main():
     costs = cost_data["costs"]
     print(f"[alloc] stats: {len(stats)} Linears, costs: {len(costs)} Linears")
 
+    # ---- Fisher renormalization ----
+    # Older incremental probes finalized `h_trace = h_trace_raw /
+    # n_tokens_seen` PER ROW, where per-expert Linears only count their
+    # ROUTED tokens — inflating a rarely-routed expert's Fisher by
+    # (global/routed), up to ~33,000x, i.e. inverted importance weighting.
+    # The raw accumulators are stored, so recompute every derived field here
+    # with the one correct shared denominator: the global calib token count
+    # (probe meta nsamples x seqlen). Dense rows saw exactly that many
+    # tokens, so their values are unchanged; probes written by the fixed
+    # finalize (meta.fisher_norm_tokens) renormalize to identical values —
+    # the recompute is idempotent.
+    _meta = probe.get("meta", {}) or {}
+    _norm_tokens = int(_meta.get("fisher_norm_tokens", 0) or 0)
+    if _norm_tokens <= 0:
+        _norm_tokens = (int(_meta.get("nsamples", 0) or 0)
+                        * int(_meta.get("seqlen", 0) or 0))
+    if _norm_tokens > 0:
+        _renormed = 0
+        for _entry in stats.values():
+            if not isinstance(_entry, dict) or "h_trace_raw" not in _entry:
+                continue
+            _entry["h_trace"] = float(_entry["h_trace_raw"]) / _norm_tokens
+            if "h_w2_sum_raw" in _entry:
+                _entry["h_w2_sum"] = (
+                    float(_entry["h_w2_sum_raw"]) / _norm_tokens)
+            _per_raw = _entry.get("h_trace_per_expert_raw")
+            if _per_raw is not None:
+                _entry["h_trace_per_expert"] = [
+                    float(v) / _norm_tokens for v in _per_raw]
+            _renormed += 1
+        print(f"[alloc] Fisher renormalized: {_renormed}/{len(stats)} rows "
+              f"→ h_trace_raw / {_norm_tokens} global calib tokens "
+              "(per-expert routed-count normalization corrected)", flush=True)
+    else:
+        _needs_fix = sum(
+            1 for _e in stats.values()
+            if isinstance(_e, dict) and "h_trace_raw" in _e
+            and int(_e.get("n_tokens_seen", 0) or 0) > 0)
+        if _needs_fix:
+            print("[alloc] WARNING: probe meta lacks nsamples/seqlen; cannot "
+                  "renormalize h_trace by the global calib token count. "
+                  "Per-expert Fisher may be inflated by routed-token "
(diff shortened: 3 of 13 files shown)


RobTand:
**Verdict: merge with notes.** The math is right — I derived it independently rather than accepting the framing, because it is the crux.

For a mean-over-tokens loss `L = (1/N) Σ_t ℓ_t`, the diagonal empirical Fisher is `(1/N) Σ_t g²`. For a weight in expert `e`, tokens not routed to `e` give `g = 0` — a **genuine zero that belongs in the average**. Dividing by `|R_e|` yields the *conditional* per-routed-token Fisher, which is the right object for a tail objective but not for the mean-Δloss the allocator actually optimizes. **Global is correct.**

**Two corrections to the PR's framing, both of which cut in your favour on risk:**

1. **DSv4-Flash — your motivating architecture — is packed-3D**, per `specs/deepseek_v4.json` (`packed_experts.param_names = [gate_up_proj, down_proj]`), i.e. one of the classes you list as unchanged. The layouts this actually changes are, per `model_profiles/base.py:470-482`, the MiniMax-M2 / Qwen3 / Qwen3.5 MoE per-expert-`nn.Linear` layouts.
2. But under the current build venv (`transformers 5.6.0`) those instantiate `*Experts` containers with 3-D packed params — I checked empirically (`Qwen3MoeExperts`, `packed=True`). **So under today's toolchain the change is near-inert on every shipped artifact class**, and bites only where per-expert `nn.Linear` rows genuinely get probed: older transformers (MiniMax pins 4.57.5), vendored modeling files, `--model-override` runs on older images.

Invariants verified: dense-trunk unchanged (one forward/backward per layer over the full batch, so `n_tokens_seen` already equals the global count); packed-3D unchanged; the back-correction is genuinely idempotent (`h_trace_raw` is never mutated); the `"h_trace_raw" not in _entry → continue` guard correctly skips AURA's stats payload, so `COST_MODE=aura` is unaffected.

**Please address before merge:**

- **The old behaviour was deliberate, and the PR does not engage with it.** `sensitivity_probe.py:1834-1842` documents it as "the one implicit ÷token-fraction the MoE convention prescribes" and records that an explicit `÷route_prob` was removed in audit M4 *because* the implicit one already applied it. CLAUDE.md §3 says the same. I think your reversal is right — but it reverses a documented decision, so it should say so, and both places need updating in the same commit or they become actively misleading.
- **The fix lands in two of three places.** `incremental_probe.py:2528` and `sensitivity_probe.py:1887,:1909` still normalize the h-detail blobs by `n_tokens_seen`. Those feed `predicted_dloss` via `measure_quant_cost.py:620`, which the allocator consumes. Before this PR the scalar and the detail shared one denominator; after it they disagree by `global/routed` on expert rows — the same mixed-units error, relocated. A run mixing measured-`output_mse` rows with h-detail-fallback rows would price them on incompatible scales inside one knapsack.
- **Probes lacking meta ship the old wrong values behind a WARNING** (`allocator.py:1325-1334`). Given the house style, a hard fail or an explicit `--allow-legacy-fisher-norm` is more consistent.
- "up to ~33,000×" is a degenerate worst case (1 token of 32,768); typical is `n_experts/top_k` ≈ 32×.

**Done well:** extracting `finalize_fisher_stats` as a testable pure function; the self-describing `meta.fisher_norm_tokens` marker; back-correcting from stored raw accumulators so existing pickles need no re-probe. `tests/test_fisher_normalization.py` builds a real two-Linear autograd reproducer instead of asserting hand-written numbers — better than this repo's median.

**Validation before it influences a ship:** a probe + allocation diff on a model whose experts genuinely probe as per-expert `nn.Linear` (MiniMax under 4.57.5), then served KL-vs-BF16 + WikiText PPL at matched bpp. On the packed-3D families a byte-for-byte `layer_config.json` diff is sufficient proof of no-op and needs no GPU.

tylerwagler:
All four addressed — two commits pushed on top of the branch (plus a lockfile-noise cleanup).

**1. The documented decision — agreed, and now reversed *on the record* (`95b3993`).** The M4 comment block, `finalize_fisher_stats`'s docstring, and CLAUDE.md §3 now all state the same thing in one commit: per-routed-token was deliberate, M4's real invariant (one division, `route_prob` metadata-only) still holds, and the *denominator* is what this PR reverses — the implicit ÷token-fraction is the same 1/p_e inflation M4 set out to remove, just relocated. Also caught `docs/prismaquant_design.md` §2.2 still describing the pre-M4 RouterTracker ÷route_prob as current behavior and fixed it there too.

**2. Confirmed, and it was slightly worse than cited (`95b3993`).** Besides `incremental_probe.py:2528` and `sensitivity_probe.py:1887/:1909`, the sensitivity backend's *scalar* finalize was also still per-row, and the MoE-block-flush rows write blobs with `n_tokens_seen == 0` → divided by 1, i.e. raw token-summed blobs labeled "per_token". All writers now funnel the global count: `finalize_fisher_stats` moved next to `h_detail_blob` as the single convention owner, `FisherAccumulator.finalize` takes `global_tokens`, blobs are stamped `h_detail_version: 4` + `norm_tokens`, and the sensitivity/MTP pickles stamp `meta.fisher_norm_tokens` so your load-time renorm is idempotent on them. Tests pin `sum(h_diag) == h_trace` (shared-denominator identity) at the writer level and end-to-end through the sensitivity backend. One caveat documented rather than enforced: v3 blobs for unpacked expert Linears are indistinguishable from dense v3 blobs from the blob alone (same shape as the M4 route_prob caveat), so the consumer accepts them with a regenerate-if-expert-rows-matter note — packed-3D and dense v3 blobs are numerically identical to v4.

**3. Hard fail shipped (`612fc38`).** `renormalize_probe_fisher()` raises SystemExit when a probe carries raw accumulators but no token meta; `--allow-legacy-fisher-norm` restores the warning path for reproducing historical allocations. Detection widened while there: the old `n_tokens_seen > 0` gate was blind to exactly the flush-path rows that are worst-scaled. Unit-tested (renorm, idempotence, hard fail, escape hatch, no-raw no-op).

**4. Tempered (`612fc38`).** The one in-code "~33,000×" now reads "typically ~n_experts/top_k (≈32×); ~33,000× only in the degenerate 1-routed-token case", and the PR body is updated with both that and your scope correction (DSv4-Flash probes packed-3D on the current toolchain; the change bites on per-expert `nn.Linear` layouts — older pins, vendored modeling files, `--model-override` on older images). The commit message's phrasing I can't amend without a rebase.

On your validation ask: agreed — a probe + allocation diff on a model that genuinely probes per-expert `nn.Linear` (MiniMax under 4.57.5) plus the served KL/PPL lane before this influences a ship, and the byte-for-byte `layer_config.json` no-op diff on the packed-3D families. Queued on our side.

Full suite: 969 passed; the 9 failures reproduce bit-identically at the pre-change head (3 stale-base doc-consistency tests that heal on merge since the PR doesn't touch README/runtime_flags; the rest need `/home/rob` fixtures, dataset access, or the vendored-transformers env).

RobTand:
**Merged.** All four items land, and I re-derived the invariants rather than accepting them: `h_detail_blob` is now the only division site, all four writer call sites pass the global count, the `sum(h_diag) == h_trace` identity holds at writer level, and the renormalization is idempotent *structurally* (it always recomputes from `h_trace_raw` and never touches a derived `h_trace`), not merely stamp-dependent.

**Your scope correction is confirmed by measurement on the real probes** — this is the number I most wanted:

| probe | rows | `h_trace` rows changed by the renorm |
|---|---|---|
| 35B-A3B (packed-3D) | 391 | **0** |
| Hy3 295B (packed MoE) | 719 | **0** |
| Qwen3.6-27B | 505 | 8 |

The 27B's eight are exactly the MTP Linears, each by ratio **8176/8192 = 0.998** — MTP's two-token shift, uniform across its rows and ~100× below cost-model noise. Body rows are bit-identical. So "near-inert on today's shipped artifact classes" is right, and no shipped artifact is implicated.

**I landed two of the follow-ups rather than filing them** (`4bf27ba`):

1. **The v3 h-detail hole was still open.** You stamp `h_detail_version` and `norm_tokens`, but **no consumer read either field**, so a stale v3 blob directory was silently consumed alongside v4-normalized scalars — the same mixed-units error this PR set out to close, relocated to the blob path, with unpacked per-expert rows entering the knapsack ~`n_experts/top_k` hotter than every scalar-priced row. `HDetailIndex` now refuses on a `norm_tokens` mismatch at construction *and* on every read, distinguishing "no stamp" from "written at a different calibration size". Deliberately no env override: h-detail is optional, so dropping `--h-detail-dir` is already the safe escape, which beats a flag that admits known-mis-scaled units.
2. **Visual-probe rows broke the idempotence claim.** They're finalized on their own token count but merged into a body probe whose meta wins, so `renormalize_probe_fisher` re-divided them by the body's count — 32× low under pipeline defaults. The denominator is now resolved **per row** from the `h_trace_norm_tokens` stamp your `1fc9775` already writes, falling back to meta. That also made the v3 gate stricter for free, since a visual blob stamped with the body's count is now caught too.

Note that the renormalization runs unconditionally on the default path, so this is the one live change in the stack for per-expert-`nn.Linear` layouts. Your validation plan stands as the gate before it influences a ship: a probe + allocation diff on a model that genuinely probes per-expert (MiniMax under 4.57.5), plus served KL/PPL — and the byte-for-byte `layer_config.json` no-op diff on the packed-3D families, which I've now done for you at `TARGET_BITS` (0 of 500 assignments move on the 35B).

PR #15, opened 2026-07-17
Footprint: per-tensor source-byte manifest; reject negative floors

tylerwagler:
> Stacked on #14 — only the last commit is new here; I'll rebase as predecessors merge.

**Commit:** `footprint: per-tensor source-byte manifest; reject negative floors`

### Summary
`source_regime` / `reencoded_source_bytes_for_shape` charge EVERY re-encoded
Linear at the FP8_SOURCE layout (1 B/param + fp32 block scales) as soon as
any F8 dtype appears in the checkpoint. On a mixed-precision source (an
MXFP4-packed DSv4-Flash checkpoint: I8 nibble experts at ~0.53 B/param on
disk + F8 attention + BF16 floor) that removed more source bytes than the
checkpoint holds — a negative non-quantizable floor, which let an artifact
more than twice the byte budget "fit" it.

- `source_tensor_bytes_manifest()`: exact on-disk bytes per weight tensor
  (safetensors header `data_offsets` span + `.scale` / `.weight_scale_inv`
  siblings), keyed by live qname. Sums of actual spans can never exceed the
  checkpoint total, so a manifest floor is ≥ 0 by construction.
- The allocator's byte-budget selector and `floor_bytes_for_model` (which
  has the model path) now price from the manifest; names missing from the
  manifest stay in the floor (over-count only, loud warning).
- `check_floor_non_negative()`: a negative floor is always an accounting
  bug — raise `ValueError` with the per-tensor-class byte breakdown instead
  of shipping a selection.

Stated plainly: `assignment_artifact_bytes` prices candidates from
precomputed scalars and has no model path, so it keeps the regime rate — on
a mixed source it now **fails loudly** with `ValueError` rather than being
silently wrong; callers that hit this should price via
`floor_bytes_for_model` / the manifest.

Uniform bf16 or fp8 sources: byte-identical accounting (the manifest spans
equal the regime rates there). Mixed sources: correct floors where the old
path under- or over-counted, and a hard error instead of a silently negative
floor.

### How to verify
`pytest tests/test_footprint.py -q` — includes a tiny mixed-source
safetensors fixture (I8-packed expert + E8M0 scales, F8 attention + fp32
scale_inv, BF16 floor) showing the regime path's floor going negative and
rejected while the manifest floor is exact and non-negative.

### Tests
New in `tests/test_footprint.py`: manifest span accounting, the
negative-vs-non-negative floor contrast, and `floor_bytes_for_model` on the
mixed fixture. Existing footprint identities unchanged and still pass.

https://claude.ai/code/session_01QXh7ricvRaq4pMg4RKPDf5

### DIFF
--- a/prismaquant/allocator.py
+++ b/prismaquant/allocator.py
@@ -2042,7 +2042,18 @@ def main():
                 "floor (lm_head/embed/norms).")
         budget_bytes = float(args.target_disk_gb) * _fp.GB
         src_total, src_by_dtype = _fp.source_checkpoint_bytes(probe_model_path)
-        regime = _fp.source_regime(src_by_dtype)  # robust bf16/fp8 (not by mass)
+        regime = _fp.source_regime(src_by_dtype)  # recorded for reporting only
+        # Per-tensor source-byte manifest: each re-encoded Linear is charged
+        # its ACTUAL header byte span (weight + scale siblings), never a
+        # regime-wide per-param rate. On mixed sources (an MXFP4-packed
+        # DSv4-Flash checkpoint: I8 nibble experts + E8M0 scales, F8
+        # attention, BF16 floor) the old regime accounting removed more
+        # bytes than the checkpoint holds, driving the floor negative.
+        src_manifest = _fp.source_tensor_bytes_manifest(
+            probe_model_path,
+            name_map=getattr(model_profile, "checkpoint_to_live_name", None),
+        )
+        manifest_missing: set[str] = set()
 
         def _artifact_for_target(t: float):
             assign_t, ach_t, tot_t, _mut = _solve_for_target(t)
@@ -2050,13 +2061,23 @@ def main():
                 return None
             expanded_t = _expand_assignment_for_seed_json(assign_t)
             body_aux = _assignment_bits_total(expanded_t) / 8.0
-            reenc_src = 0
+            reenc_by_name: dict[str, int] = {}
             for n in expanded_t:
-                e = _stats_entry_for_assignment_name(n)
-                if isinstance(e, dict):
-                    reenc_src += _fp.reencoded_source_bytes_for_shape(
-                        _shape_from_stats(e), regime)
-            floor = float(src_total) - reenc_src
+                nb = src_manifest.get(n)
+                if nb is None and n.endswith(".weight"):
+                    nb = src_manifest.get(n[: -len(".weight")])
+                if nb is None:
+                    # Not resolvable in the checkpoint headers: leave the
+                    # tensor priced in the floor (conservative — the artifact
+                    # estimate can only over-count, never under-count) and
+                    # report it loudly after the grid pass.
+                    manifest_missing.add(n)
+                    continue
+                reenc_by_name[n] = int(nb)
+            floor = float(src_total) - sum(reenc_by_name.values())
+            _fp.check_floor_non_negative(
+                floor, float(src_total), reenc_by_name,
+                context=f"byte-budget selector (target_bits={t:.3f})")
             return {
                 "target_bits": float(t), "achieved_bits": float(ach_t),
                 "bpp": float(ach_t), "dloss": float(tot_t),
@@ -2070,6 +2091,15 @@ def main():
             r = _artifact_for_target(float(row["target_bits"]))
             if r is not None:
                 grid.append(r)
+        if manifest_missing:
+            print(
+                "[alloc] WARNING: "
+                f"{len(manifest_missing)} re-encoded Linears not found in the "
+                "source checkpoint manifest; their source bytes stay in the "
+                "floor (artifact size over-counted, never under-counted). "
+                f"Sample: {sorted(manifest_missing)[:8]}",
+                flush=True,
+            )
         sel = select_under_byte_budget(grid, budget_bytes)
 
         rd = knee_summary.get("rd_curve") if isinstance(knee_summary, dict) else None
@@ -2080,6 +2110,8 @@ def main():
             "budget_bytes": budget_bytes,
             "source_total_bytes": float(src_total),
             "source_regime": regime,
+            "source_accounting": "per_tensor_manifest_v2",
+            "manifest_missing_linears": len(manifest_missing),
             "source_bytes_per_param": int(
                 _fp.dominant_source_bytes_per_param(src_by_dtype)),
             "feasible": bool(sel["feasible"]),
--- a/prismaquant/allocator.py
+++ b/prismaquant/allocator.py
@@ -2049,11 +2049,16 @@ def main():
         # DSv4-Flash checkpoint: I8 nibble experts + E8M0 scales, F8
         # attention, BF16 floor) the old regime accounting removed more
         # bytes than the checkpoint holds, driving the floor negative.
+        # `expert_parent_for_projection` bridges the per-expert-on-disk /
+        # packed-live MoE layouts (…experts.{i}.gate_proj -> the packed
+        # …experts.gate_up_proj the allocator names), same mapping the
+        # layer-streaming pack bridge uses.
         src_manifest = _fp.source_tensor_bytes_manifest(
             probe_model_path,
             name_map=getattr(model_profile, "checkpoint_to_live_name", None),
+            expert_parent_for_projection=getattr(
+                model_profile, "packed_expert_parent_for_projection", None),
         )
-        manifest_missing: set[str] = set()
 
         def _artifact_for_target(t: float):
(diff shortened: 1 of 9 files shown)


RobTand:
**Verdict: needs changes.** The diagnosis is right, the direction is right, and the negative-floor invariant is exactly the kind this codebase wants. But as written it silently regresses every packed-MoE architecture, which I reproduced.

**Blocking — the manifest cannot resolve packed-expert names.** `footprint.py:176-177` skips any checkpoint key not ending in `.weight`, and `checkpoint_to_live_name` does not map per-expert checkpoint keys to the packed live name. Allocator assignment names for packed MoE are `…experts.gate_up_proj` / `…experts.down_proj`. Both on-disk layouts miss:

```
CASE A (per-expert 2D on disk):
  manifest keys: [...experts.0.gate_proj, ...experts.0.up_proj, ...experts.0.down_proj]
  allocator name ...experts.gate_up_proj -> manifest hit? None
  allocator name ...experts.down_proj    -> manifest hit? None

CASE B (packed 3D on disk, checkpoint key does not end in ".weight"):
  manifest keys: ['model.embed_tokens']      # experts skipped entirely
```

At `allocator.py:2064-2075` every expert then falls into `manifest_missing`, its source bytes stay in the floor **while `_assignment_bits_total` still adds its quantized body bytes** → `artifact_bytes ≈ source_total + body_quant`. On a 35B-A3B, where experts are ~90% of params, that is ~90 GB instead of ~20 GB, so every rung returns `below_floor: True` and the run exits with "`--target-disk-gb=… is below the floor`".

The old code priced these correctly via `_stats_entry_for_assignment_name` → `_shape_from_stats` (packed stats entries carry `n_params`), so this is a genuine regression rather than a pre-existing gap. It affects Qwen3.5/3.6-MoE, Hy3, LFM2, Gemma4 and DSv4 — every shipped MoE class. Mitigating: `--target-disk-gb` is opt-in and `run-pipeline.sh` does not pass it, so the default pipeline is unaffected — but byte-budget "fit the card" is the current bit-rate selection method, so this would bite in practice.

The new fixture uses `layers.0.experts.0.down_proj.weight`, an *unpacked* name, which is why the tests stay green. A packed-MoE case in `tests/test_footprint.py` would have caught it.

Fix direction: either teach `checkpoint_to_live_name` the per-expert→packed mapping (the bridge logic already exists at `layer_streaming.py:650-680`) or have `source_tensor_bytes_manifest` take a name-resolution callback from the profile.

**Also:**

- **`assignment_artifact_bytes` was guarded, not fixed** (`footprint.py:343`). It still uses the buggy regime path and merely raises when the result goes negative, so it now disagrees with `floor_bytes_for_model`, and on an fp8-native model whose block-scale layout is not 128×128 fp32 it can **raise on a legitimate config** where it previously returned a slightly-negative floor. Finish it or deprecate it.
- **The `manifest_missing` warning fires after the numbers are consumed** (`allocator.py:2091-2100`), so on a packed-MoE model the operator sees a fatal "below the floor" exit and *then* the explanation. Invert that, or make a non-empty `manifest_missing` a hard error naming the tensors.

**Verified true:** the regime path does charge every re-encoded Linear at FP8_SOURCE rates once any F8 appears; it does go negative on a mixed source (your test builds it); and uniform bf16/fp8 sources are byte-identical before and after — I measured on the real Qwen3-0.6B (`4194304 == 4194304`) and on a synthetic fp8 with a non-128-multiple shape (`60024 == 60024`).

**Done well:** summing actual safetensors header spans is strictly more faithful than any per-param rate, "a negative floor is always an accounting bug" is the right invariant, and naming the offending tensor class in the exception turns a silent selection error into a diagnosable one.

*(edited: a shell quoting slip mangled two lines of the original)*

tylerwagler:
All three reproduced against the code as described — no disputes. Fixed and pushed (2 commits, `28490e3` + `83ad768`).

**Packed-expert resolution (blocking):** `source_tensor_bytes_manifest` now handles both layouts. Case B: suffix-less keys are kept — only sidecar suffixes (`.scale` / `.weight_scale_inv` / `.weight_scale`) fold into their base, and the base may itself be suffix-less. Case A: per-expert spans additionally aggregate into the packed parent via a new `packed_expert_alias()`, driven by the profile's `packed_expert_parent_for_projection` — the same mapping the `layer_streaming` per-expert→packed bridge uses, so no duplicated architecture logic (legacy gate/up/down fallback when no profile). Per-expert entries are kept alongside the aggregate so per-expert-named allocations (MiniMax-style) still resolve; `reencoded_names` uses one naming scheme per tensor, so there's no double-subtraction path.

**`assignment_artifact_bytes`:** fixed rather than deprecated. It takes `source_manifest=` and prices through the same shared resolver as `floor_bytes_for_model` (which now returns the manifest and accepts the profile mappers), so the two agree exactly; a `source_accounting` field says which path priced it. The regime path survives only for the uniform bf16/fp8 cases where it's exact, and still rejects (never ships) a negative floor.

**Hard error before consumption:** the selector now prices each rung via `resolve_reencoded_source_bytes`, which raises a `ValueError` naming the unresolvable tensors (sample of 10 + count) before any grid numbers are computed. The post-grid warning and the now-always-zero `manifest_missing_linears` selection field are gone. Behavior change worth noting: a genuinely unresolvable pinned/aux tensor now fails fast with names instead of silently over-counting — if that surfaces on a profile, the mapping is the thing to fix.

**Tests:** packed-MoE coverage for both cases (Case A end-to-end incl. gate+up fusion and NVFP4 per-expert sidecars; Case B suffix-less 3-D keys with sidecar attach), manifest-vs-`floor_bytes_for_model` agreement on the mixed fixture, and hard-error coverage through all three consumers. `test_footprint.py` 20/20; full allocator/selection suites 133/133.

RobTand:
**Merged** — but this one arrived at `needs-changes` and landed with a fix from my side, because the highest-value question in my round-1 review turned out to be a real regression.

**Reproduced on the real Qwen3.6-27B checkpoint.** The new hard error fires on a *working* `--target-disk-gb` invocation:

```
ValueError: [footprint] 8 re-encoded Linear(s) not resolvable in the source checkpoint
manifest (byte-budget selector (target_bits=4.500)): mtp.fc, mtp.layers.0.mlp.down_proj,
mtp.layers.0.mlp.gate_proj, ... mtp.layers.0.self_attn.v_proj
```

`main` completed the same run. Cause is a semantic mismatch rather than a missing tensor: `Qwen3_5DenseProfile.checkpoint_to_live_name("mtp.fc.weight")` returns `None` because transformers v5 dropped the MTP module from the live graph, and the manifest read that as "no source bytes" — while the exporter still re-encodes `mtp.*` from exactly those bytes and the allocator assigns them under their raw names. My reviewer independently found `model.visual.*` hits the identical path on any `CALIBRATION_MODALITY=multimodal` probe. Since `--target-disk-gb` is the current bit-rate selection lane, the hard error you added to prevent a silent over-count instead broke the path it was protecting.

Fixed at the accounting layer, not by relaxing the check (the check is right — an unresolved re-encoded tensor really would leave its source bytes in the floor while its quantized bytes are added): a declined key falls back to its raw checkpoint name. Inert for tensors nothing re-encodes, since the floor is `checkpoint_total − Σ(resolved re-encoded spans)`. See `0a9dc00`.

**With that in place your central claim checks out on real data, not fixtures.** Floors are byte-identical to `main` on both checkpoints — 27B **6.012 GB**, 35B **4.661 GB** — so "uniform bf16 source ⇒ byte-identical accounting" holds. At a feasible budget the selection completes and moves slightly: at a 26 GB card the 35B now fills 25.951 GB vs 25.859 before, which is #16's cleaner rung x-axis rather than anything in this PR.

**(a) and (c) verified.** Both on-disk layouts resolve, `packed_expert_alias` is driven by the profile mapper and that mapper *is* passed on the real path (`allocator.py:2200-2205`), and the raise genuinely precedes the grid numbers.

**(b) is overstated and I'm recording it rather than holding the merge.** The two functions do agree when handed the same manifest, but `assignment_artifact_bytes`, `assignment_artifact_gb` and `floor_bytes_for_model` are referenced *only* from `tests/test_footprint.py` — production inlines a third copy of manifest→resolve→check in the selector. So the tested "the two paths agree exactly" property is never exercised against the code that ships selections, and `source_manifest` defaults to `None`, meaning the regime path with its 128×128-fp32 assumption is still the public function's default behaviour. Follow-up filed.

Two latent items also filed: a per-expert tensor's bytes live in the manifest twice (own key + packed aggregate) with only a docstring convention preventing double subtraction, and `check_floor_non_negative` only catches that when the over-subtraction exceeds the entire floor.

PR #16, opened 2026-07-17
Solver: feasible-only termination + solve memoization

tylerwagler:
> Stacked on #15 — only the last 6 commits are new here (2 original + review round 1); I'll rebase as predecessors merge.

**Commits:** `solver: only feasible iterates terminate solve_with_promotion` ·
`allocator: memoize target solves; bound byte-budget ratchet ceiling`

### Summary
`solve_with_promotion` had three silent-acceptance exits that returned
iterates violating the rung's contract: `solve_allocation` returning None
handed back the previous massively over-target iterate; `overshoot <=
tolerance` accepted arbitrarily deep undershoot (a 6.0 rung shipping
achieved 4.95 at 25× worse predicted loss); and the stall exit returned an
iterate still far above target. Scoped precisely (review round 1): the
genuinely dangerous case was the `--target-bits` emit path, which shipped an
over-budget layer_config with no error. On the byte-budget path the damage
was milder — over-target iterates were priced at their true, larger disk
size, so the selector's feasibility calls stayed correct; the rung was
mislabeled (a "5.0" row that actually cost more), skewing the Pareto curve's
x-axis rather than fabricating feasibility.

New termination contract: a rung either returns a genuinely feasible iterate
(`achieved <= target + tolerance`) or reports INFEASIBLE (`(None, nan)`) and
callers exclude it from the Pareto curve and byte-budget bisection. Among
feasible iterates the solver keeps the one with minimum predicted Δloss
(tie-break toward higher achieved bits) — the solver's actual objective;
CLAUDE.md §5 records that denser is not monotonically better. Two-phase
search: damped descent on the tightened DP target (accelerating across
promotion plateaus caused by serving-group atomicity) until the first
feasible iterate, then bracket bisection ratcheting the best feasible
assignment — undershoot is recovered by bisecting back toward the target,
not shipped. Floor corners terminate explicitly (a floor solve that still
promotes over target breaks immediately instead of re-running the identical
DP until max_iters).

The memoization commit removes duplicate work around it: `_solve_for_target`
is a pure function of the target given fixed stats/candidates, so the
byte-budget grid stops re-solving rungs the sweep already solved, and the
fill-the-card ratchet's ceiling is clamped to the cheapest rung whose exact
footprint already exceeds the budget (no more multi-minute DP solves at
targets that cannot fit).

### Review round 1
- `061853e` re-ratchets on min `total_predicted_dloss` among feasible
  iterates (the value was already computed and discarded), tie-break toward
  denser; discriminating test included (fails against the old max-bits
  ratchet).
- `f89d378` hands out copies of memoized solves (hit and first-miss paths)
  and labels the `search_hi` bound a HEURISTIC in-code.
- `340885e` documents `PRISMAQUANT_SOLVER_TRACE` in `docs/runtime_flags.md`
  (placed to merge cleanly with #18's refresh — verified via merge-tree).
- `dbfa95d` scopes the silent-fallback claim honestly in the docstring.

Pareto rows that previously violated the contract now come back as
INFEASIBLE or as a genuinely feasible iterate. Curves become weakly
monotone; downstream byte-budget bisection no longer chases mislabeled rows.
`--target-bits` runs that previously shipped an over-budget config now exit
with an explicit infeasibility error. `stall_grace` is retained for call
compatibility but unused.

### How to verify
`pytest tests/test_allocator_sibling_aggregation.py tests/test_allocator_solver_bins.py -q`
— includes `test_solve_with_promotion_returns_feasible_iterate_on_plateau`
(renamed from the stale `..._stops_on_stall`), which pins that a plateaued
loop returns a genuinely feasible iterate, never an over-target one.
`PRISMAQUANT_SOLVER_TRACE=1` prints per-eval solve traces for inspection.

### Tests
Renamed/extended: `test_solve_with_promotion_returns_feasible_iterate_on_plateau`
(asserts `achieved <= target + tolerance`). New:
`test_ratchet_keeps_min_dloss_feasible_iterate_over_denser` and
`test_ratchet_tie_breaks_toward_denser_iterate`. Existing solver-bin suite
passes unchanged.

### DIFF
--- a/docs/runtime_flags.md
+++ b/docs/runtime_flags.md
@@ -80,6 +80,12 @@ The production pipeline fails fast when archived Fisher levers are requested.
 | `PRODUCTION_CACHE_PREFETCH` | `require` | Standalone recache prefetch policy. `require` fails fast when assignment-required weights cannot fit resident, preventing silent NVMe-bound replay. |
 | `PRODUCTION_CACHE_PREFETCH_WORKERS` | `4` | Thread count for eager production-cache prefetch. |
 
+## Allocator / solver flags
+
+| Flag | Default | What it does |
+|---|---|---|
+| `PRISMAQUANT_SOLVER_TRACE` | `0` (off) | Per-evaluation trace for `solve_with_promotion`: prints each tightened-target DP eval with achieved bits, predicted Δloss, and wall time, plus DP-infeasible probes. Read once at module import (no per-call getenv on the solve path) — set it before launching the allocator. Observability only; does not change the solve. |
+
 ## CUDA / system flags
 
 | env var | recommended | what it does |
@@ -189,6 +195,7 @@ PRISMAQUANT_PROD_ACT_SCALES
 PRISMAQUANT_RENDER_GATE_MIN_GAIN
 PRISMAQUANT_RENDER_PROGRESSIVE_GATES
 PRISMAQUANT_SHARED_WEIGHT_FORMAT_CACHE
+PRISMAQUANT_SOLVER_TRACE
 PRISMAQUANT_STRICT_ASSIGNMENT_COVERAGE
 PRISMAQUANT_STRICT_PRODUCTION_CACHE
 PRISMAQUANT_TMPDIR
--- a/prismaquant/allocator.py
+++ b/prismaquant/allocator.py
@@ -2042,7 +2042,18 @@ def main():
                 "floor (lm_head/embed/norms).")
         budget_bytes = float(args.target_disk_gb) * _fp.GB
         src_total, src_by_dtype = _fp.source_checkpoint_bytes(probe_model_path)
-        regime = _fp.source_regime(src_by_dtype)  # robust bf16/fp8 (not by mass)
+        regime = _fp.source_regime(src_by_dtype)  # recorded for reporting only
+        # Per-tensor source-byte manifest: each re-encoded Linear is charged
+        # its ACTUAL header byte span (weight + scale siblings), never a
+        # regime-wide per-param rate. On mixed sources (an MXFP4-packed
+        # DSv4-Flash checkpoint: I8 nibble experts + E8M0 scales, F8
+        # attention, BF16 floor) the old regime accounting removed more
+        # bytes than the checkpoint holds, driving the floor negative.
+        src_manifest = _fp.source_tensor_bytes_manifest(
+            probe_model_path,
+            name_map=getattr(model_profile, "checkpoint_to_live_name", None),
+        )
+        manifest_missing: set[str] = set()
 
         def _artifact_for_target(t: float):
             assign_t, ach_t, tot_t, _mut = _solve_for_target(t)
@@ -2050,13 +2061,23 @@ def main():
                 return None
             expanded_t = _expand_assignment_for_seed_json(assign_t)
             body_aux = _assignment_bits_total(expanded_t) / 8.0
-            reenc_src = 0
+            reenc_by_name: dict[str, int] = {}
             for n in expanded_t:
-                e = _stats_entry_for_assignment_name(n)
-                if isinstance(e, dict):
-                    reenc_src += _fp.reencoded_source_bytes_for_shape(
-                        _shape_from_stats(e), regime)
-            floor = float(src_total) - reenc_src
+                nb = src_manifest.get(n)
+                if nb is None and n.endswith(".weight"):
+                    nb = src_manifest.get(n[: -len(".weight")])
+                if nb is None:
+                    # Not resolvable in the checkpoint headers: leave the
+                    # tensor priced in the floor (conservative — the artifact
+                    # estimate can only over-count, never under-count) and
+                    # report it loudly after the grid pass.
+                    manifest_missing.add(n)
+                    continue
+                reenc_by_name[n] = int(nb)
+            floor = float(src_total) - sum(reenc_by_name.values())
+            _fp.check_floor_non_negative(
+                floor, float(src_total), reenc_by_name,
+                context=f"byte-budget selector (target_bits={t:.3f})")
             return {
                 "target_bits": float(t), "achieved_bits": float(ach_t),
                 "bpp": float(ach_t), "dloss": float(tot_t),
@@ -2070,6 +2091,15 @@ def main():
             r = _artifact_for_target(float(row["target_bits"]))
             if r is not None:
                 grid.append(r)
+        if manifest_missing:
+            print(
+                "[alloc] WARNING: "
+                f"{len(manifest_missing)} re-encoded Linears not found in the "
+                "source checkpoint manifest; their source bytes stay in the "
+                "floor (artifact size over-counted, never under-counted). "
+                f"Sample: {sorted(manifest_missing)[:8]}",
+                flush=True,
+            )
         sel = select_under_byte_budget(grid, budget_bytes)
 
         rd = knee_summary.get("rd_curve") if isinstance(knee_summary, dict) else None
@@ -2080,6 +2110,8 @@ def main():
             "budget_bytes": budget_bytes,
             "source_total_bytes": float(src_total),
             "source_regime": regime,
+            "source_accounting": "per_tensor_manifest_v2",
+            "manifest_missing_linears": len(manifest_missing),
             "source_bytes_per_param": int(
                 _fp.dominant_source_bytes_per_param(src_by_dtype)),
             "feasible": bool(sel["feasible"]),
--- a/prismaquant/allocator.py
+++ b/prismaquant/allocator.py
@@ -1751,8 +1751,24 @@ def main():
             return original_entry
         return None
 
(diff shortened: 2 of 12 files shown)


RobTand:
**Verdict: merge with notes.** Real improvement, and I measured the effect rather than taking it on trust. A/B of `solve_with_promotion` on an identical synthetic dense fixture (168 Linears, NVFP4/FP8_DYNAMIC/BF16):

```
target   main achieved / dloss          pr-17 achieved / dloss
4.75     4.7281 / 4.893323e+05          4.7281 / 4.893323e+05     identical
5.00     4.9889 / 4.334389e+05          4.9889 / 4.334389e+05     identical
5.50     5.4995 / 3.214231e+05          5.4995 / 3.214231e+05     identical
6.00     5.9775 / 2.345992e+05          5.9884 / 2.322705e+05     <- denser, -1.0% dloss
6.50     6.5099 / 1.503046e+05          6.5099 / 1.503046e+05     identical
```

Identical wherever the first iterate already lands inside the tolerance band, strictly better where the old code accepted a deep undershoot. Your "rung 6.0 returning achieved 4.95" complaint is real and Phase 2 fixes it.

**One change I'd ask for before merge — the ratchet optimizes the wrong quantity.** `allocator_solver.py:470` does `achieved, _ = compute_achieved(...)`, discarding the `total_predicted_dloss` that the same call already returns, then ratchets on **max achieved bits** at `:478-481`. The solver's objective is min-dloss subject to a bit budget; "denser is better" is a heuristic, and CLAUDE.md §5 explicitly records that more bits is *not* monotonically better (5.5 bpp beats 6.0 bpp on Qwen3-4B PPL). Ratcheting on min `predicted_dloss` among feasible iterates is free — the value is already computed — and is strictly aligned with the stated objective.

**Notes:**

- **"The three silent fallbacks all fabricated `feasible=True` rows" is overstated.** The old code returned achieved bits honestly, and the byte-budget path prices `disk_bytes` from the actual expanded assignment, so an over-target rung was priced correctly, just larger. The genuine problem is the `--target-bits` path, which would emit an over-target `layer_config.json` — i.e. ship above the requested budget. That is worth saying precisely, because it is the argument that carries the PR.
- **New hard-failure mode on an existing recipe.** `allocator.py:2284-2287` now raises `SystemExit` where `main` would silently emit an over-target config. That is the honest behaviour, but it can break a working invocation — worth a dry run against a real 27B/35B probe before it rides into the default path.
- **`search_hi` tightening contradicts its own premise** (`allocator.py:2167-2176`): it bounds the ratchet by `min(target_bits of over-budget rungs)` while the adjacent comment concedes `disk(target)` is locally non-monotone. Safe, but a heuristic.
- **Memoization returns a shared mutable dict** (`allocator.py:1770-1774`). Nothing mutates it today — I checked — but `dict(cached)` on the hit path is one character of insurance against a future caller that does.
- `PRISMAQUANT_SOLVER_TRACE` is undocumented. `docs/runtime_flags.md` was already 5 flags stale on `main`; I fixed those in #18, so please add this one.

**Done well:** the new docstring is a genuine termination contract rather than prose. The plateau-doubling in Phase 1 is a real insight — promotion is a staircase, since a packed group only leaves a format when every row does — and it targets the wasted-DP-eval problem directly. Keeping `stall_grace` in the signature with an explicit `del` preserves call compatibility while documenting the deprecation. And the rewritten test asserts the contract (`achieved <= target + 0.01`) where the old one was a tautology.

**Validation before it influences a ship:** re-solve an existing 27B and 35B probe/cost pair at the shipped `TARGET_BITS` and diff `layer_config.json` against the archived one. If it differs, that is a new allocation and needs the gold lane (n=8 x seqlen=512 full-vocab vLLM KL + direct WikiText PPL). Also confirm no rung that previously produced a config now hard-exits.

tylerwagler:
All addressed and pushed (4 commits).

**Ratchet → min Δloss: done (`061853e`).** You're right that this was the wrong quantity and the right one was already in hand — `_evaluate` now keeps the feasible iterate with minimum `total_predicted_dloss` from the same `compute_achieved` call, tie-broken toward higher achieved bits. Docstring contract updated; the `achieved <= target + tolerance` feasibility assertion is untouched, and there's a new test where a feasible 5.0-bit / Δloss-1.0 iterate beats a denser 6.0-bit / Δloss-3.0 one (verified it fails against the old ratchet), plus a tie-break test.

**Memo dict (`f89d378`):** hand-out is now a copy on both the hit and first-miss paths — the first caller of an uncached solve shared the dict with the memo too, so copying only on hits would have left half the hole open.

**`PRISMAQUANT_SOLVER_TRACE` (`340885e`):** documented. This branch predates #18's file state, so the entry sits in a region #18 didn't touch and I verified with `git merge-tree` that it merges cleanly with main — both your new sections and this one survive.

**`search_hi` (`f89d378`):** kept, comment now says "HEURISTIC" out loud: it assumes only-local non-monotonicity, and the bounded downside is shipping the proven grid pick.

**Overstated claim:** agreed — the byte-budget path priced over-target rungs at their true disk size, so it mislabeled rather than fabricated; the `--target-bits` emit path was the genuinely dangerous one. PR body reworded to say exactly that, and the docstring scoped the same way (`dbfa95d`).

**SystemExit on infeasible `--target-bits`:** no code change; agreed it needs a dry run against a real 27B/35B probe before it rides the default path — we don't have probe artifacts on this box, so that's queued as the pre-ship validation together with your re-solve + `layer_config.json` diff (and gold lane if it differs).

Tests: solver suites 21 passed (incl. the 2 new); all PR-touched suites 44 passed; the 9 other allocator-importing suites 252 passed.

RobTand:
**Merged.** The ask is genuinely fixed and the new test genuinely discriminates — I reproduced it: the old ratchet returns `({'w': 'F6'}, 6.0)`, the new one `({'w': 'F5'}, 5.0)`. Min-Δloss governs every acceptance path including phase-1 first-feasible, and the feasibility conjunct still gates the returned iterate. `(b)`, `(c)` and `(d)` are clean: the memo copies on both the hit and first-miss paths (shallow is sufficient — values are format-name strings, and no caller mutates), the `search_hi` comment now matches the code with its downside genuinely bounded, and the `PRISMAQUANT_SOLVER_TRACE` doc entry matches the implementation exactly.

**I measured the effect on the real 35B probe/cost pair rather than a fixture,** and it's a clean win in the direction you predicted:

| target | main achieved / Δloss | this PR achieved / Δloss |
|---|---|---|
| 4.50 – 4.85 | identical | identical |
| 5.00 | 4.942 / 8.7585e-02 | **4.982 / 8.5915e-02** |
| 5.25 | 5.208 / 7.0861e-02 | **5.233 / 7.0050e-02** |
| 5.50 | 5.471 / 6.1504e-02 | **5.485 / 6.0186e-02** |
| 7.00 | 6.982 / 2.5187e-02 | **6.994 / 2.4819e-02** |

Denser *and* lower predicted Δloss at every rung ≥5.0 — phase 2 recovering undershoot instead of shipping it. Identical wherever the first iterate already landed in the band. Scope stated plainly: that's the surrogate objective, not a served metric.

**Your (e) worry resolved empirically, both ways.** Every caller handles `(None, nan)` — no nan reaches a Pareto row and nothing unpacks naively. And on both shipped models **no rung became INFEASIBLE** and the 12-rung Pareto set is unchanged, so nothing that previously produced a config now hard-exits.

**Two fixes landed on top** (`4bf27ba`):

1. **The new objective was contaminated.** `compute_achieved` scored a priced row whose assigned format has no candidate as **0.0 Δloss** — cosmetic before, load-bearing the moment the ratchet started optimizing that sum, since the too-low total is exactly what min-Δloss prefers. It now raises, like `compute_assignment_predicted_dloss` already does on the same input. Verified unreachable in normal operation first: the only production call site's key set is identically `candidates.keys()`, and aux/pinned/fixed rows are merged only after solving.
2. **The infeasibility exit now carries numbers.** You were right that it needed a dry run; what it actually needed was diagnosability, since the solver computes the format floor and sees every over-target iterate and then discards both. `solve_with_promotion` takes an optional `diagnostics` dict filled in place on every return path (kept beside the memo so a cache hit can't lose it), and the exit reports the floor, what the floor solve itself promotes to, and the closest achieved bits — the numbers that separate "target below the format floor" from "serving-group promotion overshoots by 0.02 bpp". That matters because this exit kills stage [3/4] under `set -euo pipefail`.

Also added the test that was missing: nothing pinned the `(None, nan)` contract this PR is built on, so a regression restoring a silent over-target return would have passed the whole suite.

**One consequence worth recording:** the solver now ratchets on min Δloss while the byte-budget selector ratchets on max disk bytes, so `_artifact_for_target(t)` can legitimately return a sparser artifact for a denser `t` — which makes `disk(target)` *more* non-monotone by design, precisely the premise `search_hi` assumes away. Neither side is wrong (a fewer-bits/lower-Δloss iterate dominates), but `has_slack`/`headroom_gb` can now report an unfilled card with nothing explaining why. Filed as a follow-up.

PR #17, opened 2026-07-17
Allocator: packed serving groups become first-class DP units

tylerwagler:
> Stacked on #16 — split at review round 1: this PR is now the core refactor only; the two riders moved to #20 (bit-exact pricing) and #21 (role split), each with its review blocker fixed. I'll rebase as predecessors merge.

**Commit:** `allocator: packed serving groups become first-class DP decision units`
(+ review round 1: revert pair removing the riders, and `f148feb` minor fixes)

### Summary
A packed-MoE serving group (DSv4: 768 per-expert Linears per layer) is
atomic at serve time — every projection of every routed expert loads under
ONE scheme — but the DP priced upgrades PER ROW (~0.0001 avg bits per bin)
while `promote_serving_units` charged the WHOLE group (~0.1 avg bits): a
~1000x price mismatch. Mispriced expert rows topped the per-bin ranking, the
feasibility bisection over-tightened, and attention/shared rows (orders of
magnitude more dloss recovered per GB) starved at the floor while headroom
sat unused.

**`aggregate_packed_serving_groups()`** collapses each packed group into ONE
multi-choice DP item (per-format dloss/bytes = exact member sums), so the DP
and the serving constraint price identical moves and post-DP MoE promotion
is a validated no-op. Groups with no common legal format fall back to
individual rows. `--no-packed-aggregation` restores the old per-row pricing
(mirrors `--no-fused-aggregation`).

Synthetic before/after (in-tree, `tests/test_allocator_packed_group_units.py`,
end-to-end DP solve on synthetic costs): with per-row pricing the dense row
starves at the floor while the budget leaks into mispriced expert rows; with
group units the dense row wins the headroom and the group flips as one unit
when the budget affords it.

### Review round 1
Branch split as requested. This branch cannot be force-pushed from here, so
the split is revert-based: `be2a525` reverts the role-split commit and
`4a2ea52` reverts the bit-exact-pricing commit — net tree = core only
(squash-merge is clean; happy to force-push a linear core-only history
instead if you prefer). The riders are resubmitted with their blockers fixed
as #20 and #21, both stacked on this PR and independent of each other.

`f148feb` adds the two minor fixes from review:
- the packed super-item stat entry no longer copies `members[0]`'s
  `in_features`/`out_features` (no single (out, in) exists for a mixed
  gate_up/down group, so scaling was not the safe option): the entry omits
  per-member features, `_shape_from_stats` yields the total-parameter-
  consistent rank-1 view, and `_assignment_bits_total` — the one byte path
  that did not prefer the candidate — now prefers `_memory_bytes_by_format`;
- `predicted_dloss_stderr` aggregates in quadrature (`sqrt(sum stderr²)`)
  onto the super cost entry, and the group candidate carries
  `Σbase + z·sqrt(Σ(stderr·gain)²)` (aggregation takes `calibrated_gains` to
  mirror member pricing). Nuance found while verifying: at `z > 0` the old
  path over-hedged linearly (`z·Σstderr`, up to √N vs independence) at the
  DP level, while the dropped field zeroed the hedge for consumers of the
  aggregated cost table. Both fixed; `z == 0` stays byte-identical (the
  exact member-sum test still pins it).

Dense models: no change (no packed groups → aggregation is a no-op).
Packed-MoE models: allocations change by design — whole-group pricing is the
fix; `--no-packed-aggregation` restores the old behavior for comparison.

### How to verify
`pytest tests/test_allocator_packed_group_units.py -q`

### Tests
`tests/test_allocator_packed_group_units.py`: aggregation mechanics, exact
member-sum pricing, expansion, no-common-format fallback, end-to-end
starvation before/after, whole-group flip, super-item shape/stderr fixes,
and that packed super-item names keep the `.mlp.experts.` prefix so
name-scoped `when.regex` profile rules bind at both gates.
`test_per_row_pricing_disagrees_with_serving_charge_old_path` demonstrates
the bug on the old path; the fix test asserts
`promote_serving_units(expanded, ...) == expanded`.


RobTand:
**Verdict: needs changes — but the core refactor is the best work in the stack and is ready.** The problem is that three unrelated changes ride on one branch, and only the first is ready.

**The serving invariants hold.** This was the highest-stakes question and it checks out:

- One format per group is *structural*, not enforced after the fact: `aggregate_packed_serving_groups` makes each `packed_expert_format_group` key one multi-choice DP item, offers only formats legal for every member (`common_fmts` intersection), and `expand_packed_group_assignment` broadcasts one format to all. A mixed-format group is not representable.
- `promote_serving_units` is **still called** — in the solver, in `_expand_assignment_for_seed_json`, and finally at `allocator.py:2314`. The union-find backstop is intact; aggregation layers on top rather than replacing it.
- The empty-intersection fallback keeps members as individual rows, and I checked the interaction with the pre-existing `validate_final_serving_promotion_noop`: fallback rows are un-aggregated, the solver's internal promotion coheres them, and the final promotion stays a no-op. No new hard-exit path.
- `aggregate_fused_siblings` and the packed path skip each other's names, so no Linear lands in two super-items.

**Blocking (commit `e5c4993`) — `cost_entry_is_bit_exact` prices W·A· formats at zero.** `allocator_candidates.py:222-253` asserts "a bit-identical weight tensor cannot perturb the layer output" and returns `predicted_dloss == 0.0`, discarding the measured `output_mse`. That premise is false for every quantized format in the menu. Measured:

```
NVFP4       act_quant_changes_input=True
FP8_DYNAMIC act_quant_changes_input=True
MXFP4       act_quant_changes_input=True
MXFP8_E4M3  act_quant_changes_input=True
BF16 / FP8_SOURCE / NVFP4A16   act_quant_changes_input=False
```

`measure_quant_cost.py:607` applies `activation_quantize_dequantize(X)` before computing `output_mse`, so for a weight-lossless W4A4 format that number is **real A-side error, not noise**. The failure case is your own motivating one: on an MXFP4-packed DSv4 source, MXFP4 re-encode gives `weight_mse == 0.0` exactly, so it prices at `dloss == 0.0` — the global minimum, unbeatable at any budget — and the RD curve claims zero loss for a W4A4 assignment.

There is a clean derivable rule available: short-circuit to zero only when the format's `activation_quantize_dequantize` is the identity. That is a dtype-level fact rather than a heuristic, it keeps BF16/FP8_SOURCE/NVFP4A16 passthrough correct, and it forces W·A· formats to carry their measured A-side cost. Blast radius today is bounded — `production_render_cost` and `aura_cost` both set `cost_source` and are exempted — so the exposed paths are `COST_MODE=local` and the GGUF lane.

**Blocking (commit `5f7072f`) — `--packed-role-split` has no serving-profile gate.** The help text says it is "legal when the serving lane supports per-role expert schemes (e.g. GGUF)", but nothing enforces it: `--packed-role-split --target-profile vllm_packed_moe` is accepted and can emit gate_up=NVFP4 with down=FP8 in the same MoE layer. vLLM's `CompressedTensorsMoEMethod` selects one scheme per FusedMoE layer, so that is a checkpoint vLLM cannot load — the hard-serving-invariant class. It should hard-error unless the resolved profile declares per-role expert schemes. The flag is off by default and the DP↔promotion consistency is correct; the gap is purely the missing servability gate.

**Minor:** the super-item stat entry sets `n_params` to the sum but `in_features`/`out_features` to `members[0]`'s, so `_shape_from_stats` on it yields a wrong shape (harmless today since every byte path prefers the candidate, but a live trap). Aggregation also drops per-member `predicted_dloss_stderr`, silently disabling the UCB hedge on expert rows when `COST_UCB_Z > 0` — the correct aggregate is `sqrt(sum of stderr^2)` under independence. Impact nil today at `COST_UCB_Z=0`.

**Please split the branch.** `7e4978f` is ready and I would merge it as-is; `e5c4993` and `5f7072f` need the changes above. Landing them together holds up the good one.

**Done well — and this is genuinely sharp.** The core insight is correct: pricing a serving-atomic group per-row inside the DP while `promote_serving_units` charges the whole group is a *systematic* mispricing that starves cheap dense rows, and making the group a first-class DP unit is the right fix rather than a correction factor. `tests/test_allocator_packed_group_units.py` is exemplary — `test_per_row_pricing_disagrees_with_serving_charge_old_path` demonstrates the bug on the old path before the next test shows the fix, and it asserts `promote_serving_units(expanded, ...) == expanded`, pinning the "promotion is now a validated no-op" claim instead of asserting it in prose.

**Validation before it influences a ship:** re-solve the 27B/35B probe+cost pairs with and without `--no-packed-aggregation`; assert `validate_final_serving_promotion_noop` passes and that the exported checkpoint loads in vLLM eager **and** graph mode. If `--packed-role-split` is to be used at all it needs its own vLLM load smoke on the target lane.

tylerwagler:
Split done as requested — with one mechanical caveat: this branch isn't force-pushed from here, so the split is revert-based.

- `be2a525` reverts `5f7072f` (`--packed-role-split`), `4a2ea52` reverts `e5c4993` (bit-exact pricing), so this PR's **net tree is the core refactor only** — squash-merge is clean. If you'd rather have a linear core-only history without the revert pair, happy to force-push a reset branch (`7e4978f` + fixes) — your call.
- `f148feb` adds the two minor fixes you flagged:
  - the packed super-item stat entry no longer copies `members[0]`'s `in_features`/`out_features`. One judgment call: no single (out, in) exists for a mixed gate_up/down group (`in_features` differ by role), so scaling `out_features` wasn't the safe option — the entry now omits per-member features, `_shape_from_stats` falls to the total-parameter-consistent rank-1 view, and `_assignment_bits_total` (the one byte path that did *not* prefer the candidate) now prefers `_memory_bytes_by_format`.
  - `predicted_dloss_stderr` aggregates in quadrature onto the super cost entry, and the group candidate carries `Σbase + z·sqrt(Σ(stderr·gain)²)` (aggregation takes `calibrated_gains` to mirror member pricing). One nuance we hit verifying: at `z > 0` the old path wasn't hedge-*free* at the DP level — member candidates baked a **linear** `z·Σstderr` into the sums (an over-hedge up to √N vs independence); the dropped-field problem was real for consumers of the aggregated cost table (aura gate, repricing). Both are fixed; `z == 0` stays byte-identical (the exact member-sum test still pins it).

The riders are resubmitted as separate PRs, each = clean cherry-pick of the reviewed commit + a fix commit for its blocker:

- **#20** (`rob-pr5b-actquant-cost`) — `e5c4993` + the activation-identity gate: the shortcut now requires `FormatSpec.act_quant_changes_input == False`, a dtype-level fact backed by `act_bits is None`, with a registry-wide test pinning `act_bits is None ⟺ identity activation callable`; unknown formats never short-circuit.
- **#21** (`rob-pr5c-role-split-gate`) — `5f7072f` + the serving-profile gate: profiles declare `supports_per_role_expert_schemes` (default false, inherited through `extends`, true only for GGUF), and `main()` hard-errors naming flag + profile + reason before the profile view is wrapped. (Checked: `5f7072f` has no dependence on `e5c4993`, so both riders stack directly on this PR, independent of each other.)

Both are drafts pending your look; flip them to ready at will. On your validation ask: the with/without `--no-packed-aggregation` re-solve on the 27B/35B pairs + `validate_final_serving_promotion_noop` + vLLM eager/graph load smoke is queued on our side before any of this influences a ship, and #21 gets its own GGUF-lane load smoke if the flag is ever used in anger.

Tests: core branch 117 passed; #20's branch 158 + 35 KL-lane passed; #21's branch 114 passed.

RobTand:
**Merged.** The core refactor was the best work in the stack and it went in as-is.

**The revert-based split verified mechanically,** which was my main concern with that approach. `merge-base pr/16 pr/17 = 0aa9df2`, and `git diff 0aa9df2 pr/17` is 3 files (`allocator.py` +67, `allocator_candidates.py` +206, the new test +433) with **zero hits** for `role_split|_PACKED_ROLE_GROUPS|bit_exact|weight_mse == 0|zero dloss`. So a squash of this branch introduces the aggregation core plus `f148feb` and nothing else. (A raw `git diff pr/16 pr/17` *looks* like it reverts your round-2 work only because this branch is stacked on the round-1 tip — stale base, not a revert; `main` now carries both the min-Δloss ratchet and the quadrature.)

**Both minor fixes verified down to the arithmetic.** Member dloss is `(cost_pred + z·stderr)·gain`; aggregation subtracts exactly `z·stderr·gain` and adds `z·√Σ(stderr·gain)²`, with gains resolved by the identical expression and alias loop member pricing uses — so no silent hedge mismatch. At `z == 0` both hedge terms are identically `0.0`, making it bit-for-bit, which your exact member-sum test still pins. The shape fix is right too: I traced **every** `_shape_from_stats` consumer, and no group-16 scale count and no shape-legality mask is derived from the rank-1 view (the legality mask runs pre-aggregation and its other callers never see the aggregated dicts). `_assignment_bits_total` agreement is exact.

**Serving invariants hold after `f148feb`:** one format per group stays structurally unrepresentable otherwise, `promote_serving_units` still runs on the final expanded assignment with the no-op validation, the super-item name can't re-enter a packed group, and packed/fused still skip each other's names.

**On the real 35B probe/cost pair, aggregation is allocation-neutral at `TARGET_BITS`:** 0 of 500 assignments change vs `main`, same achieved bits and Δloss. The `−40` NVFP4 unit count at the identical-Δloss rungs is exactly this PR collapsing 2 packed expert rows per MoE layer into 1 DP item — accounting, not allocation.

**Two things landed on top** (`4bf27ba`):

1. **`aggregate_fused_siblings` still had both bugs `f148feb` fixed for packed groups** — a linear `z·Σstderr` hedge (3× vs √3 on a qkv triple) and no aggregated stderr, with `weight_mse` derived from the hedge-contaminated sum. At `COST_UCB_Z > 0` that would have hedged attention/MLP fused groups ~1.7× harder than independence-correct while packed expert groups were correct, biasing bits *away* from attention — the opposite of this PR's intent. Both paths now share one helper so they can't drift; `z == 0` stays bit-identical.
2. **The empty-intersection fallback's docstring was wrong in both directions.** It claimed promotion "repairs coherence rather than the group silently vanishing"; in fact disjoint member menus make every promoted format illegal for some member, so `compute_achieved` now refuses to price it at every target. Corrected to say that, and to name it an upstream cost/legality bug rather than a state to allocate around.

Your judgment call on omitting per-member features was the right one — there genuinely is no single `(out, in)` for a mixed gate_up/down group, and scaling `out_features` would have been the worse guess.

PR #20, opened 2026-07-29
Allocator: bit-exact re-encode pricing, gated on an identity activation path

tylerwagler:
> Stacked on #17 (`rob-pr5-packed-units`) — merge that first; only the last 2 commits are new here. Split out of #17 per review.

**Commits:** `allocator: bit-exact re-encodes (weight_mse == 0.0) price at zero dloss` (`0164c77`, clean cherry-pick of `e5c4993` as reviewed in #17) · `allocator: bit-exact short-circuit requires an identity activation path` (`2d7cd08`, the review blocker fix)

### Summary
A measured `weight_mse == 0.0` proves W' == W — but as reviewed, the short-circuit priced every such entry at `dloss == 0.0`, which only silences the weight side: `measure_quant_cost` applies `activation_quantize_dequantize(X)` before computing `output_mse`, so for W·A· formats (NVFP4, FP8_DYNAMIC, MXFP4, MXFP8_E4M3) that number is real A-side error. On an MXFP4-packed DSv4 source, an MXFP4 re-encode would have priced at the unbeatable global minimum and the RD curve would claim zero loss for a W4A4 assignment.

The fix gates the shortcut on a dtype-level fact rather than a heuristic: `FormatSpec.act_quant_changes_input`, backed by the existing declaration `act_bits is None` ("no activation quant"), with a registry-wide test pinning `act_bits is None ⟺ the activation callable is the identity` for every registered format. The `cost_entry_is_bit_exact` family takes an optional `format_name` (unknown ⇒ never short-circuit), threaded through `build_candidates`, fused-sibling aggregation, `kl_measurement.l2_cost_value`, `sensitivity_response`, and `propagated_sensitivity_costs`.

Net behavior: BF16 / FP8_SOURCE / NVFP4A16 / MXFP8A16 and other A16 passthrough formats keep the zero-dloss shortcut for bit-identical re-encodes; W·A· formats carry their measured A-side cost. Entries with an explicit `cost_source` (production-render, aura) are exempt as before.

### How to verify
`pytest tests/test_bit_exact_cost_pricing.py tests/test_allocator_packed_group_units.py -q`

### Tests
`tests/test_bit_exact_cost_pricing.py` pins: weight-bit-exact + activation-quantizing format does NOT short-circuit (entry, `build_candidates`, and packed-group level, exercised for every registered W·A· format); weight-bit-exact + passthrough-activation format does; explicit-`cost_source` entries keep their own pricing. Suite on this branch: 158 passed + KL suites 35 passed.


RobTand:
**Merged.** The blocker is genuinely closed, and closed the right way — on a dtype-level declaration rather than a name list or a callable probe.

**I verified the load-bearing equivalence empirically instead of trusting the registry test:** across all 27 registered formats, 6 have `act_bits` absent with a bitwise-identity activation callable (max |Δ| = 0.0) and 21 have `act_bits` set with non-identity callables (max |Δ| 4.8e-2 … 1.8e0). **Zero violations in either direction.** Worth noting the near-miss: BF16/FP8_SOURCE use `lambda x: x.clone()` (value-identical, so `torch.equal` holds), and their `act_bits=16` lives in the `autoround_config` dict, not the FormatSpec.

**The threading is exhaustively correct.** Every call site of the family passes a `format_name` that matches the entry by construction — `build_candidates` and fused aggregation via the same alias discipline, `l2_cost_value` / `sensitivity_response` / `propagated_*` from their own lookup keys — and no site passes a mismatched name. The default is the safe direction, and the packed aggregation from #17 never calls the family (it sums already-gated member candidates).

**Two things landed on top** (`4bf27ba`):

1. **A real behaviour regression from this pair.** For a bit-exact passthrough row, `uses_output_mse` is now `False`, so the L3 propagated penalty is written into `predicted_dloss` — which the short-circuit then discards, while `total_scaled_member_penalty` still reports it applied. Measured: an FP8_SOURCE row with penalty 0.034647 priced at `0.0`, and the summary counted 0.0846 of penalty the DP never saw. Penalized bit-exact rows now carry an explicit `cost_source` **and** `output_mse_measured = False` — the second is required, not belt-and-braces, since `cost_source` alone only defeats the short-circuit, after which the passthrough format's noise-level `output_mse` would outrank `predicted_dloss`. Verified the one value-branching `cost_source` reader (`aura_additivity_gate`) classifies these rows identically before and after.
2. **A predicate convention split.** `act_quant_changes_input` said `act_bits is not None` while four existing consumers said `act_bits is not None and act_bits < 16`. Unified on the property, with `act_bits >= 16` deliberately meaning passthrough — the spelling `format_registry`'s own `autoround_config` dicts already use — so a future A16 rung declared the natural way can't be mispriced *or* turn your registry test red. Behaviour-identical today; no registered format declares `act_bits >= 16`.

**Two scope limits I want on the record, since neither is visible from the PR text.** First: the gate closes the `output_mse` branch, so "W·A· formats carry their measured A-side cost" holds **only where `output_mse` was actually measured**. On packed-expert rows with unmeasured `output_mse` — including under `PRISMAQUANT_EXPERT_COST_SAMPLE` — a weight-lossless W4A4 re-encode still prices at exactly 0.0 through the pre-existing weight-only fallback, on the same DSv4/MXFP4-packed model class that motivated the rejection. That's pre-existing rather than a regression here, but it means the pathology isn't fully gone. Second: the A16 rungs that keep the shortcut (NVFP4A16, MXFP8A16, INT-W·A16) are **not exportable** — `export_native_compressed` raises for anything outside {NVFP4, MXFP8_E4M3/E5M2, FP8_E4M3/E5M2, MXFP4, BF16} — while `vllm_packed_moe`'s dense rule only denies three of them, so the DP is now mildly attracted to a rung that hard-fails at export. Loud, not silent. Both filed as follow-ups.

PR #21, opened 2026-07-29
Allocator: opt-in gate_up/down role split, hard-gated on per-role serving-profile support

tylerwagler:
> Stacked on #17 (`rob-pr5-packed-units`) — merge that first; only the last 2 commits are new here. Split out of #17 per review; independent of the bit-exact-pricing rider (`5f7072f` has no dependence on `e5c4993` — the cherry-pick applies cleanly without it).

**Commits:** `allocator: opt-in gate_up/down role split for packed expert DP units` (`cf8256f`, clean cherry-pick of `5f7072f` as reviewed in #17) · `allocator: --packed-role-split hard-errors unless the profile declares per-role expert schemes` (`8093cf7`, the review blocker fix)

### Summary
`--packed-role-split` keys expert groups as (layer, {gate_up|down}) instead of per-layer-uniform, halving unit granularity so the DP can fund the valuable role of ~2x as many layers at a fixed budget. As reviewed, its help text claimed it is "legal when the serving lane supports per-role expert schemes" but nothing enforced that: with `--target-profile vllm_packed_moe` it could emit gate_up=NVFP4 with down=FP8 in one MoE layer — a checkpoint vLLM cannot load (`CompressedTensorsMoEMethod` selects one scheme per FusedMoE layer).

Serving profiles now declare an explicit capability, `supports_per_role_expert_schemes` (default false, OR-inherited through `extends`), set true only for the GGUF lane, where expert tensors stack per projection. `require_per_role_expert_scheme_support()` runs in `main()` before the profile view is wrapped and hard-errors naming the flag, the resolved profile, and the reason. Default behavior unchanged; the flag remains opt-in.

### How to verify
`pytest tests/test_allocator_packed_group_units.py tests/test_serving_profiles.py -q`

### Tests
Gate coverage: `--packed-role-split` + gguf accepted; + research / vllm_packed_moe / vllm_qwen3_5_packed_moe / unknown / default all hard-error with the flag and profile named. Role-split atomicity tests ride the cherry-picked commit. Suite on this branch: 114 passed.


RobTand:
**Merged, clean — no changes needed from my side.** This is the only PR in the stack I didn't have to touch.

**The gate is on the only reachable path.** `packed_role_split_profile` has exactly one production caller, immediately after the check, and a repo-wide grep found no env var, config key, seed-json path or library entry point that could enable role splitting without passing through `main()`. It fires before the profile view is wrapped, before `build_candidates`, before any solve, and before the first write — everything between arg-parse and the gate is reads. And the vLLM lane has an independent backstop anyway: `export_native_compressed` hard-fails on `FusedMoE at … has mixed states across packed expert projections`.

**The capability declaration is sound and the GGUF claim is verified in code, not prose.** Default `False`, `from_dict` defaults false, only `gguf.json` sets it true, and none of `research` / `vllm_packed_moe` / `vllm_qwen3_5_packed_moe` extends `gguf`, so OR-inheritance can't leak. I checked the GGUF half rather than taking it: `export_gguf_direct` stacks experts into one 3-D tensor **per (layer, projection)** with one quant type per stacked tensor, and resolves the stacked `gate_proj`/`up_proj` from the allocator's `experts.gate_up_proj` entry and `down_proj` from `experts.down_proj` — exactly the two units the split creates, with no per-layer uniformity requirement. So the lane really can serve what the flag emits.

**The re-landed mechanics are atomic.** Role keying suffixes the layer-scoped key with `::role:{gate_up|down}`, and `promote_serving_units` keys through the *same wrapped* profile, so union-find never merges the two role units and promotion stays a genuine no-op. The role vocabulary covers every expert leaf name in today's structure specs, and an unknown leaf falls back to the layer key, which is the safe direction. On packed-param models each role group has one member and falls to individual rows — same 2-units-per-layer granularity, still coherent.

Two nits recorded as follow-ups, neither worth holding this: `packed_projection_role_group` parses model names with a hardcoded projection table inside the allocator, duplicating the profile accessor and crossing the boundary `model_profiles/base.py` states explicitly ("the solver asks the profile for groups; it does not parse model names itself") — complete for today's archs, but a new arch with different expert leaves silently loses the split instead of erroring. And `--packed-role-split` leaves no trace in any persisted artifact, so a downstream consumer can't distinguish an intentional role split from a coherence bug, and re-promoting a saved role-split assignment with the unwrapped profile would silently collapse the two roles.

If the flag is ever used in anger, your own note stands: it needs the GGUF-lane load smoke.

PR #44, opened 2026-08-01
Release PrismaQuant 0.5.1 with versioned NVFP4 activation contracts

RobTand:
## Summary

- add a versioned, digest-bound NVFP4 W4A4 activation contract and calibrated per-target static scales
- make resident and streaming export use the same target namespace, sidecar accounting, and fused-sibling scale policy
- consolidate activation-scale policy in one owner so native and CB export cannot drift
- prepare the 0.5.1 package and release metadata

Fused serving remains explicit opt-in. This release establishes artifact eligibility and fail-closed provenance; it does not claim served-quality promotion or enable fused NVFP4 by default.

## Validation

- full CPU suite: 2,530 passed, 85 skipped, 2 xfailed, plus 145 subtests
- focused activation/accounting suite: 286 passed, plus 134 subtests
- `uv lock --check`
- wheel and sdist build
- strict Twine validation
- 23/23 required runtime assets in both artifacts
- clean non-editable installed-wheel profile/spec/table/helper/CLI gate

The regenerated lock also fixes the inherited Python-floor and test-extra mismatch that made the prior lock fail `uv lock --check`.


RobTand:
Reopened immediately: the comment above was posted here by mistake — it belongs to **tessera#44** (trailing partial superblock load cost). A `cd` into this repo carried over into the `gh` call. Nothing about PrismaQuant 0.5.1 was measured, decided, or changed; this issue's own state is untouched. Apologies for the noise.

PR #58, opened 2026-08-03
Declare the block-FP8 floor instead of dropping its scale planes

RobTand:
## The bug (present in artifacts already on disk)

A block-FP8 weight that **no allocation target claimed** reached the verbatim copy
loop, which was written for BF16 norms and buffers. For those, "copy the tensor and
list it in `ignore`" is right. For a block-FP8 weight it was wrong twice, and the two
errors compounded into silence:

- the loop skips `.scale` siblings as *"consumed with their fp8 weight"* — but nothing
  consumed this one, because the weight was never a target. **The scale plane was dropped.**
- the weight was then listed in `ignore`, i.e. declared to be plain unquantized floats.

A consumer honouring that allocates a bf16 parameter, **passes the size assertion**
because the element counts agree, and casts the fp8 bytes **with no scale applied**.
Every 128×128 block ends up wrong by its own power of two. Nothing raises.

On DSv4-Flash: **43 × `attn.wo_a` + 21 × `attn.indexer.wq_b`, 1.44 GB**. Confirmed in
artifacts already on disk — `export-smoke/artifact_attn_FP8_CB_K36` ships
`layers.0.attn.wo_a.weight` as F8_E4M3 with **no `.scale` beside it** and the unit in
`ignore`, while the source checkpoint has the scale.

## The fix

A floor block-FP8 unit ships its weight **and** its scale, and is declared
`fp8_e4m3_ue8m0_block128` — the same delegated-native route an *allocated* passthrough
unit of that format already gets. Nothing about the bytes differs between the two cases;
only whether the DP happened to choose the unit, which is not a property the consumer
can see or should have to.

Membership is narrow **by construction, not by name**: an F8_E4M3 weight paired with an
F8_E8M0 scale over the checkpoint's declared 128×128 block. That excludes MXFP4
nibble-packs (int8 elements) and BF16/F32 verbatim tensors (no scale plane — the
`ignore` cast IS correct for them). A non-128×128 declared block is a **refusal**, not a
silent skip: declaring the wrong grid loads, and is wrong.

`mtp.*` stays verbatim-undeclared per the DSpark audit — vLLM 0.24 builds no module for
those blocks, so declaring them would assert a route nobody exercises.

## Route-pending, honestly recorded

No route-table edits — main's three-state record stands. These units face the **same**
ship gate as an allocated passthrough and record the **same** provenance
acknowledgement. One asymmetry is worth naming: for an allocated rung, refusing means
the DP picks something else; for a floor unit **there is no something else**, so the
refusal is loud rather than a quiet downgrade back to `ignore`.

`PQ_ALLOW_ROUTE_PENDING=1` is an env spelling of `--allow-route-pending-passthrough` for
driver scripts. Off for every value except exactly `"1"`, and it announces itself on
stderr — the one failure mode an env knob has that a flag does not is being inherited by
an invocation nobody meant to acknowledge for.

## The gate (the part that matters)

`prismaquant/artifact_completeness.py` reads a **published** artifact — safetensors
headers and `quant_config.json`, never the source — and asks the only question a
consumer can ask: *for every scale-bearing weight actually on disk, which mechanism
decodes it, and is that mechanism complete?* Four failure modes: fp8-in-`ignore`,
declared-but-missing-scale, orphan scale plane, claimed-by-nothing.

It runs at the end of **every** export and inside the test helper **every** DSv4 fixture
goes through, so this class now fails in whichever test produced it rather than only in
a test written to look for it. Headers only — a few file opens even on 92 GB. Also
runnable standalone:

```
python -m prismaquant.artifact_completeness /path/to/artifact
```

Pointed at the two smoke artifacts on disk, it reports exactly the bug above.

## Byte accounting

Block-128 means **one scale byte per 16,384 weight elements**, so the 64 restored scale
planes are ~90 KB-class against the 1.06 MB of headroom the 92 GB selection has. Asserted
against the emitted header rather than argued.

## Namespace note (surfaced, not fixed here)

Units are spelled in several legitimate vintages at once — recipe
(`model.layers.0.self_attn.wq_a`) for allocator-chosen units, checkpoint
(`layers.0.attn.wq_a`) for exporter-only ones, and the collapsed shared-MLP form that
config-group targets use. The gate normalizes through the profile's own
recipe→checkpoint map rather than demanding internal uniformity. Making the producer
emit one spelling everywhere is a real improvement and a **separate** change; a gate
that failed every mixed-vintage artifact would be useless for finding this bug.

Worth a follow-up: an allocated DSv4 passthrough unit would be declared
`model.layers.N.self_attn.wq_a`, which does **not** match vLLM's serving prefix
`model.layers.N.attn.wq_a`. Latent today (no DSv4 artifact has shipped passthrough
units), but it will bite the first one that does.

## Tests

Full suite green locally (2771 passed). New coverage: floor unit ships both planes
byte-verbatim + declared + absent from `ignore`; the completeness gate catches the
original shape, a declared-but-scaleless unit, and passes healthy artifacts; the env
knob defaults off and needs exactly `"1"`; unacknowledged export refuses loudly.

One pre-existing test needed the acknowledgement added — it was silently producing a
reference artifact **with** this bug, and only passed because it compared CB tensors.

Do not merge without review.

https://claude.ai/code/session_01HUHqNEHMu7FAJQSNGCZhqW

### DIFF
--- a/prismaquant/artifact_completeness.py
+++ b/prismaquant/artifact_completeness.py
@@ -90,6 +90,10 @@ class CompletenessReport:
 
     #: route-pending formats the producer explicitly acknowledged shipping
     route_pending_acknowledged: list[str] = field(default_factory=list)
+    #: namespaces the producer recorded as deliberately OMITTED. An absence
+    #: covered by one of these is intentional; an absence not covered by one is
+    #: a dropped tensor, which is the failure this module exists to catch.
+    excluded_namespaces: list[str] = field(default_factory=list)
 
     @property
     def ok(self) -> bool:
@@ -253,10 +257,25 @@ def check_artifact_completeness(
         (quant_config.get("provenance") or {}).get(
             "route_pending_passthrough_acknowledged") or ())
 
+    excluded = [
+        str(prefix) for prefix in
+        ((quant_config.get("provenance") or {}).get("excluded_namespaces")
+         or ())
+    ]
     report = CompletenessReport(
         declared_units=dict(raw_declared),
         route_pending_acknowledged=acknowledged,
+        excluded_namespaces=excluded,
     )
+    # A declared unit whose tensors are absent because its namespace was
+    # excluded is an intended absence, not a broken promise. Drop those
+    # declarations before checking, so "declared but no tensor" keeps meaning
+    # the one thing that IS a bug.
+    if excluded:
+        declared = {
+            unit: wire for unit, wire in declared.items()
+            if not any(unit.startswith(prefix) for prefix in excluded)
+        }
 
     def _scale_unit(name: str) -> str | None:
         for suffix in _SCALE_SUFFIXES:
--- a/prismaquant/artifact_completeness.py
+++ b/prismaquant/artifact_completeness.py
@@ -1,0 +1,417 @@
+"""Post-export gate: every quantized weight in an artifact has a decoder.
+
+THE FAILURE CLASS THIS EXISTS TO MAKE IMPOSSIBLE. An exported artifact is a
+promise that each tensor can be read back. The promise is carried by three
+different mechanisms — a CB config group, a ``source_passthrough`` declaration,
+or an ``ignore`` entry meaning "plain unquantized floats" — and nothing used to
+check that every tensor is covered by exactly one of them, correctly.
+
+A block-FP8 weight that no allocation target claimed fell through all three: it
+was copied verbatim (right), its ``.scale`` sibling was skipped as "consumed
+with its weight" though nothing consumed it (wrong — the scale was DROPPED),
+and it was listed in ``ignore`` (wrong — it is not unquantized). A consumer
+honouring that reads fp8 bytes into a bf16 parameter, passes the size check
+because the element counts agree, applies no scale, and serves weights that are
+each off by their own power of two. Nothing raises. On DSv4-Flash that was 43
+``attn.wo_a`` + 21 ``attn.indexer.wq_b`` units — 1.44 GB, silently wrong.
+
+So this module asks the question the exporter's own asserts could not: **for
+every scale-bearing weight tensor actually present in the artifact, which
+mechanism decodes it, and is that mechanism complete?** It reads only the
+artifact — safetensors headers and ``quant_config.json``, never the source
+checkpoint — because that is exactly what a consumer has.
+
+Cheap by construction: safetensors headers only, no tensor data, so it runs on
+a 92 GB artifact in about the time it takes to open the shards.
+
+Run standalone::
+
+    python -m prismaquant.artifact_completeness /path/to/artifact
+"""
+from __future__ import annotations
+
+import json
+import struct
+from dataclasses import dataclass, field
+from pathlib import Path
+
+
+__all__ = [
+    "ArtifactIncomplete",
+    "CompletenessReport",
+    "check_artifact_completeness",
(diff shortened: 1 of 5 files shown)


RobTand:
## Second commit: namespace-exclusion knob (`c86dd90`)

Adds `--exclude-namespace` (repeatable) / `PQ_EXPORT_EXCLUDE_NAMESPACES` (comma-separated),
**default empty — zero behaviour change unless set**. Omits matching tensors from the
artifact entirely: no tensor, no index entry, no `ignore` line, no declaration. Built for
the pending decision that may ship without `mtp.*` (10.86 GB back to the allocator).

The prefix test runs on the raw checkpoint key inside the verbatim-copy fallback, so a
unit's scale and other companions leave **with** it rather than by a second rule that
could disagree. Excluded units are also withheld from the floor block-FP8 declaration
pass, so exclusion and declaration can never both claim the same tensor.

**Excluding an allocated unit is a hard refusal**, checked in *both* namespace vintages
(`mtp.` and `model.mtp.`) — a prefix that missed only because it was written in the other
spelling would be a silent no-op, the worst outcome available. Rationale: the DP priced
allocated units and the selection's achieved bits and predicted loss assume they ship.

The omission is recorded as `provenance.excluded_namespaces`, and the completeness gate
reads it, so *absent-on-purpose* and *dropped-by-a-bug* stay distinguishable from the
artifact alone. The exemption is the **record**, not the prefix — pinned by a test that
strips the record from an otherwise byte-identical artifact and demands the failure come
back.

Also threads `-e PQ_ALLOW_ROUTE_PENDING` and `-e PQ_EXPORT_EXCLUDE_NAMESPACES` through the
driver's `DOCKER_COMMON` as **value-less** forwards (host value when set, absent when not),
so the launch wrapper decides per invocation. `-e NAME=` would define them empty *inside*
the container — a different statement, and the wrong one for an acknowledgement flag.

Tests: `mtp.*` gone from tensors/index/config, byte total drops by exactly the namespace's
measured size, everything else byte-identical by SHA-256; env defaults to nothing; both
refusal spellings; gate exemption. Full suite green locally (**2776 passed**), 4/4 CI green.

---

### ⚠️ Consumer-binding finding — affects the DP variant choice, not this PR

Executed against gridbook master's real `_canonical_prefix` / `_candidate_bases`
(`_passthrough_format` is a plain exact lookup over those, `config.py:867-873`):

| declaration spelling | served prefix | binds? |
|---|---|---|
| `model.layers.N.self_attn.wq_a` (recipe) | `model.layers.N.attn.wq_a` | ❌ |
| `layers.N.attn.wq_a` (checkpoint) | `model.layers.N.attn.wq_a` | ✅ |
| `layers.N.attn.wq_a` / `.wkv` | `model.layers.N.attn.fused_wqa_wkv` | ❌ |

Two independent gaps: there is **no `self_attn`↔`attn` normalization anywhere in gridbook**,
and `_passthrough_format` does **no fused-shard resolution at all** — unlike `_is_ignored`
(`:747`) and `shard_target_keys` (`:782`), which do consult `_FUSED_FALLBACK`. vLLM merges
`wq_a`+`wkv` into one `fused_wqa_wkv` module (`deepseek_v4/attention.py:199`), so those two
units cannot bind under **either** spelling.

**This PR's floor declarations use the checkpoint spelling and target unfused modules
(`attn.wo_a`, `attn.indexer.wq_b`), so they bind correctly.**

PR #80, opened 2026-08-12
Fix hybrid linear-attention masking in layer streaming for Qwen3.5/3.6 on transformers >= 5.15

TechPrototyper:
Hi Rob — first: thanks for AURA. We've been running PrismaQuant as the production quantization method for dense Qwen3.6 on a DGX Spark (GB10) and an RTX 5090, and the KL-Fisher allocation has held up to every measurement we've thrown at it. This PR fixes the one thing that broke when our environment resolved a newer transformers than your lockfile pins.

**What breaks**

With `transformers>=5.15.0` and a hybrid model whose `config.layer_types` mixes `full_attention` with `linear_attention` (Qwen3.5/3.6 GatedDeltaNet), the layer-streaming forward crashes. Two bugs in the same code path, found in sequence because the second was masked by the first:

1. `_compute_attention_mask()` only special-cases `sliding_attention` (Gemma3/4). Any other hybrid falls through to a single dense `[1,1,T,T]` causal mask fed to every layer — including `linear_attention` layers, whose `apply_mask_to_padding_states()` expects the raw 2D `[batch, seq_len]` padding mask (or `None` for an unpadded batch), same as HF's own `Qwen3_5TextModel._update_linear_attn_mask()`. The recurrence is already causal by construction; the dense mask broadcasts against `hidden_states` and crashes with a trailing-dim mismatch (hidden_size vs. seq_len). Thematically related to transformers#47087 (padding-mask correctness for linear-attention hybrids on cache continuation) — different failure mode, same subsystem.

2. Fixing (1) exposed the second: `_layer_attention_type()`'s attribute probe returns `None` for `linear_attention` layers on transformers >= 5.15.0. Upstream transformers#47630 ("Refactor all linear attn models & native kernels fallback", merged 2026-08-05) renamed the attribute `Qwen3_5DecoderLayer` stores from `.layer_type` to `.block_type` (confirmed via source diff, 5.14.1 vs 5.15.0), and these layers carry no `self_attn`/`attention` for the idx-fallback to key off — so `_call_layer`'s "don't silently guess" guard raises.

**The fix** (`layer_streaming.py` + two test files; reworked after review)

- `_compute_attention_mask()` branches independently on `has_sliding` / `has_linear` (plus `conv` and declared non-attention block types) and routes `linear_attention`/`conv` through the current (transformers >= 5.15) recurrent-mask contract, always via a local shim (`_recurrent_padding_mask`) — deliberately never the upstream helper, whose 5.13/5.14 incarnation ships the pre-fix cache-state contract. Contract: `None` for missing/non-2D/single-token/all-ones cases, growing continuation masks trimmed to the local sequence, contiguous result. Declared `moe`/`mlp` blocks receive `None` (upstream `.get()` parity, explicit allowlist); anything else unresolvable stays fail-closed.
- `_layer_attention_type()` adds `layer.block_type` to the primary probe and resolves the Qwen3.5/3.6 recurrent child via the existing lookups (`linear_attn.layer_type`, `linear_attn.layer_idx` + `config.layer_types`). No structural guessing; unknown layers keep failing closed.
- Regression tests for both halves, runnable under 5.8 (fallback arm) and >= 5.15 (helper arm), plus an integration file against real `Qwen3_5` modules.

Your `uv.lock` pins `transformers==5.8.0`, where the rename hasn't happened and the bug stays latent, so stock environments are unaffected. `pyproject.toml` allows `>=5.0` with no upper bound, though, so any fresh resolution hits it. `layer_streaming.py` is byte-identical between v0.10.0 and v0.11.0; the patch was developed against 0.10.0 and re-verified against this branch's base (branch since rebased onto 0.12.0 main).

**Validation**

Beyond reproducing and clearing the crash, we ran the patched pipeline end-to-end: Qwen3.6-27b dense quantized with AURA (mixed NVFP4-MLP + FP8-attn + NVFP4-KV, ~5.5 bpp) under transformers 5.15.0 + this patch, then a paired battery against your `prismaaura55` reference, both served sequentially on the same GB10:

- GSM8K, n=250 screening: 97.2% vs 95.6% (Δ +1.6 pp; McNemar exact p=0.34 — statistical parity, discordant 7/3)
- long-context needle: 9/9 both; determinism: 5/5 both

So the patched masking path reproduces your reference quality at screening level; the full n=1319 run is queued and I'll drop the result here when it lands. Write-up with raw JSON: **[local-ai-lab write-up](https://github.com/TechPrototyper/local-ai-lab/blob/master/notes/qwen36-aura-head-to-head.md)**.

Happy to add a regression test for the layer-type resolution if you'd like one, or reshape anything to fit the codebase's conventions. And genuinely: AURA is doing on our hardware exactly what the method promises — glad to be building on it.

### DIFF
--- a/tests/test_qwen35_linear_attention_integration.py
+++ b/tests/test_qwen35_linear_attention_integration.py
@@ -1,0 +1,135 @@
+"""Integration: streaming forward through REAL Qwen3.5 hybrid layers.
+
+The fakes in test_linear_attention_mask_routing.py pin the contract; this
+file pins it against transformers' actual Qwen3_5 modules (tiny random-weight
+hybrid, CPU) — the family whose transformers>=5.15 refactor (`.block_type`
+rename, `apply_mask_to_padding_states` without a shape guard) motivated the
+fix. The forward-path tests run on every transformers version that ships
+qwen3_5 (the locked 5.8 environment included — same contract, pre-refactor
+attribute names); the two assertions specific to the 5.15 refactor
+(`.block_type`/child `layer_type`, guardless ``apply_mask_to_padding_states``)
+are version-gated.
+"""
+import pytest
+import torch
+
+pytest.importorskip(
+    "transformers.models.qwen3_5",
+    reason="qwen3_5 modeling not available in this transformers version",
+)
+
+import transformers
+from packaging.version import parse as _parse_version
+
+_POST_REFACTOR = _parse_version(
+    transformers.__version__) >= _parse_version("5.15.0")
+
+from transformers.models.qwen3_5 import Qwen3_5TextConfig
+from transformers.models.qwen3_5.modeling_qwen3_5 import (
+    Qwen3_5DecoderLayer,
+    Qwen3_5TextRotaryEmbedding,
+)
+
+from prismaquant.layer_streaming import (
+    _call_layer,
+    _compute_attention_mask,
+    _layer_attention_type,
+)
+
+
+def _tiny_hybrid():
+    cfg = Qwen3_5TextConfig(
+        hidden_size=64, intermediate_size=128, num_hidden_layers=2,
+        num_attention_heads=4, num_key_value_heads=2, head_dim=16,
+        layer_types=["linear_attention", "full_attention"],
+        linear_num_value_heads=4, linear_num_key_heads=2,
+        linear_key_head_dim=16, linear_value_head_dim=16,
+        linear_conv_kernel_dim=4,
+        vocab_size=128, max_position_embeddings=64,
+    )
+    cfg._attn_implementation = "eager"
+    torch.manual_seed(0)
+    layers = [Qwen3_5DecoderLayer(cfg, i) for i in range(2)]
+    rotary = Qwen3_5TextRotaryEmbedding(cfg)
+
+    class _Base(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.config = cfg
+
+    return cfg, _Base(), layers, rotary
+
+
+def _stream(base, layers, rotary, hidden, attention_mask):
+    position_ids = torch.arange(hidden.size(1)).unsqueeze(0).expand(
+        hidden.size(0), -1)
+    masks = _compute_attention_mask(base, hidden, position_ids,
+                                    attention_mask=attention_mask)
+    pe = rotary(hidden, position_ids)
+    out = hidden
+    for layer in layers:
+        out = _call_layer(layer, out, position_embeddings=pe,
+                          attention_mask=masks, position_ids=position_ids)
+    return masks, out
+
+
+def test_real_layers_resolve():
+    # Passes pre- and post-refactor: `.layer_type` on 5.8, `.block_type`
+    # (plus the recurrent child's own attributes) from 5.15 on.
+    _, _, layers, _ = _tiny_hybrid()
+    assert _layer_attention_type(layers[0]) == "linear_attention"
+    assert _layer_attention_type(layers[1]) == "full_attention"
+
+
+@pytest.mark.skipif(not _POST_REFACTOR,
+                    reason="`.block_type` rename landed in transformers 5.15")
+def test_post_refactor_child_carries_fallback_attributes():
+    _, _, layers, _ = _tiny_hybrid()
+    assert layers[0].block_type == "linear_attention"
+    assert layers[0].linear_attn.layer_type == "linear_attention"
+    assert layers[0].linear_attn.layer_idx == 0
+
+
+def test_streaming_forward_left_padded_batch():
+    _, base, layers, rotary = _tiny_hybrid()
+    hidden = torch.randn(2, 6, 64)
+    pad = torch.tensor([[0, 0, 1, 1, 1, 1],
+                        [1, 1, 1, 1, 1, 1]])
+
+    masks, out = _stream(base, layers, rotary, hidden, pad)
+
+    lin = masks["linear_attention"]
+    assert lin is not None and lin.ndim == 2 and torch.equal(lin, pad)
+    assert out.shape == hidden.shape
+    assert bool(torch.isfinite(out).all())
+
+
+def test_streaming_forward_unpadded_batch():
+    _, base, layers, rotary = _tiny_hybrid()
+    hidden = torch.randn(1, 6, 64)
+
+    # all-ones → linear_attention mask is None; forward must still work
+    masks, out = _stream(base, layers, rotary, hidden, torch.ones(1, 6))
+
+    assert masks["linear_attention"] is None
+    assert out.shape == hidden.shape
+    assert bool(torch.isfinite(out).all())
+
+
+@pytest.mark.skipif(
+    not _POST_REFACTOR,
+    reason="pre-5.15 apply_mask_to_padding_states still shape-guards "
+           "non-2D masks, so the dense mask is silently ignored there")
+def test_dense_mask_crashes_real_linear_layer():
+    # Documents the pre-fix failure mode this branch removes: a dense
+    # [1, 1, T, T] additive mask fed to a real GatedDeltaNet layer
+    # broadcasts against hidden_states and raises a trailing-dim mismatch.
+    _, _, layers, rotary = _tiny_hybrid()
+    hidden = torch.randn(1, 6, 64)
+    position_ids = torch.arange(6).unsqueeze(0)
+    pe = rotary(hidden, position_ids)
+    dense = torch.zeros(1, 1, 6, 6)
+    with pytest.raises(RuntimeError, match="must match the size"):
+        layers[0](hidden_states=hidden, attention_mask=dense,
+                  position_ids=position_ids, past_key_values=None,
+                  use_cache=False, position_embeddings=pe)
--- a/tests/test_qwen35_linear_attention_integration.py
+++ b/tests/test_qwen35_linear_attention_integration.py
@@ -2,9 +2,9 @@
 
 The fakes in test_linear_attention_mask_routing.py pin the contract; this
 file pins it against transformers' actual Qwen3_5 modules (tiny random-weight
-hybrid, CPU) — the family whose transformers>=5.15 refactor (`.block_type`
-rename, `apply_mask_to_padding_states` without a shape guard) motivated the
-fix. The forward-path tests run on every transformers version that ships
+hybrid, CPU) — the family whose transformers hybrid refactors (`.block_type` rename in
+5.13, removal of the `apply_mask_to_padding_states` shape guard in 5.15)
+motivated the fix. The forward-path tests run on every transformers version that ships
 qwen3_5 (the locked 5.8 environment included — same contract, pre-refactor
 attribute names); the two assertions specific to the 5.15 refactor
 (`.block_type`/child `layer_type`, guardless ``apply_mask_to_padding_states``)
@@ -21,8 +21,9 @@ pytest.importorskip(
 import transformers
 from packaging.version import parse as _parse_version
 
-_POST_REFACTOR = _parse_version(
-    transformers.__version__) >= _parse_version("5.15.0")
+_TV = _parse_version(transformers.__version__)
+_HAS_BLOCK_TYPE = _TV >= _parse_version("5.13.0")   # .block_type + child attrs
+_GUARDLESS_MASK = _TV >= _parse_version("5.15.0")   # padding-mask shape guard removed
 
 from transformers.models.qwen3_5 import Qwen3_5TextConfig
 from transformers.models.qwen3_5.modeling_qwen3_5 import (
@@ -35,6 +36,30 @@ from prismaquant.layer_streaming import (
     _compute_attention_mask,
     _layer_attention_type,
 )
+
+
+@pytest.fixture(autouse=True)
+def _force_pure_torch_kernels(monkeypatch):
+    """Keep the CPU integration CPU-safe in CUDA-toolkit environments.
+
+    transformers selects `causal_conv1d`/`fla` extensions by package
+    availability, not tensor device — in a cu130 stack the installed CUDA
+    extension gets picked for CPU tensors and fails with
+    `Expected x.is_cuda()`. The kernel-dispatch wrappers keep the pure
+    PyTorch implementation reachable via `__wrapped__`; unwrap them so the
+    test always runs the torch path, everywhere. No-op in clean CPU envs
+    and on transformers versions without those wrappers."""
+    import transformers.models.qwen3_5.modeling_qwen3_5 as m
+    for name in ("causal_conv1d_fn", "causal_conv1d_update",
+                 "chunk_gated_delta_rule", "recurrent_gated_delta_rule"):
+        fn = getattr(m, name, None)
+        if fn is None:
+            continue
+        inner = fn
+        while hasattr(inner, "__wrapped__"):
+            inner = inner.__wrapped__
+        if inner is not fn:
+            monkeypatch.setattr(m, name, inner)
 
 
 def _tiny_hybrid():
@@ -74,15 +99,15 @@ def _stream(base, layers, rotary, hidden, attention_mask):
 
 
 def test_real_layers_resolve():
-    # Passes pre- and post-refactor: `.layer_type` on 5.8, `.block_type`
-    # (plus the recurrent child's own attributes) from 5.15 on.
+    # Passes pre- and post-refactor: `.layer_type` up to 5.12, `.block_type`
+    # (plus the recurrent child's own attributes) from 5.13 on.
     _, _, layers, _ = _tiny_hybrid()
     assert _layer_attention_type(layers[0]) == "linear_attention"
     assert _layer_attention_type(layers[1]) == "full_attention"
 
 
-@pytest.mark.skipif(not _POST_REFACTOR,
-                    reason="`.block_type` rename landed in transformers 5.15")
+@pytest.mark.skipif(not _HAS_BLOCK_TYPE,
+                    reason="`.block_type` rename landed in transformers 5.13")
 def test_post_refactor_child_carries_fallback_attributes():
     _, _, layers, _ = _tiny_hybrid()
     assert layers[0].block_type == "linear_attention"
@@ -117,7 +142,7 @@ def test_streaming_forward_unpadded_batch():
 
 
 @pytest.mark.skipif(
-    not _POST_REFACTOR,
+    not _GUARDLESS_MASK,
     reason="pre-5.15 apply_mask_to_padding_states still shape-guards "
            "non-2D masks, so the dense mask is silently ignored there")
 def test_dense_mask_crashes_real_linear_layer():
--- a/tests/test_qwen35_linear_attention_integration.py
+++ b/tests/test_qwen35_linear_attention_integration.py
@@ -44,14 +44,34 @@ def _force_pure_torch_kernels(monkeypatch):
 
     transformers selects `causal_conv1d`/`fla` extensions by package
     availability, not tensor device — in a cu130 stack the installed CUDA
-    extension gets picked for CPU tensors and fails with
-    `Expected x.is_cuda()`. The kernel-dispatch wrappers keep the pure
-    PyTorch implementation reachable via `__wrapped__`; unwrap them so the
-    test always runs the torch path, everywhere. No-op in clean CPU envs
-    and on transformers versions without those wrappers."""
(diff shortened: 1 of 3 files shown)


RobTand:
Glad you're enjoying it! Will try to fix your PR asap.

RobTand:
Also: it's just Rob Tand -- not tandler :)

RobTand (changes requested):
Thanks for the detailed reproduction and measurements—the underlying bug is real, and adding `block_type` is the right compatibility direction. I don't think this version is safe to merge yet.

1. Please route `linear_attention` through Transformers' recurrent-mask contract, not the raw incoming mask. Transformers 5.14.1 and 5.15 call `create_recurrent_attention_mask(**mask_kwargs)`. That helper is intentionally not equivalent to `attention_mask`: it rejects non-2D masks, returns `None` for single-token decode and all-ones input, trims a growing/cache-continuation mask to the current local sequence length, and makes the result contiguous. In 5.15, `apply_mask_to_padding_states` no longer has the shape guard described in this patch, so a raw growing mask can itself cause a length mismatch. Please call the upstream helper when available and provide a version-compatible fallback for PrismaQuant's locked Transformers 5.8 environment with the same observable contract.

2. Please remove the structural `hasattr(layer, "self_attn") -> "full_attention"` guess. It converts the current fail-closed behavior for an unknown hybrid layer into a silent full-attention assumption and is not a no-op for unrelated models. `block_type` directly fixes the outer-layer rename. For a fallback, include `linear_attn` in the existing child attribute/index/config lookup—Qwen's recurrent submodule already carries `layer_type`/`layer_idx`—rather than inferring semantics from mere attribute presence.

3. This shared streaming-forward change needs regression tests before merge. Please cover at least: `.block_type` resolution; legacy `.layer_type`; a `linear_attn` child fallback; left-padded 2D mask routing; all-ones -> `None`; current-sequence trimming for a longer continuation mask; single-token -> `None`; and non-regression of the existing sliding/full mapping. Also pin that an otherwise unknown `self_attn` layer still fails closed.

Take another crack (or have your agent take another crack) and address these issues and then we can merge and hopefully fix what's ailing you.

TechPrototyper:
Thanks for the thorough review — you were right on all three counts, and the rework follows your prescription exactly. (And apologies for the surname, got it just from an obviously not too well functioning memory...).

**1. Recurrent-mask contract.** `linear_attention` no longer sees the raw incoming mask. `_compute_attention_mask` now routes it through `create_recurrent_attention_mask(**mask_kwargs)` whenever the installed transformers ships it (>= 5.14), and through a new `_recurrent_padding_mask()` fallback on older versions (your locked 5.8) with the same observable contract, mirrored from the v5.15.0 source: `None` for missing/non-2D masks, `None` for single-token decode, `None` for all-ones outside tracing, and a growing cache-continuation mask trimmed to the local sequence, made contiguous. And of course you were right that the raw mask could itself mismatch on continuation — the trim is covered by a dedicated test.

**2. No structural guessing.** The `hasattr` fallback is gone. Instead, `linear_attn.layer_type` joined the existing attribute probe chain and `linear_attn.layer_idx` the `config.layer_types` index fallback — `Qwen3_5GatedDeltaNet` carries both (verified against the 5.15.0 source). An otherwise-unknown layer keeps failing closed in `_call_layer`; a test pins that a bare `self_attn` layer with no resolvable type raises instead of resolving to anything.

**3. Regression tests.** Two files:

- `tests/test_linear_attention_mask_routing.py` (fakes only, same pattern as `test_multilayer_rope_forward.py`): `.block_type` resolution; legacy `.layer_type`; `linear_attn` child fallback (attribute and index paths); left-padded 2D routing; all-ones → `None`; continuation-mask trimming; single-token → `None`; non-2D → `None`; sliding/full mapping non-regression; unknown-layer fail-closed; end-to-end mask selection through `_call_layer`.
- `tests/test_qwen35_linear_attention_integration.py`: the same contract against real `Qwen3_5` modules (tiny random-weight hybrid, CPU) — streaming forward with left-padded and unpadded batches, resolution on real layers, and a version-gated test documenting the dense-mask crash this branch removes. It runs under 5.8 too (pre-refactor attribute names); the two 5.15-refactor-specific assertions skip there.

**Verification.**

- Both files green under transformers **5.8.0** (your lock — fallback arm, 18 passed / 2 version-gated skips) and **5.15.0** (helper arm, 20/20), on macOS/arm64 and on the GB10 (aarch64, torch 2.13.0+cu130, transformers 5.15.0: 20/20).
- Full suite at the rebased base: 3654 passed; the 4 failures in our environment are identical on a clean `origin/main` checkout (pre-existing/platform, unrelated — happy to detail).
- End-to-end: fresh dedicated run of the AURA quantization pipeline on the GB10 (box to itself) against the real Qwen3.6-27B BF16 under transformers 5.15.0 with this branch. The calibration streaming forward that previously died at the very first decoder layer (~30 s in, both failure modes) now walks the **full 64-layer hybrid stack cleanly** — phase-1 forward complete in 194 s (48 linear + 16 full layers, zero errors), phase-2 loss+head backward complete, phase-3 per-layer backward underway when we bounded the run. (The complete pipeline pass lands with the queued n=1319 re-validation.)
- Branch rebased onto current main (0.12.0); `layer_streaming.py` moved upstream since your review, no interaction with this change.

The n=1319 full-battery verdict is still queued as promised. Thanks again for the precision here — the trim contract was a real gap in what I had.

*Disclosure per AGENTS.md: AI-assisted (rework and tests by an agent under my direction); measurements from our hardware.*

RobTand (changes requested):
Thanks — the previous three review points are addressed, and the lookup-only `block_type` / `linear_attn` resolution now looks right. I found one remaining supported-version blocker before this can merge to production.

`_linear_attention_mask()` currently selects the upstream helper based only on whether it imports. That helper first exists in Transformers 5.13, but 5.13.0–5.14.1 still use the pre-fix contract: they return `None` whenever `past_key_values.has_previous_state()`, including a multi-token cached continuation that contains padding, and they do not special-case a single-token step. Transformers 5.15/current instead keeps and trims the 2D mask for a padded multi-token continuation and returns `None` for a single-token step.

I reproduced this in PrismaQuant's known `/home/rob/dq-runs/venvs/prismaquant-hy3` environment (Transformers 5.13.0):

```
pytest tests/test_linear_attention_mask_routing.py \
       tests/test_qwen35_linear_attention_integration.py -q
# 1 failed, 17 passed, 2 skipped
# test_single_token_decode_returns_none: got tensor([[1]])
```

More importantly, with `past_key_values.has_previous_state() == True`, a padded multi-token continuation is silently reduced to `None` under 5.13/5.14 instead of being trimmed, which is the recurrent-state corruption fixed by the newer upstream contract. `pyproject.toml` allows `transformers>=5.0`, so helper presence is not a sufficient compatibility gate.

Please:

1. Make 5.13/5.14 use the 5.15/current observable contract (using the local shim consistently would be the simplest option, or gate only a demonstrably current helper).
2. Add a regression for the old-helper/cache-state case, ideally exercising both single-token and padded multi-token continuation semantics.
3. Correct the compatibility comments/version gates: `.block_type` landed in Transformers 5.13 (5.12 still used `.layer_type`); the relevant 5.15 change was removal of the padding-mask shape guard.

Everything else I reviewed is sound, including fail-closed unknown-layer handling and the current 5.15 mask behavior. Once this focused follow-up is green across the supported versions, I expect we can merge.

RobTand:
Two additional shared-path/test-harness edges surfaced while the main review ran. Please include these in the same revision.

### Non-attention block types in a hybrid schedule

Transformers hybrids can declare non-attention blocks alongside `linear_attention`. For example, the 5.13 Nemotron-H default `config.layer_types` is:

```python
["linear_attention", "moe", "full_attention", "mlp"]
```

The new `_compute_attention_mask()` returns a dict containing `full_attention` and `linear_attention` only (plus the existing optional sliding/DS aliases). `_call_layer()` then rejects a `NemotronHBlock(block_type="moe" | "mlp")` because that key is absent. Upstream intentionally dispatches with `causal_mask_mapping.get(mixer_block.block_type)`, so non-attention blocks receive `None`.

Please preserve that behavior without weakening fail-closed handling for a genuinely unknown attention type, and add a small mixed-schedule regression (linear/full plus a non-attention block). This does not affect Qwen3.5/3.6's two-type schedule, but `layer_streaming.py` is shared infrastructure and currently presents the branch generically.

### Make the real-module CPU integration deterministic

The two streaming-forward integration tests depend on ambient optional packages. In the known cu130 build environment with Transformers 5.15 supplied via a temporary target install, the installed `causal_conv1d` extension is selected for the CPU tensors and both tests fail:

```
test_streaming_forward_left_padded_batch
test_streaming_forward_unpadded_batch
Qwen3_5GatedDeltaNet -> causal_conv1d_fn -> causal_conv1d_cuda
RuntimeError: Expected x.is_cuda() to be true
```

A clean CPU environment passes 20/20, so this is test isolation rather than the mask implementation. Please force the real-module tests onto Transformers' PyTorch fallback (or otherwise neutralize optional CUDA kernels within the fixture) so the advertised CPU integration remains CPU-safe in PrismaQuant's known Docker/venv stacks.

RobTand:
thanks again. as soon as you make the fixes i'll merge + cut a new release for you

tests/test_qwen35_linear_attention_integration.py
```diff
@@ -0,0 +1,160 @@
...
+    transformers selects `causal_conv1d`/`fla` extensions by package
+    availability, not tensor device — in a cu130 stack the installed CUDA
+    extension gets picked for CPU tensors and fails with
+    `Expected x.is_cuda()`. The kernel-dispatch wrappers keep the pure
+    PyTorch implementation reachable via `__wrapped__`; unwrap them so the
+    test always runs the torch path, everywhere. No-op in clean CPU envs
+    and on transformers versions without those wrappers."""
+    import transformers.models.qwen3_5.modeling_qwen3_5 as m
+    for name in ("causal_conv1d_fn", "causal_conv1d_update",
+                 "chunk_gated_delta_rule", "recurrent_gated_delta_rule"):
```
RobTand: The pure-Torch selection needs to handle both API shapes, not just add the 5.15 names:

- Transformers 5.15 calls wrapped globals `torch_chunk_gated_delta_rule` / `torch_recurrent_gated_delta_rule`; unwrap those along with the decorated `causal_conv1d_*` functions.
- Transformers 5.13/5.14 expose raw optional-extension callables with no `__wrapped__`, then bind them onto each DeltaNet instance during construction. With a GPU visible, the current fixture leaves those callables intact and both CPU forwards fail `Expected x.is_cuda()`.

For the pre-5.15 shape, explicitly select its package-provided fallbacks before `_tiny_hybrid()` constructs layers: `causal_conv1d_fn = None` (the forward then uses `self.conv1d`), `causal_conv1d_update = torch_causal_conv1d_update`, `chunk_gated_delta_rule = torch_chunk_gated_delta_rule`, and `fused_recurrent_gated_delta_rule = torch_recurrent_gated_delta_rule`. Retain the wrapper-unwrapping path for 5.15. Please validate with the GPU visible; `CUDA_VISIBLE_DEVICES=''` changes import-time fast-path selection and can hide this issue.

RobTand (changes requested):
The production masking/routing changes now look right: the 5.13/5.14 recurrent-mask contract is normalized locally, the version gates are corrected, and mixed `moe`/`mlp` schedules preserve upstream `None` dispatch while unknown types still fail closed.

One requested test-isolation issue remains on the latest head. On exact Transformers 5.15.0 in PrismaQuant's cu130 environment with optional FLA/CUDA extensions installed, the two real Qwen3.5 CPU forward tests still enter FLA and fail. Transformers 5.15 renamed the relevant module globals to `torch_chunk_gated_delta_rule` and `torch_recurrent_gated_delta_rule`; the fixture only probes the unprefixed pre-5.15 names.

Reproduction:

```text
env CUDA_VISIBLE_DEVICES='' \
  PYTHONPATH='<PR worktree>:<transformers-5.15 target>:<safetensors-0.8 target>' \
  prismaquant-cu130/bin/python -m pytest \
  tests/test_linear_attention_mask_routing.py \
  tests/test_qwen35_linear_attention_integration.py -q
# 2 failed, 21 passed
# Qwen3_5GatedDeltaNet -> torch_chunk_gated_delta_rule -> FLA
# AttributeError: module 'torch.cpu' has no attribute 'device'
```

Please add both 5.15 `torch_...` names to the fixture's unwrap list, retaining the old names for 5.8–5.14. I validated that exact adjustment dynamically: the same cu130/Transformers 5.15 run becomes 23/23 green. The version matrix is otherwise green (5.8: 21+2 skips; 5.13: 22+1; 5.14.1: 22+1; clean CPU 5.15: 23). This is the only remaining blocker I found; after that focused test fix, this should be ready to merge.

RobTand:
Correction/addendum to my latest review after repeating the test with the GPU visible: adding the two Transformers 5.15 `torch_...` names fixes that version, but it is not sufficient for 5.13/5.14. In those releases, `causal_conv1d_fn` / `_update` are raw optional-extension functions without `__wrapped__`, and the decoder binds them to each DeltaNet instance at construction. The current fixture therefore still fails both CPU forwards (`Expected x.is_cuda()`) whenever the GPU/extension fast path is available; setting `CUDA_VISIBLE_DEVICES=''` changes import-time selection and hides it.

I updated the inline thread with the complete version-robust recipe: explicitly select the pre-5.15 package fallbacks before layer construction (`causal_conv1d_fn = None`, `_update = torch_causal_conv1d_update`, chunk/fused recurrent globals = their `torch_...` fallbacks), while retaining wrapper unwrapping for the 5.15 decorated globals. This remains test-only; I found no unresolved production-code issue.

tests/test_qwen35_linear_attention_integration.py
RobTand: Fixed in `85d40d5`. The fixture now selects all pre-5.15 native Torch paths before layer construction (`causal_conv1d`, both delta-rule functions, and `FusedRMSNormGated`) and unwraps the actual four 5.15 dispatch globals. GPU-visible CUDA/FLA matrix: TF 5.8 = 31 passed/2 skips; 5.13 = 32/1; 5.14.1 = 32/1; 5.15 = 33/0. Clean CPU-only 5.15 = 5/5. No production files changed.

TechPrototyper:
First off: thank you — both for the merge-and-release offer and for how fast and thorough these review rounds have been. You reproduced our suite in your own environments, caught the 5.13 contract trap, and then went and fixed the fixture on our branch yourself with a full GPU-visible four-version matrix (`85d40d5`). That's well beyond what a contributor can ask for, and it's genuinely appreciated. The round-2 items are addressed; everything below is on the branch.

**1. Shim consistently, no upstream helper.** `_linear_attention_mask()` is gone; the routing now always applies the local `_recurrent_padding_mask()` implementing the 5.15/current observable contract, on every transformers version — its docstring documents exactly why the upstream helper is never called even when importable (5.13/5.14 ship the pre-fix cache-state contract). Regressions: a booby-trapped test (`test_routing_never_consults_upstream_helper`) fails if the upstream helper is ever consulted while exercising both semantics you named — padded multi-token and single-token — and `test_cached_continuation_contract_is_cache_state_independent` pins that a growing continuation mask is trimmed-and-kept, never `None`-dropped (the shim takes no cache argument at all, by design).

**2. Version facts corrected.** Comments and gates now match your correction: `.block_type` and the recurrent child's attributes landed in 5.13 (`.layer_type` up to 5.12); the 5.15-specific change is the removal of the padding-mask shape guard. The integration gates are split accordingly (`_HAS_BLOCK_TYPE >= 5.13`, `_GUARDLESS_MASK >= 5.15`) — which also brings your 5.13 run from 2 skips to the expected 1.

**3. Non-attention block types.** Declared `moe`/`mlp` blocks receive `None`, mirroring upstream's `.get(block_type)` dispatch — implemented as an explicit allowlist (`_NON_ATTENTION_BLOCK_TYPES`) rather than a "not `*_attention`" heuristic, so fail-closed stays strict for anything else (including upstream types like `minimax_m3_sparse` we don't handle yet — happy to extend per family). `conv` joins `linear_attention` on the recurrent-mask route, per the upstream mapping. Regression: your Nemotron-H schedule end-to-end through `_call_layer`, including the fail-closed counter-case.

   While pinning this we found one adjacent gap you hadn't flagged: a schedule declaring `moe`/`mlp` *without* any linear/sliding/conv layer would still have taken the single-dense-mask early return — feeding the dense mask to the non-attention blocks. The dict path now also engages for that shape, with a regression (`test_nonattention_schedule_without_recurrent_layers_takes_dict_path`) that also pins the non-regression: a plain all-`full_attention` schedule keeps the bare dense mask.

**4. CPU integration.** Your `85d40d5` is exactly right — thanks for just fixing it rather than bouncing it back; the <=5.14 instance-binding subtlety (and `CUDA_VISIBLE_DEVICES=''` masking the fast-path selection) is a lesson we've filed. We've left your fixture untouched and validated on top of it.

**Verification** (our CPU envs; your GPU-visible CUDA/FLA matrix in `85d40d5` stands on its own):
- transformers 5.8.0 (lock): 22 passed / 2 skipped · **5.13.0: 23 / 1** · **5.14.1: 23 / 1** · 5.15.0: 24 / 0.
- Full suite on the branch: 3661 passed — same 4 pre-existing/platform failures as clean `origin/main`, unrelated.

n=1319 update: the reference arm (prismaaura55, prod serving config) is mid-run; the candidate arm follows overnight — verdict lands here as promised.

*Disclosure per AGENTS.md: AI-assisted (fixes and tests by an agent under my direction); measurements from our hardware.*

RobTand (approved):
Thanks for the careful follow-through—and for catching the adjacent non-attention-only schedule case. I reviewed `b09f968` on top of the focused fixture fix and agree with the implementation: the dict path is narrowly enabled for the explicit `moe`/`mlp` allowlist, all-full schedules retain the prior tensor mask, and unknown types remain fail-closed.

Final local validation is green:
- GPU-visible CUDA/FLA matrix: Transformers 5.8 (32 passed, 2 skipped), 5.13 (33/1), 5.14.1 (33/1), 5.15 (34/34).
- Broader streaming/probe regression selection: 92 passed, 2 skipped.
- `git diff --check` clean.

No further changes requested. Approved for production; I’ll merge once the fresh CI run on this exact head finishes green. The ongoing n=1319 comparison will be useful follow-up evidence, but it is not a blocker for this correctness fix.

RobTand:
Merged to `main` as `f0747a44dfed55361f787360de314fa0a2402e4e`. No further fixes requested—the production path, supported Transformers matrix, broader regression selection, and required CI are all green. Thanks again for the excellent report, measurements, and careful follow-through. Please do share the n=1319 result when it lands; it will be useful follow-up evidence, but it was not a blocker for this correctness fix.

TechPrototyper:
As promised, the n=1319 head-to-head — the full-split follow-up evidence for this fix.

**Setup.** Both arms served with the identical production configuration (same vLLM sm121 image, NVFP4 KV cache, same sampling overrides, concurrency 4), run sequentially on the same GB10 box. Baseline is the shipped `rdtand/Qwen3.6-27B-PrismaAURA-5.5bit` production checkpoint (quantized pre-fix on the ≤5.9 path); candidate is the same source model re-quantized end-to-end on transformers **5.15.0** through the fixed streaming path (as merged here, pipeline pinned to v0.12.2).

| metric | baseline (shipped AURA-5.5) | candidate (5.15 + this fix) |
|---|---|---|
| GSM8K full split, n=1319 | 94.84 % | **95.00 %** |
| McNemar (exact, discordant n=46: +24/−22) | — | p = 0.883 |
| needle | 6/6 | 6/6 |
| tool-calling | 5/6 | 6/6 |
| determinism (greedy repeat) | 4/5 | 5/5 |
| single-stream tok/s | 10.7 | 10.8 |
| batched-8 agg tok/s | 79.2 | 79.5 |

**Reading.** Statistically indistinguishable on GSM8K (the +0.15 pp delta is well within noise), performance identical within ±1 %. The tool-calling and determinism edges are single-digit sample sizes, so I would not claim more than "no regression" there. Net: the fixed 5.15 masking path reproduces the shipped production quality on the full split — zero errors across 2,638 requests over both arms.

Raw JSONs (both arms + compare output) are checked into our pipeline repo; happy to share them here if useful.

One more thing, since your `qwen38_27b_readiness.md` planning note mentions re-deriving against the real checkpoint: Qwen3.8-27B dropped on the 14th, and its config is byte-identical to Qwen3.6-27B except `transformers_version` — so this fix applies as-is. We are running AURA quants against the real checkpoint this weekend and would be glad to share the resulting numbers.

Thanks again for the fast review rounds and the release cuts — much appreciated.

PR #84, opened 2026-08-24
fix: scope CB warm-state LDLQ identity per format

RobTand:
Dispatches the only merge-worthy subset of the archived `wip/rescue/pq-prereq-major1` branch without cherry-picking its stale snapshot.

What changes:
- Projects artifact-wide LDLQ scope onto the effective per-format warm identity.
- Treats NVFP4 `nvfp4` and `all` scopes as equivalent, and FP8 `none` and `nvfp4` scopes as equivalent.
- Canonicalizes historical aggregate stamps only after validating bool/scope consistency and the exact active packed-kernel ABI.
- Preserves current per-rung learned-bundle and qname-bound initializer identity.
- Keeps full metadata equality after the narrow projection, so every unrelated mismatch remains a cold fallback.

Validation:
- `tests/test_cb_warm_state.py`: 27 passed, including real packed-byte identity and the complete scope-equivalence matrix.
- Broader LDLQ, gate/chunk, per-rung, raw-sidecar, docs-staleness, and architecture suites: 131 passed.
- `py_compile` and `git diff --check` passed.

The integration base is active, so this is offered as a reviewed one-commit PR rather than moving `merge/proven-rescues` behind the ongoing work.

### DIFF
--- a/prismaquant/cb_warm_state.py
+++ b/prismaquant/cb_warm_state.py
@@ -32,6 +32,7 @@ from .cb_layout import (
 from .format_registry import canonical_format_name
 from .nvfp4_cb_footprint import (
     CBSerializationContext,
+    _ldlq_for_format,
     _physical_codebook_refs,
     cb_serialization_context_stamp,
     codebook_source_for_format,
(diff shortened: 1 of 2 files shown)


RobTand:
CI triage (2026-08-24): the focused rescue is not the source of the red full-suite jobs.

- Both Python jobs reached the same result: 4,576 passed, 140 skipped, 3 xfailed, and 3 failures.
- The PR diff from base `fee7a7f` is limited to `prismaquant/cb_warm_state.py` and `tests/test_cb_warm_state.py`.
- The two `test_dsv4_w8a16_export_handoff.py` failures reproduce unchanged at base `fee7a7f`: `production_weight_cache.py` now hashes to `5dc1cd6...`, while the frozen closure still expects `1cc27e3...`.
- `test_fisher_quadratic_form_exact` is also pre-existing: both Python jobs on `main` run 32594513995 failed with the same `7.450580596923828e-09 <= 1e-10` assertion.
- The import-surface and pinned-Gridbook-contract jobs passed. The port-specific local suite passed 27 tests; the broader LDLQ/per-rung/docs suite passed 131 tests.

These baseline failures should be repaired on the active base line; folding their re-freeze/math-policy decisions into this narrow branch-rescue PR would obscure review provenance.

PR #86, opened 2026-08-24
FP8-source wrapped-MoE checkpoints build end to end (imatrix scale contract + live-namespace bridging in the codebook exporter)

smb209:
## Summary

Quantizing **Qwen3.6-35B-A3B-FP8** (wrapped-VLM MoE, FP8 e4m3 source with
block-wise `weight_scale_inv`) through the codebook container hits two
independent defect families. Each was a fail-closed gate naming its own
remediation; this PR implements those remediations. As far as I can tell this
is the first FP8-*source* MoE through the packed-expert replay path (dense FP8
sources and BF16 MoE sources both work today, which is why nothing had caught
these).

## 1. `moe_imatrix._load_tensors` refuses the contract it names

The packed-expert replay refuses *every* float8 tensor with "…cannot decode an
FP8 checkpoint tensor without its serialized scale contract" — but never looks
for the `<name>_scale_inv` companion next to the tensor. The fix loads the
companion and dequantizes exactly, with tiling resolved fail-closed against
ambiguity: the checkpoint's declared `weight_block_size` when present (partial
trailing blocks slice exactly), exact-division inference when undeclared, and
refusal when neither is unambiguous (a uniform mis-inference over a partial
trailing block is a silent calibration corruption, so it refuses rather than
guesses). The original refusal stands for tensors genuinely lacking a
companion. Verified block-exact against a manual dequant of
`layers.0.experts.0.gate_proj` (512×2048, 4×16 scale). New tests: contract
round-trip, declared-block partial-trailing exactness, undeclared-ambiguous
refusal, companion-missing refusal preserved.

## 2. The streaming exporter cannot bridge a wrapped source's live namespace

A wrapped checkpoint puts three spellings of one module in play:

| namespace | spelling | who speaks it |
|---|---|---|
| checkpoint | `model.language_model.layers.N…` | on-disk tensors; gridbook's loader bridge |
| live module tree | `language_model.model.layers.N…` | the skeleton → the group planner's matched key |
| recipe | `model.layers.N…` | the allocator's layer_config; `_resolve_target`'s lookups |

`_plan_expert_stacks` keys groups by whichever spelling its regex matched. For
DSv4-class sources that is the **recipe** spelling — the planner's documented
contract ("so `_resolve_target` finds the group under the recipe name"), which
every consumer handles. For a wrapped source it is the **live** spelling, which
nothing downstream can bridge: the coverage gate KeyErrors on every uniform
(no-member-plan) group, and 30,720 consumed per-expert sources ship verbatim
into `ignore` — the exact "35B first-contact: 31511 copied tensors, 82 GB"
failure mode the passthrough comment already records. Three scoped fixes:

- **Live-spelled group keys are normalized to the recipe spelling**, derived
  from the group's own member tensors (real checkpoint names, which
  `_canonical_qname` always maps). Recipe- and checkpoint-keyed groups are left
  exactly as planned; the DSv4 passthrough/collapse suite
  (`test_nvfp4_cb_streaming`) passes in full. dspark sidecar groups are skipped
  — their naming contract is handled separately.
- **Packed expert stacks are named by their group's checkpoint prefix**,
  through a member-derived recipe→checkpoint bridge (the packed parent,
  e.g. `…experts.gate_up_proj`, is not a checkpoint leaf, so the profile mapper
  returns None for it). A recipe-spelled stack falls through gridbook's
  top-level loader rename — which reuses the model's own `hf_to_vllm_mapper`,
  i.e. checkpoint prefixes only — to the arch loader, which dies on the unknown
  root (`ValueError: no module or parameter named 'model'`). gridbook's
  `moe_toplevel_loader` docstring describes this exact fall-through and calls it
  "the bug".
- **Delegated (stock-CT) config-group targets are tower-relative in this
  container.** Measured on gridbook 0.8.11/0.9.0 + pristine vLLM 0.27.1:
  live-tree regexes match no module, the Linear stays an unquantized BF16
  fallback, and gridbook refuses fail-closed at weight load ("resolved to plain
  bf16 parameter"). Tower-relative spellings match and serve. **Scoped to the
  codebook container.** `export_native_compressed` is untouched — its shipped
  wrapper artifacts correctly keep live-tree targets, which *vanilla*
  compressed-tensors serving matches (verified against the published
  Qwen3.6-35B-A3B AutoRound artifact's `quantization_config`). The namespace a
  target must use depends on the consuming quant method; each exporter speaks
  its own consumer's dialect.

For wrapped sources, per-expert (Tier-2) config files address groups in the
recipe namespace after this change — consistent with the allocator's
layer_config, and unchanged for DSv4-class sources.

## Evidence

- Build: full chain on Qwen3.6-35B-A3B-FP8 (40 layers × 256 experts, top-8,
  moe_intermediate 512), FP8-CB ladder K28–K48 + FP8_DYNAMIC at the 3.6 bpp
  Pareto floor → 18.1 GB artifact.
- `artifact_completeness` passes **unmodified** — no checker changes in this PR.
- **vLLM compatibility gate (export metadata changed):** the artifact
  load-and-serves on stock gridbook **0.8.11 and 0.9.0** + pristine vLLM 0.27.1
  (installed file diffed against the wheel): sanity correct (17×24=408,
  capital-of-Japan=Tokyo, clean `finish=stop`), CUDA graphs captured (FULL
  decode + PIECEWISE mixed), single-stream decode 177.6 / 178.1 tok/s
  (RTX 5090). No forked runtime, no core patch, no vendored gridbook.
- Tests (known-good vLLM env): `test_moe_imatrix` + `test_nvfp4_cb_streaming` =
  **90 pass** (incl. the DSv4 passthrough/collapse family); adjacent namespace
  suites (`test_qwen3_5_text_only_namespace`,
  `test_artifact_completeness_namespaces`,
  `test_routed_expert_namespace_aliasing`, `test_cb_export_config`) pass;
  `test_docs_staleness` + `test_architecture_doc` = 18 pass.

## AGENTS.md compliance

- CHANGELOG entry under Unreleased/Fixed.
- `docs/ARCHITECTURE.md` not touched (no pipeline default, stage graph, format
  menu, or plugin-contract change); the mechanical gates pass. Happy to add a
  namespace note to the export section if you'd prefer it recorded there.
- GPU-bound, no new cache/residency mechanism; cross-repo contact stays behind
  the gridbook pin.
- Not an experimental method — production export bug fixes.
- Test gap, named: the exporter's three namespace fixes are function-local
  closures exercised through the end-to-end build plus the DSv4 non-regression
  rather than a dedicated unit fixture. If you'd like a wrapped-VLM regression
  fixture in `test_nvfp4_cb_streaming.py` style, point me at the preferred
  fixture pattern and I'll add it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,43 @@
 
 ## Unreleased
 
+### Fixed
+
+- **FP8-source MoE checkpoints now build end to end** — first wrapped-VLM
+  MoE with an FP8 source through the codebook container
+  (Qwen3.6-35B-A3B-FP8, fmt e4m3, block-wise `weight_scale_inv`). Two
+  independent defect families, both previously fail-closed with named
+  errors:
+  - `moe_imatrix._load_tensors` refused every float8 tensor instead of
+    fulfilling the serialized scale contract it named: it now loads the
+    `<name>_scale_inv` companion and dequantizes exactly — using the
+    checkpoint's declared `weight_block_size` when present (partial
+    trailing blocks handled exactly), exact-division inference when not,
+    and refusal when neither is unambiguous. The original refusal stands
+    for tensors lacking the companion. Tests: contract round-trip,
+    declared-block partial-trailing exactness, undeclared-ambiguous
+    refusal.
+  - `export_nvfp4_cb_streaming`: for a wrapped source the group planner's
+    regex-matched key lands in the LIVE module-tree namespace
+    (`language_model.model.*`) — neither the recipe spelling
+    `_resolve_target` looks up (the planner's own documented contract)
+    nor the checkpoint spelling emission bridges — so the coverage gate
+    KeyErrors on uniform groups and 30,720 consumed per-expert sources
+    ship verbatim into `ignore`. Live-spelled keys are now normalized to
+    the RECIPE spelling derived from the group's own member tensors;
+    recipe- and checkpoint-keyed groups (DSv4-class sources) are left
+    exactly as planned (`test_nvfp4_cb_streaming` passes in full).
+    Packed expert stack tensors are named by their group's checkpoint
+    prefix via a member-derived recipe-to-checkpoint bridge (a
+    recipe-spelled stack falls through gridbook's top-level loader to the
+    arch loader and dies). Delegated (stock-CT) config-group targets are
+    tower-relative in THIS container — measured on gridbook 0.8.11 +
+    pristine vLLM 0.27.1, where live-tree spellings leave the module
+    unquantized and gridbook refuses fail-closed at load. The vanilla
+    compressed-tensors container (`export_native_compressed`) is
+    untouched: its shipped wrapper artifacts correctly keep live-tree
+    targets, which vanilla vLLM matches.
+
 ## 0.16.2 — 2026-08-22
 
 ### Fixed
--- a/prismaquant/export_nvfp4_cb_streaming.py
+++ b/prismaquant/export_nvfp4_cb_streaming.py
@@ -2940,6 +2940,47 @@ def export_nvfp4_cb_streaming(
         if dspark_cb_sidecar
         else _plan_expert_stacks(skeleton, profile)
     )
+    # The planner's contract is that "whichever spelling matched becomes the
+    # group key, so `_resolve_target` finds the group under the recipe name"
+    # (see `_plan_expert_stacks`). For a WRAPPED source (Qwen3.5-VLM) the
+    # skeleton speaks the live module tree (`language_model.model.*`), which
+    # is neither the recipe (`model.*`) nor the checkpoint
+    # (`model.language_model.*`) spelling, so the matched key is in a
+    # namespace nothing downstream can bridge: the coverage gate KeyErrors on
+    # uniform groups and every consumed per-expert source ships verbatim into
+    # `ignore`. Normalize ONLY such keys to the RECIPE spelling, derived from
+    # the group's own member tensors (real checkpoint names, which
+    # `_canonical_qname` always maps); groups already keyed by a recipe or
+    # checkpoint spelling — DSv4-class sources — are left exactly as planned.
+    # The recipe->checkpoint prefix bridge is recorded for emission either
+    # way: packed-stack tensor NAMES must carry the checkpoint prefix (the
+    # packed parent is not a checkpoint leaf, so it cannot be mapped
+    # directly), and gridbook's top-level loader bridges checkpoint
+    # spellings only.
+    _canon_to_ckpt_prefix: dict[str, str] = {}
+    if not dspark_cb_sidecar:
+        def _member_prefixes(_projs):
+            for _members in _projs.values():
+                for _member in _members.values():
+                    _ck = str(_member).rsplit(".", 2)[0]
+                    _cm = _canonical_qname(str(_member), profile)
+                    _rc = _cm.rsplit(".", 2)[0] if _cm else None
+                    return _rc, _ck
+            return None, None
+
+        _rekeyed = {}
+        for _prefix, _projs in expert_groups.items():
+            _recipe, _ck = _member_prefixes(_projs)
+            if _recipe is not None and _ck is not None:
+                _canon_to_ckpt_prefix.setdefault(_recipe, _ck)
+            _key = _prefix
+            if (_recipe is not None
+                    and _prefix not in (_recipe, _ck)
+                    and _recipe not in expert_groups
+                    and _recipe not in _rekeyed):
+                _key = _recipe
+            _rekeyed[_key] = _projs
+        expert_groups = _rekeyed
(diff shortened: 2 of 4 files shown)


RobTand:
Merged. Both defect families are real, and your gridbook citations check out line for line: `moe_toplevel_loader::_hf_mapper_rename` (`:877`) does reuse the model's own mapper and its docstring calls the fall-through "the bug", and `config.py::_canonical_prefix` (`:233`) names the shipped 27B vintage at `:243`. On the merge into `main` at 0.16.2 the suites you named pass — `test_moe_imatrix`, `test_nvfp4_cb_streaming`, `test_qwen3_5_text_only_namespace`, `test_artifact_completeness_namespaces`, `test_routed_expert_namespace_aliasing`, `test_cb_export_config`, `test_docs_staleness`, `test_architecture_doc`, 135 passed before my deltas and 145 after.

Three review deltas ride in #89 rather than a review round, since they are mechanical. All three move in one direction: one contract per checkpoint, and refuse where the code inferred.

**1. The FP8 dequant grid is read, never inferred.** `_weight_block_size` now defers to `layer_streaming._declared_weight_block_size`, which refuses a checkpoint that pairs fp8 weights with scale tensors but declares no `weight_block_size`. Your own test comment names the hole the exact-division fallback leaves open — 200 rows over a 2-row scale plane divides exactly at 100 and is equally a 128-block tiling with a partial trailing block, and the two dequants differ on every row from 128 up. The deciding argument is not the arithmetic though: the streaming load path already refuses that checkpoint, so shipping a second mechanism that accepts it means two of our own paths disagreeing about one file. That is the failure class we spend the most on here. Your declared-block partial-trailing test survives unchanged; the undeclared test now asserts the refusal, and I added one for a transposed scale plane.

**2. Delegated targets go out canonical, not stripped.** Same conclusion as yours, one rule wider. `_delegated_target_name` now mirrors `_canonical_prefix` in full: `language_model.model.` to `model.`, and `language_model.<rest>` to `model.<rest>`. The bare strip mapped `language_model.lm_head` to `lm_head`, which matches neither the as-given nor the canonical spelling `_candidate_bases` tries, so an lm_head-shaped stock target would have gone out unmatchable. Since `_candidate_bases` tries both vintages, the canonical form is also the strictly safer one whatever the proximate cause on your stack. The bare `layers.` lift that `_canonical_prefix` also performs is deliberately not mirrored, so DSv4-class export metadata is byte-unchanged.

**3. The rekey fails closed.** A live-spelled group whose recipe key is already occupied, and a recipe prefix claimed by two different checkpoint prefixes, now raise. Both were silent pass-throughs that would have left a group under the unbridgeable live key — the 82 GB failure you are fixing, arriving quietly.

On your two named openings:

- **The namespace note is written.** `docs/ARCHITECTURE.md` §6.2 now records the three spellings on the codebook lane and which consumer each emission speaks to, with the recipe-namespace consequence for Tier-2 config files. Principle 13 does bite here: a config-groups emission rule changed.
- **The wrapped-VLM regression fixture is welcome but not blocking.** The pattern to follow is `test_nvfp4_cb_streaming.py`'s synthetic-skeleton fixtures — build the skeleton with `model.language_model.layers.N.*` checkpoint keys and a profile whose `recipe_to_vllm` carries the wrapper, then assert the planner's group keys, the packed-stack tensor names, and the delegated target namespace directly. If you send it, target the three closures individually rather than the end-to-end build.

One scope note recorded in the CHANGELOG so nobody reads this as full FP8-source coverage: sources that store scales as `.scale` siblings instead of `<name>_scale_inv` — DSv4-class, handled elsewhere through the profile's `fp8_scale_pairs` — still hit the replay's original refusal. Unchanged by this PR, and worth a second pass if you hit it.

Your sm_120 serve evidence is yours; there is no 5090 on this side to reproduce it. It gates nothing here, because the downstream route and completeness checks fail closed on their own. Thank you for the end-to-end build receipt — that is what made this reviewable.

smb209:
The `_canonical_prefix`-mirroring rework in #89 is tighter than the bare strip we
sent. `tests/test_wrapped_moe_namespace.py` — CPU-only synthetic skeleton, rebased onto
main (past #113/v12). Three closures asserted individually for a
`model.language_model.layers.N.*` source:

1. planner group-key normalization → `model.layers.0.mlp.experts`;
2. packed-stack names `model.language_model.layers.0.mlp.experts.{gate_up,down}_proj.cb_qweight`, no `language_model.model.` leak;
3. delegated target `re:^model[.]layers[.]0[.]self_attn[.]o_proj$`, no `re:^language_model`.

sm_121 has no CB cell in v12, so it declares a synthetic non-native target via your
`synthetic_cb_target` conftest fixture (not a route-status override). Guard verified
against current code: neutralizing `_delegated_target_name`
(`export_nvfp4_cb_streaming.py:5055`, the #89 `_canonical_prefix`-mirroring version)
leaks the prefix and fails closure 3. Branch `test/wrapped-moe-namespace-fixture-v2` —
PR or lift it.

PR #92, opened 2026-08-30
feat(trellis): wire 7 of 8 seam links; the anchors' currency stays a refusal

RobTand:
## What lands

`prismaquant/trellis_menu.py` shipped with `UNWIRED_LINKS`, an eight-entry ledger
that was both the refusal message and the re-enable checklist: with
`PRISMAQUANT_TRELLIS_SURFACE` set, the seam refused outright. This branch wires
**seven of those eight links**, each with a test that exercises the behaviour
rather than the source text, so an enabled seam builds the menu. **The eighth
entry stays, and stays refusing** — details below.

The seam's reach is unchanged in one respect that matters: it does not render,
export, or serve. `ProductionWeightCache` has no trellis mechanism and
`export_native_compressed` refuses a TCQ assignment. A selected rung is an
exactly priced report of what the DP would choose, not a shippable artifact.
The exact-bytes map is process-local — nothing rewrites `probe.pkl` — so a TCQ
assignment escaping into a later stage fails at the registry lookup. That is the
next wiring frontier, and `docs/ARCHITECTURE.md` §4.9 says so.

## The seven links that are wired

**Links 3 and 4, the silent ones, on the default path.**
`aggregate_fused_siblings` and `aggregate_packed_serving_groups`
(`prismaquant/allocator_candidates.py`) built each super-item menu by iterating
`FormatSpec` objects, so any candidate whose format has no registry spec was
dropped from every fused and packed group with no error. Both now build the menu
from the members' candidate lists through `_super_menu_format_names`: the run's
spec names first, in order, then any registry-free format every member offers,
sorted. A registry-free rung is priced as the exact sum of the member
candidates' bytes and `predicted_dloss` — the packed path's existing
construction — with an aggregated stderr of 0.0, because the solver `Candidate`
carries no stderr and fabricating one would be a claim.

**Link 5.** `allocator._extend_format_rank_with_candidate_menu` extends
`format_rank` from the candidate menu by each registry-free format's exact
aggregate serialized rate, `8*sum(bytes)/sum(params)` — the same quantity
`_serialized_format_rates` computes for a spec — so "higher rank" keeps meaning
"more bytes on this model". Existing names keep their relative order, and an
all-registry menu returns the input table key for key.
`promote_serving_units` now names the rank table on a miss instead of raising a
bare `KeyError` that reads as a corrupt assignment.

**Links 1, 2 and 6: no TCQ `FormatSpec` is registered, deliberately.**
`FormatSpec.memory_bytes_for_shape` is a closed form over
`weight_bits` / `scale_bits` / `group_size`, while a rung's exact size needs the
layout, the per-column schedule and the alphabet directory the manifest declares
(`trellis_footprint.trellis_tensor_payload_breakdown`). `(name, shape)` does not
determine the bytes, so a registered spec could only be plausible and wrong —
silently consumed by `_serialized_format_rates`, `footprint` and the payload
filter — and it would expose TCQ to every `act_quant_changes_input` /
`quantize_dequantize` consumer with no render behind it. Instead the seam writes
the exact bytes into `_memory_bytes_by_format`, the repo's existing single
mechanism (principle 8), which the payload filter, `footprint`,
`compute_achieved` and bit attribution already prefer. The payload filter
(`prismaquant/allocator.py`) and `footprint.py` refuse pointedly when the map is
absent instead of falling through to `fr.get_format`.
`layer_config.canonicalize_format` round-trips `TCQ_<grid>_R<q256>` so the
exporter's pointed refusal is reachable. The cost is N pointed refusals instead
of one registry entry, and each one refuses where a closed form would have
guessed.

**Link 7.** The run's cost mode is the **attested** one from the cost table's
provenance (re-vet R2), never `os.environ` — `run-pipeline.sh` assigns
`COST_MODE` with `:=` and never exports it. The surface provenance (manifest
SHA-256, declared currency, the run's objective currency, anchor activation
contract) travels into `layer_config.json`'s `__prismaquant__` block under
`trellis_surface`, present only when the seam contributed (principles 12, 14
and 6).

The pinned menus — `lm_head`, MTP, visual — pass `apply_trellis_surface=False`:
their single format is an operator declaration, not a DP choice.

## The eighth link is not wired, and it is not a wiring gap

`UNWIRED_LINKS` keeps exactly one entry, `trellis_rate_surface.py:43-52`
(`prismaquant/trellis_menu.py:170`).

The ladder's measured anchors are **weighted SSE under a per-input-channel
activation second moment** — an output-MSE proxy. The production DP ranks
`predicted_dloss` in the **AURA KL-adjoint**. A DP that ranks a weighted-SSE
trellis rung against an AURA-priced NVFP4 rung is not solving any stated
objective, and no amount of plumbing changes what the anchors measured.
Producing AURA-priced anchors is a dW-supply problem: `aura_cost`'s two dW
sources, `ProductionWeightCache` and `fr.get_format(...).quantize_dequantize`,
both require a registered format. That work is owned elsewhere, and the size of
the gap is under measurement on a second box as this PR opens. Nothing here
should be relaxed on the strength of an expectation about that result.

`_require_run_currency` (`prismaquant/trellis_menu.py:312`) is the enforcement.
It compares the manifest's declared currency against
`COST_MODE_OBJECTIVE_CURRENCY[cost_mode]` — the same `COST_RENDER x
COST_OBJECTIVE` decomposition `run-pipeline.sh` resolves — so the expected value
is definitional rather than chosen (principle 2). Two adjacent refusals come
with it: an unstamped cost table is refused rather than compared against a
default, and a cost mode that names no objective is refused rather than
defaulted. An enabled seam on an `aura` run therefore refuses on the currency,
which is the honest failure, instead of refusing on missing links, which was the
placeholder.

## Default-path safety (principle 6)

Both aggregators are default-path code: every production allocation runs both.
`tests/test_super_item_menu_byte_identity.py` and its golden fixture were
generated and committed against the **unmodified** aggregators in the first
commit of this branch, then re-run after the change. The digest is unchanged:
`033adb45e71a51b831f5335047fa56e5fb00dc3137894a983fa0146bf456197d`.

The golden pins every super item's whole menu in order — `fmt`,
`bits_per_param`, `memory_bytes`, `predicted_dloss`, both serialized identities,
`activation_pricing` and `serving_lane`, floats by `repr` so the comparison is
exact rather than tolerance-based — plus `_memory_bytes_by_format` and the whole
`costs_ext` row including `predicted_dloss_stderr`. Scenarios cover both
aggregators, `PRISMAQUANT_COST_UCB_Z` at zero and non-zero, calibrated gains,
activation fair pricing, and the role-split packed profile.

A golden that passes is not evidence until a deliberate change to the driver
makes it fail. `test_the_golden_is_load_bearing` reorders every super item's
menu at runtime — the same set of formats, a different DP order, the minimal
perturbation an order-blind record would miss — and asserts the digest moves
(`033adb45...` to `0d7b2edf...`) and returns when the mutation is undone.

## Documentation (principle 13)

`docs/ARCHITECTURE.md` §4.9 is rewritten in the same commit as the code, and the
provenance block is re-stamped at the head of the file. The second commit
corrects two claims in that stamp that were loose in the direction the section
exists to prevent:

- `kl_measurement` does **not** read `_memory_bytes_by_format` for a TCQ row.
  Its byte helper takes a resolved `FormatSpec`, so the `specs_by_name` lookup
  raises first. `compute_achieved` and bit attribution prefer the map but do not
  refuse pointedly; only the payload filter and `footprint` do.
- The map is process-local, so the seam's reach is allocation-time only.

The merge commit changes nothing normative: the default path is byte-identical
by the golden, and the seam is env-gated and off unless
`PRISMAQUANT_TRELLIS_SURFACE` names a manifest. No further doc edit is owed.

## Tests

- `tests/test_trellis_menu.py`: 31 tests, up from 15, all behaviour. Four
  earlier source-text assertions are gone — the `for spec in formats:` count,
  the exporter greps, the `build_candidates` signature check — replaced by tests
  that call the code.
- `tests/test_super_item_menu_byte_identity.py` and
  `tests/fixtures/super_item_menu_golden.json`: new, described above.
- Full suite on the merged branch, CPU-only
  (`PYTHONPATH=. CUDA_VISIBLE_DEVICES="" pytest tests/ -q --timeout=300`,
  Python 3.12): **5510 passed, 143 skipped, 3 xfailed, 179 subtests passed,
  0 failed** in 892 s. The skips are the CUDA-gated tests.

## Merge base

Merged, not rebased, so the P6 evidence in `d08f3d9`'s message stays attached to
the commit arrangement it cites. `origin/main` at `1ca4c7d` brings in PR #90 and
PR #91; both touch files this branch does not
(`tests/test_math_reunderwrite_pins.py`,
`prismaquant/dsv4_w8a16_export_handoff.py`), and the merge had no conflicts.

## How to verify

```bash
PYTHONPATH=. CUDA_VISIBLE_DEVICES="" pytest \
  tests/test_trellis_menu.py \
  tests/test_super_item_menu_byte_identity.py \
  tests/test_docs_staleness.py \
  tests/test_architecture_doc.py -q
```

`test_the_only_remaining_unwired_link_is_the_currency` and
`test_the_ladders_own_currency_is_the_surviving_refusal` are the two that pin
the eighth link open.

## CI status: py3.12 green, py3.11 red, and the py3.11 failure is this PR's

Run [33285583387](https://github.com/RobTand/prismaquant/actions/runs/33285583387):
`import surface` passes, `pinned Gridbook contract` passes,
`tests (py3.12, cpu)` passes, `tests (py3.11, cpu)` **fails**.

The two py3.11 failures are the new byte-identity tests:

```
FAILED tests/test_super_item_menu_byte_identity.py::test_super_item_menus_are_byte_identical_to_the_golden
  - super-item menu changed in scenario 'packed_hedged_gained_priced'
FAILED tests/test_super_item_menu_byte_identity.py::test_the_golden_is_load_bearing
  - assert '43c261df...' == '033adb45...'
2 failed, 5499 passed, 154 skipped, 3 xfailed, 179 subtests passed
```

**What actually differs is 1 to 2 ULP of float64.** Reading the assertion diff,
every changed field is a `predicted_dloss` whose `repr` moved in the last digit:

| golden (recorded on 3.12) | py3.11 run | ULP | relative |
|---|---|---|---|
| `0.002071872` | `0.0020718720000000006` | 1 | 2.09e-16 |
| `0.002631341383012154` | `0.0026313413830121547` | 2 | 3.30e-16 |
| `0.21080600999999996` | `0.21080601000000002` | 2 | 2.63e-16 |

Those are float64 epsilon (2.22e-16). The structural-looking blocks in the
pytest diff are `pprint` key ordering between the JSON-loaded golden and the
freshly built payload, not missing or added entries: the menus, the members,
the byte counts and `n_params` are identical.

So the gate is doing exactly what it says — it compares floats by `repr`, which
is environment-exact, not tolerance-based — and the arithmetic associates
differently under the py3.11 job's resolved dependency set. That is a real
property of the gate that a reviewer should decide about. Three dispositions,
none of them taken here:

1. Leave it. `main` is separately narrowing the CI matrix to 3.12
   (the workflow header records "3.11 was not failing on an interpreter
   incompatibility — it was a proxy for test ORDER").
2. Serialize at float64's guaranteed round-trip decimal precision
   (`DBL_DIG` = 15 significant digits) instead of `repr`'s 17. That constant
   comes from the dtype, which is the one source principle 2 allows, but it
   requires regenerating the golden from the pre-change aggregators to keep the
   P6 evidence chain intact.
3. Record the golden per interpreter.

Regenerating the fixture against the current code to make the failure go away is
**not** on that list. A fixture edited until it passes stops being evidence
("mutate the driver, not the fixture").

For the pre-existing baseline: `main` at `1ca4c7d` — this PR's merge base — is
itself red on the same arm, with a different test:
[run 33284249771](https://github.com/RobTand/prismaquant/actions/runs/33284249771),
`tests (py3.11, cpu)` fails
`test_glm5_next_streamed_forward_parity.py::test_long_sequence_parity_per_layer_type[dsa_only]`
(`diverges by nan`) while `tests (py3.12, cpu)` passes. That failure does not
appear on this branch's py3.11 run, and this branch touches nothing in that
file.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,8 +1,30 @@
 # PrismaQuant Architecture
 
-As of: 2026-08-29 · `claude/trellis-continuous-surface` — stamps follow, newest
+As of: 2026-08-29 · `claude/trellis-seam-wiring` — stamps follow, newest
 first, each recording its own branch and date. Re-stamped (2026-08-29,
-`claude/trellis-continuous-surface`) for the **trellis seam correction**
+`claude/trellis-seam-wiring`) for the **trellis seam wiring** (§4.9): seven of
+the eight `UNWIRED_LINKS` are now wired, each with a behaviour test, so an
+enabled seam builds the menu instead of refusing outright. Both super-item
+aggregators build each menu from the MEMBERS' candidate lists rather than from
+`FormatSpec` objects (`_super_menu_format_names`) — the silent gap that dropped
+every rung from every coupled unit; `format_rank` is extended from the
+candidate menu by exact serialized rate so `promote_serving_units` can rank a
+selected rung; exact bytes travel in `_memory_bytes_by_format` written by the
+seam and read by the payload filter, `footprint`, `compute_achieved`,
+`kl_measurement` and bit attribution, each refusing pointedly rather than
+falling back to a closed form (**no TCQ `FormatSpec` is registered, and that is
+the design** — a rung's bytes are not a function of `(name, shape)`); and the
+run's **attested** `COST_MODE` plus the surface provenance (manifest sha256,
+currency, anchor activation contract) now travel into `layer_config.json`'s
+`__prismaquant__` block. The eighth link — the anchors' weighted-SSE currency —
+is not plumbing and stays a refusal (`_require_run_currency`). Because both
+aggregators are default-path code, `tests/test_super_item_menu_byte_identity.py`
+and its golden fixture were generated and committed **before** the change and
+re-run after: digest
+`033adb45e71a51b831f5335047fa56e5fb00dc3137894a983fa0146bf456197d`, unchanged.
+Earlier same-day stamp (`claude/trellis-continuous-surface`) for the
+**trellis seam correction** — *superseded by the wiring stamp above: the
+outright refusal and the eight-entry ledger it describes are history*
 (§4.9): the seam that landed earlier the same day is now **fail-closed** when
 enabled, and this document's claim that it made trellis rungs "pass the same
 legality, aggregation and byte accounting every other candidate does" is
@@ -3392,7 +3414,8 @@ is now the ONE seam that does, and it is off by default.
 **The flag, and what it does today.** `PRISMAQUANT_TRELLIS_SURFACE=<manifest.json>`.
 Unset, `augment_candidates` returns its input object unchanged, so a run
 without the flag executes exactly the path it executed before the seam existed
-— the `PRISMAQUANT_FISHER_CAP_MULTIPLIER` precedent (§ P6). **Set, it refuses.**
+— the `PRISMAQUANT_FISHER_CAP_MULTIPLIER` precedent (§ P6). **Set, it builds
+the menu, and refuses only on the anchors' currency.**
 
 > **Correction (2026-08-29, same day).** The first version of this section said
 > the seam's placement inside `allocator_candidates.build_candidates` meant
@@ -3400,49 +3423,81 @@ without the flag executes exactly the path it executed before the seam existed
 > other candidate does." That was **false on all three counts**, and it is the
 > § P13 "currency is not truth" case in its own doc: the paragraph was written
 > in the same commit as the code and was wrong about it. The seam appends
-> AFTER the per-spec legality loop, so it never runs
-> `check_stats_format_applicability` nor the `_memory_bytes_by_format` write at
-> `allocator_candidates.py:1950`; aggregation drops every rung; byte accounting
-> `KeyError`s. `trellis_menu.UNWIRED_LINKS` is now the authoritative list —
-> eight entries, each with a file:line — and it is the text of the refusal.
+> AFTER the per-spec legality loop, so it never ran the
+> `_memory_bytes_by_format` write at `allocator_candidates.py:1950`;
+> aggregation dropped every rung; byte accounting `KeyError`d.
+> `trellis_menu.UNWIRED_LINKS` is the authoritative list, and the seam refused
+> as a whole while it held eight entries.
 
-The eight, in the order a run would hit them: no TCQ `FormatSpec`
-(`format_registry.py:1267-1272`); the exact assignment-payload filter falling
-through to `fr.get_format` because nothing writes `_memory_bytes_by_format` for
-a TCQ row, which kills the allocator inside the Pareto sweep **before**
-`layer_config.json` and makes the pointed refusals in `layer_config` and the
-exporter unreachable (`allocator.py:3369-3386`); fused-sibling aggregation
-building super-item menus by iterating `FormatSpec` objects
-(`allocator_candidates.py:2464`); the identical packed-expert construction
-(`:2701`); `promote_serving_units`' `format_rank` lookup, which does not crash
-today only because aggregation guarantees a TCQ unit is a lone ungrouped
-Linear (`allocator_solver.py:340-342`); the byte-budget path's own registry
-lookup (`footprint.py:1183`); `build_candidates` being called with neither
-`cost_mode=` nor `trellis_provenance=`, so the currency gate compares against
-`os.environ.get("COST_MODE","aura")` — a variable `run-pipeline.sh` sets with
-`:=` and never exports — and the manifest identity and anchor contract are
-discarded rather than travelling with the assignment (`allocator.py:2756`,
-§§ P12/P14); and the anchors' currency being weighted SSE under an activation
-second moment, an output-MSE proxy and **not** the AURA KL-adjoint the DP ranks
-in (`trellis_rate_surface.py:43-52`).
+**Seven of the eight were wired later the same day; one is left, and it is not
+a wiring gap.**
 
-**Why refuse as a whole rather than wire it halfway.** The eight do not fail
-alike. The registry gaps crash **loudly**; the aggregation gaps are **silent** —
-they drop every rung from every fused and packed group and hand back a
-plausible frontier in which only `o_proj` and `down_proj` could carry one. A
-partial fix that removed the crashes would trade the loud failure for the
-silent one, and the packed-expert case was worse still: a hand-rolled 2-tuple
-shape priced a 128-expert row as one expert (a **128×** underprice, reported as
-"0 unit(s) skipped"), while the guard meant to catch it read a stats key
-nothing writes. Both are fixed — the seam now uses
-`allocator_solver._shape_from_stats` and
+| was | now |
+|---|---|
+| no TCQ `FormatSpec` (`format_registry.py:1267-1272`) | **still true, on purpose** — see *Why no TCQ `FormatSpec`* below. Exact bytes travel in `_memory_bytes_by_format` instead, and every site that would have resolved the registry refuses pointedly. |
+| the exact assignment-payload filter falls through to `fr.get_format` (`allocator.py:3369-3386`) | prefers the recorded map (it always did); a TCQ row with no recorded bytes now raises a pointed `AssertionError` naming what is missing, instead of a registry `KeyError`. `layer_config.canonicalize_format` round-trips `TCQ_*`, so the exporter's own refusal is now REACHABLE. |
+| fused-sibling aggregation iterates `FormatSpec` objects (`allocator_candidates.py:2464`) | both aggregators build each super item's menu from `_super_menu_format_names(formats, member_intersection)`: the run's spec names first, in order, then any registry-free format EVERY member offers. Registry-free rungs are priced as the exact sum of the member candidates' bytes and Δloss, with an aggregated hedge of 0.0 (the solver `Candidate` carries no stderr, so fabricating one would be a claim). |
+| the identical packed-expert construction (`:2701`) | same helper, same contract. The seam still refuses packed-expert ROWS (no per-expert trellis render exists), so no manifest reaches this path today; the aggregation contract is tested by injection. |
+| `promote_serving_units`' `format_rank` lookup (`allocator_solver.py:340-342`) | `allocator._extend_format_rank_with_candidate_menu` extends the rank table from the CANDIDATE MENU by each registry-free format's exact aggregate serialized rate (`8·Σbytes/Σparams`) — the same quantity `_serialized_format_rates` computes for a spec, so "higher rank" keeps meaning "more bytes on this model" (§ P2). Byte-identical no-op when every candidate format is a spec. The solver's lookup now names the rank table on a miss instead of raising a bare `KeyError`. |
+| the byte-budget path's own registry lookup (`footprint.py:1183`) | a TCQ branch before the final `else` reads `_memory_bytes_by_format` and **refuses** when it is absent — same doctrine as the CB branch above it, no closed-form fallback. |
+| `build_candidates` called with neither `cost_mode=` nor `trellis_provenance=` (`allocator.py:2756`) | both are passed. The cost mode is the **attested** one: `cost_data["provenance"]["cost_mode"]` (re-vet R2), never `os.environ` — `run-pipeline.sh` assigns `COST_MODE` with `:=` and never exports it. The provenance (manifest **sha256**, declared currency, the run's objective currency, the anchor activation contract, lane/route status) lands in `layer_config.json`'s `__prismaquant__` block under `trellis_surface`, present only when the seam contributed (§§ P12/P14, § P6). |
+| the anchors' currency is weighted SSE under an activation second moment (`trellis_rate_surface.py:43-52`) | **the one remaining entry**, and the surviving refusal. |
+
+**Why the last one stays a refusal.** It is not plumbing. The ladder's measured
+anchors are weighted SSE under a per-input-channel activation second moment —
+an output-MSE proxy — while an `aura` run's DP ranks the KL-adjoint. No code
+change converts one into the other; AURA-priced anchors are a **dW-supply**
+problem (`aura_cost`'s two dW sources both require a registered format).
+`trellis_menu._require_run_currency` therefore compares the manifest's declared
+`currency` against `COST_MODE_OBJECTIVE_CURRENCY[cost_mode]` — the same
+`COST_RENDER × COST_OBJECTIVE` decomposition `run-pipeline.sh` resolves, so the
+expected value is definitional rather than chosen (§ P2) — and refuses with
+`TrellisSeamUnwiredError`. An **unstamped** cost table is refused too, rather
+than compared against a default.
+
+**Why refuse as a whole rather than wire it halfway (the reasoning that held
+until the links landed).** The eight did not fail alike. The registry gaps
+crashed **loudly**; the aggregation gaps were **silent** — they dropped every
+rung from every fused and packed group and handed back a plausible frontier in
+which only `o_proj` and `down_proj` could carry one. A partial fix that removed
+only the crashes would have traded the loud failure for the silent one. The
+packed-expert case was worse still: a hand-rolled 2-tuple shape priced a
+128-expert row as one expert (a **128×** underprice, reported as "0 unit(s)
+skipped"), while the guard meant to catch it read a stats key nothing writes.
+Both are fixed — the seam uses `allocator_solver._shape_from_stats` and
 `allocator_candidates._stats_indicates_packed_expert`, and refuses a packed row
-with a counted reason — but the fix is inside `build_trellis_menu`, which is
-research-reachable and cannot reach an artifact.
+with a counted reason.
 
-**So: `build_trellis_menu` builds a correctly priced menu; `augment_candidates`
-refuses.** Enabling the surface means landing the eight links with tests that
-exercise behaviour and then deleting the refusal — not passing a flag.
+**Why no TCQ `FormatSpec`.** The obvious alternative — register a minimal spec
+so every registry lookup just works — is unavailable as a matter of arithmetic,
+not taste. `FormatSpec.memory_bytes_for_shape` is a closed form over
+`weight_bits` / `scale_bits` / `group_size`; a rung's exact size needs the
+layout, the per-column schedule and the alphabet directory
+(`trellis_footprint.trellis_tensor_payload_breakdown`: 16-byte row-stride
+alignment, an 88-byte wire header, a nibble schedule plane, block offsets under
+`tight_offsets`, a family-specific scale plane), all per-campaign manifest
+data. `(name, shape)` therefore does not determine the bytes, and a registered
+spec could only be **plausible and wrong** — silently consumed by
+`_serialized_format_rates`, `footprint` and the payload filter, which is
+exactly the failure this seam exists to prevent. It would also expose TCQ to
+every `act_quant_changes_input` / `quantize_dequantize` consumer with no render
+behind it. So the bytes ride the `Candidate` and `_memory_bytes_by_format` —
+the repo's existing single mechanism (§ P8) — `fr.get_format("TCQ_…")` keeps
+raising, and the cost is N pointed refusals instead of one registry entry. Each
+refuses where a closed form would have guessed.
+
+**Principle-6 evidence for the aggregation change.** Both aggregators are
+default-path code. `tests/test_super_item_menu_byte_identity.py` +
+`tests/fixtures/super_item_menu_golden.json` were generated and committed
+**against the unmodified functions**, and pin every super item's whole menu in
+order (fmt, `bits_per_param`, `memory_bytes`, `predicted_dloss`, both
+serialized identities, `activation_pricing`, `serving_lane`, all floats by
+`repr`), the `_memory_bytes_by_format` map, the whole `costs_ext` row including
+`predicted_dloss_stderr`, and the pass-through row names — across five
+scenarios covering both aggregators, `PRISMAQUANT_COST_UCB_Z` at 0 and 1.5,
+calibrated gains, activation fair pricing and the role-split packed profile.
+Golden digest `033adb45e71a51b831f5335047fa56e5fb00dc3137894a983fa0146bf456197d`,
+unchanged after the change.
 
 **Why a manifest, not a `FORMATS` enum entry.** A trellis rung is
 `(family, body_rate_q256, layout, schedule, alphabets)`, and the wire carries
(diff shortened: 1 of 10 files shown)


RobTand:
## The py3.11 golden failure: root cause corrected, and the pin fixed

**The root cause is not numpy, and not serialization. It is CPython's builtin `sum()`.**

CPython 3.12 gave builtin `sum()` **Neumaier compensated summation for float inputs**
([gh-100425](https://github.com/python/cpython/issues/100425)); 3.11 sums naively
left-to-right. A packed serving group's `predicted_dloss` is exactly one builtin
`sum()` over 9 member floats
([`allocator_candidates.py` packed path](../blob/claude/trellis-seam-wiring/prismaquant/allocator_candidates.py#L2795),
and the same construction in the fused path), so the two interpreters land 1–2 ulp
apart on identical inputs through identical aggregation logic.

Verified rather than argued. Swapping **only** the summation algorithm inside
`prismaquant.allocator_candidates` — nothing else changed, same interpreter, same
numpy — reproduces every divergent value exactly, including all three the CI job
reported:

| field | golden (3.12, Neumaier) | naive sum (3.11) | ulps |
|---|---|---|---|
| `packed_hedged_gained_priced` L0 `cost_rows/FP8_E4M3/predicted_dloss` | `0.0020718720000000006` | `0.002071872` | 1 |
| `packed_hedged_gained_priced` L0 `menu[1]/predicted_dloss` | `0.0026313413830121547` | `0.002631341383012154` | 2 |
| `packed_hedged_gained_priced` L1 `cost_rows/NVFP4/predicted_dloss` | `0.21080601000000002` | `0.21080600999999996` | 2 |
| + 5 more, all `predicted_dloss`, all ≤ 2 ulps | | | |

There is **no numpy anywhere in this call chain** — `grep -n "np\.\|numpy"
prismaquant/allocator_candidates.py` returns one comment about accepting numpy ints,
and neither `activation_fair_pricing.py` nor `format_registry.py` imports numpy on
these paths. The 2.4.6 / 2.5.2 skew in the CI logs was correlation: the numpy
version tracked the Python version the job resolved.

## Why the test was wrong, independently of which job was inconvenient

Same class of defect as merged #90 (`c198593`): **a tolerance pinned below the
numerical noise floor of the quantity**, so the assertion was about summation order
rather than the property the test exists to protect. One ulp of float64 is 2.22e-16;
the test demanded 0.

`7440958` already dropped py3.11 from the matrix, which makes this symptom go away
on its own. That is not the fix. The golden was one reduction-implementation change
away from flaking again — a CPython change on 3.12+, a future compensated reduction,
anything that reassociates a 9-term float sum. **The pin was wrong; one job being
inconvenient was only how we found out.**

## The fix

Exact match is replaced by a comparison split along what each field actually is:

**Exact, no tolerance** — everything that decides bytes or DP structure: menu order,
`fmt`, `memory_bytes`, `_memory_bytes_by_format`, `n_params`, member lists, both
serialized identities, `activation_pricing`, `serving_lane`, the applied-pricing
marker, pass-through row names. These are ints, bools, `None` and identity strings.
The walker now also checks dict key-sets **both directions** and list lengths, so a
key the driver *adds* fails too — the old exact `==` got that for free and the
replacement must not lose it.

**float64 precision** — the four accumulated float fields (`predicted_dloss`,
`weight_mse`, `predicted_dloss_stderr`, `bits_per_param`), keyed by an explicit
`_FLOAT_FIELDS` set rather than by sniffing whether a string parses as a number.

The bound is derived from the dtype (principle 2), not from what made the test pass:
`16 · sys.float_info.epsilon · |golden|`. 16 ulps covers the longest chain any pinned
field accumulates over — per member ~5 roundings (`0.5·h_trace·mse`, then the
calibrated gain and the activation penalty), a 9-term reduction, then the hedge
subtract and `z·sqrt(Σ stderr²)` add, ~18 roundings at ≤0.5 ulp each over an
all-positive sum with condition number 1, giving a bound near 9; 16 is the next power
of two and the multiplier #90 derived the same way. A golden of exactly `0.0` gets a
tolerance of exactly `0.0`, so bit-exact rungs stay bit-exact with no special case.

**The fixture was not regenerated.** A golden edited until it goes green stops being
evidence. `git diff` touches one file: the test.

## Proof the gate is still load-bearing

`test_the_golden_is_load_bearing` now has two teeth, both mutating the **driver**:

* **structural** — reorder every super item's menu (same format set, different DP
  order). Comparison fails, as before. No tolerance can absorb it.
* **numeric** (new) — scale every member's `predicted_dloss` by `1 + 1e-12`.
  Comparison fails. This is asserted in-tree rather than only recorded in a commit
  message, so a future loosening of the bound trips it.

Bisected threshold, the #90-equivalent number:

```
perturb 1e-13: CAUGHT (56 mismatches)
perturb 1e-14: CAUGHT (56 mismatches)
perturb 1e-15: not caught
```

**1e-14 relative is caught, 1e-15 is not**, against a 16-ulp bound of 3.55e-15. The
largest legitimate reduction difference measured across all five scenarios is 1.5 ulps
(3.3e-16) — a factor of ~30 of margin below the bound, and the gate still resolves a
driver change nine orders of magnitude finer than the 1e-5 / 1e-6 threshold #90
settled for in float32. No change a reviewer would call real survives this.

## Why not drop the floats entirely

Considered, and rejected. A structure-only gate (order, membership, byte counts)
would be tolerance-free — but the DP ranks on `predicted_dloss`, so it would pass
through a genuine cost regression untouched. The floats are the half of this fixture
with something to say about allocation quality; they get compared, at the precision
they are actually computed to.

## Verification

* `tests/test_super_item_menu_byte_identity.py` — 4 passed (py3.12, numpy 2.4.2, aarch64).
* **The 3.11 reproduction passes the new comparison** — with the naive `sum()` patched
  in, `compare_to_golden` returns clean. This is the actual proof of fix; no 3.11
  interpreter exists on the box that produced it.
* Full CPU suite on aarch64 (`CUDA_VISIBLE_DEVICES="" PYTHONPATH=. pytest tests/ -q`):
  **5510 passed, 143 skipped, 3 xfailed, 179 subtests passed, 0 failed** in 15m22s.
  The four failures this box is known to produce (`test_rtx4090_fp8_burn.py::test_allocate_uses_exact_sealed_bundle_probe_through_child`
  and three `test_run_pipeline_gates_execute.py` gate tests) did not reproduce in this
  ordering; both files also pass in isolation on the base commit `9596dc2` (89 passed),
  so nothing in this run is attributable to the change either way.

## Owed follow-up (not done here, by instruction)

`docs/ARCHITECTURE.md` §"Principle-6 evidence for the aggregation change" (added by
this PR) still says the fixture pins "all floats by `repr`" and cites the golden
digest as unchanged evidence. Both sentences are now stale: the four accumulated
float fields are compared to float64 precision, and the digest is a provenance stamp
that the tests no longer assert. That file is owned by a concurrent branch in this
tree, so it is flagged rather than edited.

RobTand:
CI on the fix commit `6e4092b` is **green on all four jobs, py3.11 included** ([run 33288016823](https://github.com/RobTand/prismaquant/actions/runs/33288016823)): `tests (py3.11, cpu): success`, `tests (py3.12, cpu): success`, import surface, pinned Gridbook contract.

That is the direct confirmation the naive-`sum()` emulation above stood in for — the real 3.11 interpreter now agrees with the golden, on a fixture that was not regenerated.

RobTand:
OpenCode assist (Muse Spark) — suggested close-out, maintainer decides.

Recommend CLOSE unmerged: #118 proposes archiving trellis_* (5 modules + spec + tests, ~168 KB) for the retired gridbook.trellis.wire.v1, and #125 deletes build_candidates(cost_mode=) which this PR's link 7 consumes. Landing 7 wired links for a seam nothing can reach would re-anchor a retired wire.

Salvage (none covered by #99/#100/#103, verified disjoint):
- prismaquant/allocator_candidates.py _super_menu_format_names + fused/packed registry-free pricing -> only piece with live default-path value; re-propose standalone, minus trellis.
- tests/test_super_item_menu_byte_identity.py + tests/fixtures/super_item_menu_golden.json -> rescue with it as the P6 pin (keep 16-ulp bound + driver-mutation teeth from 6e4092b).
- prismaquant/allocator_solver.py rank-miss message -> trivial, rescue opportunistically.

Needs no action: trellis_menu.py wiring, allocator.py/footprint.py TCQ refusals, layer_config.py TCQ round-trip, link-7 cost_mode/provenance plumbing, docs §4.9 — all die with the #118 archive / #125 deletion by design. The py3.11 Neumaier-sum diagnosis stays in comments as record.

RobTand:
## Verdict: SUPERSEDED — close unmerged, salvage three pieces

> **Ref note.** Line numbers below are on `tessera/decouple-gridbook` at
> `a212b73` unless a line explicitly says `origin/main`. The two trees differ:
> `prismaquant/tessera_campaign.py` and `tessera_group_composites` exist only on
> `a212b73` (`git ls-tree origin/main prismaquant/tessera_campaign.py` is empty;
> `git grep -c tessera_group_composites origin/main` finds nothing). The same
> currency gate reads `allocator_candidates.py:1399` on `a212b73` and `:1286` on
> `origin/main`.

The surface this PR wires was archived three days after the PR opened (PR created
2026-08-30 01:24 UTC; `a212b73` dated 2026-09-02 22:46 EDT).
`a212b73` ("Archive the Gridbook trellis rate surface; the wire it priced has no
runtime left", closes #118) moved the five `prismaquant/trellis_*.py` modules,
their four test files and `serving_profile_specs/trellis_research_sm121.json`
whole into `archive/trellis_wire_2026-09-02/`, because Rob retired the
Gridbook/NVFP4-CB lane on 2026-09-02 and no sanctioned runtime reads
`gridbook.trellis.wire.v1` any more. `PRISMAQUANT_TRELLIS_SURFACE` now refuses
from `allocator_candidates.refuse_retired_trellis_surface`, the first statement
of `build_candidates`, pinned by `tests/test_retired_trellis_surface_refusal.py`.

**One correction to the record before anything else: `a212b73` is not yet on the
integration branch.** `git merge-base --is-ancestor a212b73 origin/main` → false;
`origin/main` is at `55d43e3` and still carries all five modules. The archival
lives on `tessera/decouple-gridbook`. So this verdict rests on Rob's lane
decision and on `a212b73`, not on the current state of `main` — and the archival
landing is a prerequisite, not a consequence, of closing this.

Also concrete, since the PR shows as CONFLICTING: exactly **one** file conflicts.
`git merge-tree --write-tree --name-only origin/main origin/claude/trellis-seam-wiring`
names `docs/ARCHITECTURE.md` and nothing else. Everything else auto-merges.

### Per-file disposition (10 files, head `6e4092b`, merge base `1ca4c7d`)

| file | +/− | status |
|---|---|---|
| `prismaquant/trellis_menu.py` | +215/−81 | **Archived** → `archive/trellis_wire_2026-09-02/prismaquant/trellis_menu.py` (`a212b73`) |
| `tests/test_trellis_menu.py` | +530/−59 | **Archived** → `archive/trellis_wire_2026-09-02/tests/test_trellis_menu.py` |
| `prismaquant/allocator.py` | +144/−0 | **Live file, dead diff.** Every hunk hangs off `from .trellis_formats import parse_trellis_format_name` (`:132`); that module is archived. Plus link 7's `run_cost_mode` / `trellis_provenance` plumbing and three `apply_trellis_surface=False` call sites. |
| `prismaquant/footprint.py` | +22/−0 | **Live file, dead diff** — same import, one `parse_trellis_format_name` refusal branch in `assignment_artifact_bytes`. |
| `prismaquant/layer_config.py` | +11/−0 | **Live file, dead diff** — same import; round-trips `TCQ_<grid>_R<q256>` so the exporter's pointed refusal stays reachable. `a212b73` deleted that pointed refusal in favour of the generic `FORMAT_SCHEME` one, verified by execution. |
| `prismaquant/allocator_candidates.py` | +147/−56 | **Partly both.** The `apply_trellis_surface` gate and the `trellis_menu.augment_candidates` call are dead — `a212b73` rewrote that exact block into a refusal. `_super_menu_format_names` / `_registry_free_menu_names` and the registry-free pricing in both aggregators are **live and default-path**, but **latent**: see below. |
| `prismaquant/allocator_solver.py` | +26/−3 | **Live, salvageable.** The `rank_of()` wrapper is a pure diagnostics improvement, no trellis dependency. |
| `tests/test_super_item_menu_byte_identity.py` | +616 (new) | **Live, salvageable, and measured to still apply.** Nothing trellis-specific. |
| `tests/fixtures/super_item_menu_golden.json` | +820 (new) | **Live, salvageable as-is** — no regeneration needed, see below. |
| `docs/ARCHITECTURE.md` | +142/−56 | **Superseded.** `a212b73` banner-dated §4.9 rather than deleting it. The only conflict. |

No live importer of `trellis_formats` or `trellis_menu` survives: `git grep
"trellis_formats\|trellis_menu" a212b73 -- prismaquant` returns two prose lines in
`tessera_formats.py`'s header and nothing else.

**On links 3 and 4 being "live default-path value":** true of the file, latent in
effect. The drop this PR fixes only bites a menu format with no `FormatSpec`, and
neither `origin/main` nor `a212b73` has a producer of one — on `a212b73` Tessera
rungs enter `fmt_names` through `expand_menu_tokens_report`
(`allocator.py:2454-2470`) and resolve to specs synthesized on demand
(`format_registry.py:1390-1393`), so the aggregators' `for spec in formats:` never
drops them. Tessera took the *opposite* architectural
branch from this PR's links 1/2/6: synthesize the spec rather than refuse to.
Worth landing as robustness, not as a bug fix.

### The eighth link: what it found outlives the branch

`UNWIRED_LINKS` kept exactly one entry, enforced by `_require_run_currency`
(`prismaquant/trellis_menu.py:312` on this branch). The finding: the ladder's
anchors were **weighted SSE under a per-input-channel activation second moment**,
an output-MSE proxy, while the production DP ranks `predicted_dloss` in the
**AURA KL-adjoint** — so a DP ranking a weighted-SSE rung against an AURA-priced
NVFP4 rung solves neither objective, and no plumbing changes what the anchors
measured. The mechanism was the good part: the expected currency was derived
*definitionally* from `COST_MODE_OBJECTIVE_CURRENCY` over the same
`COST_RENDER x COST_OBJECTIVE` decomposition `run-pipeline.sh:271-294` resolves
(P2), read from the **attested** `provenance['cost_mode']` and never from
`os.environ` — because `run-pipeline.sh:298` assigns `COST_MODE` with `:=` and
never exports it (`grep -c "export .*COST_MODE"` → 0). Two adjacent refusals came
with it: an unstamped cost table, and a cost mode naming no objective.

**It applies to the successor, and nothing checks it.** (Scope: this is on
`tessera/decouple-gridbook` at `a212b73`, the branch carrying the archival — not
on `origin/main`, which has no `tessera_campaign.py` at all. It is a finding about
the tree that is about to become main, not about today's default path.)
`prismaquant/tessera_campaign.py:72` declares
`CURRENCY = "output_mse_under_route_activation_contract"` and stamps it on every
row (`:366`, `:416`, `:437`), and its header says outright
(`tessera_campaign.py:35-41`) that this is the render-cost currency and *"not the
AURA adjoint... The payload declares its own currency so nothing downstream can
mistake one for the other."* Nothing downstream reads it — repo-wide, not just on the
paths you would guess: `git grep -n '\["currency"\]\|get("currency"' a212b73 --
prismaquant` returns **nothing**, and `git grep -n '"currency"' a212b73 --
prismaquant` outside `tessera_campaign.py` itself returns nothing. The only
currency check in the allocator is `ANCHORED_AURA_COST_CURRENCY`
(`allocator_candidates.py:65`) against a *different* field, `cost_currency`, on a
different path (`:1399` on `a212b73`, `:1286` on `origin/main`, anchored stock
rows only). The pipeline
default is `COST_MODE=aura` (`run-pipeline.sh:298`). So on that branch the
mismatch this PR refused for Gridbook is unguarded for Tessera.

That is distinct from `tessera_menu.surrogate_selection_caveat()`
(`tessera_menu.py:1214-1244`), which records a measured *slope* error inside a
monotone surface — served KL 2.00x worse than a byte-matched uniform arm at
4.0 bpp. A currency mismatch is an error of *units*, not of slope. The two may
well compound; nothing in the tree separates them.
`archive/trellis_wire_2026-09-02/README.md:69-70` already records the trellis half
of the lesson. It does not record that the successor inherited the mismatch and
lost the refusal. That is the gap worth filing.

### The golden is worth rescuing, and it still applies — measured

Copied `tests/test_super_item_menu_byte_identity.py` and the fixture from
`6e4092b`, unchanged, onto detached worktrees at **both** `origin/main`
(`55d43e3`) and `a212b73`. Identical result on both — `3 passed, 1 failed`:

- `test_super_item_menus_are_byte_identical_to_the_golden` **passes**. The fixture
  was recorded against `1ca4c7d`-era aggregators and still matches both trees
  exactly — including `a212b73`'s, which have since gained
  `tessera_group_composites`, `_fused_member_formats` and `_tessera_group_menu`
  (`allocator_candidates.py:2967-3045`). **No regeneration needed on either** —
  which matters, because a regenerated fixture stops being evidence.
- Numeric tooth, run standalone: clean = `0 problems`; with every member's
  `predicted_dloss` scaled by `1 + 1e-12` = `56 problems`. Same on both refs. The
  16-ulp bound is a real gate on this tree too.
- `test_the_golden_is_load_bearing` fails on its *structural* tooth only, which
  monkeypatches `_super_menu_format_names` — a helper this PR introduces. Without
  the PR it must perturb the aggregators' `for spec in formats:` order instead.
  Mechanical.

The py3.11 diagnosis in the comments above is worth keeping as record on its own:
the 1–2 ULP divergence was CPython 3.12's Neumaier compensated `sum()`
(gh-100425) against 3.11's naive reduction over a 9-member sum — not numpy, not
serialization — and the fix loosened the *bound*, derived from
`sys.float_info.epsilon`, without touching the fixture.

### Recommendation

Close unmerged once `a212b73` reaches `origin/main`, and file three follow-ups:

1. **Nothing reads the currency a Tessera cost table declares.** Port
   `_require_run_currency`'s mechanism — a definitional `COST_MODE` → objective
   currency table, read from the attested `provenance['cost_mode']`, with the
   unstamped-table and unnamed-objective refusals — and decide the Tessera
   disposition deliberately rather than by silence.
2. **Rescue the super-item byte-identity golden** standalone, minus trellis,
   fixture unchanged, structural tooth re-pointed.
3. **Lift the `rank_of()` diagnostic** from `allocator_solver.py`.

Plus: append the generalised currency lesson to
`archive/trellis_wire_2026-09-02/README.md`. The durable part is not "the Gridbook
anchors were weighted SSE" — it is that a rate-surface anchor set carries a
currency, the DP ranks in one currency, and the two have to be compared against
the run's attested cost mode. This repo's graveyard is a lesson, not a deletion.

The Muse Spark close-out comment above reaches the same disposition and the same
three salvage items. Two notes on it: it cites #118 as a *proposal* and does not
engage `a212b73`, the commit that actually landed the archival on
`tessera/decouple-gridbook` — worth pinning, since the archival's landing state is
the one fact that changes the timing of this close. And its "only piece with live
default-path value" for links 3/4 should read *latent* — see the note in the table
above. Its #125 point stands: that issue removes the `cost_mode=` kwarg link 7
consumes, and Issue 1 above wants link 7's *idea*, not its kwarg.

RobTand:
Closing as superseded: Rob retired the Gridbook/NVFP4-CB lane on 2026-09-02 and the five `trellis_*` modules this PR wires are archived at `archive/trellis_wire_2026-09-02/` on the integration branch. The three salvageable pieces named above — the `rank_of` diagnostic, the byte-identity golden (which was measured to pass unmodified on both trees), and the currency gate the eighth link was refusing for — are being filed as issues so they outlive the branch.

PR #94, opened 2026-08-30
fix(tests): glm5_next parity NaN — the polyfill leaves from-config expert weights uninitialized

RobTand:
## Verdict

**Not a sparse-attention defect.** The `dsa_only` NaN in run 33284249771 is an
**import-order / polyfill artifact that only affects tests** (category (c)),
manifesting through a gap in the test's own weight-repair list (a (b) component).
Nothing on a serving path changes; the streamed forward is exonerated by
measurement, not by argument.

## Where the NaN is born

`prismaquant/__init__.py::_polyfill_transformers` replaces
`transformers.modeling_utils.PreTrainedModel._initialize_weights` with a global
no-op (before this PR, `prismaquant/__init__.py:106-107`). That is sound for every
PrismaQuant load path — they build a `from_config` skeleton and immediately
overwrite every parameter from the checkpoint — but it is *not* sound for a caller
that then **uses** the from-config model.

`nn.Linear` / `nn.Embedding` initialize themselves in their constructors, so most
of the model is fine. The tensors a modeling file allocates as a bare
`nn.Parameter(torch.empty(...))` are not: their only initializer is
`_init_weights`, which the no-op prevents from running. In GLM-5.3-Flash those are
the routed-expert weights —
`transformers/models/glm5_next/modeling_glm5_next.py:1394`
(`init.normal_(module.gate_up_proj, ...)`) — so
`model.language_model.layers.N.mlp.experts.gate_up_proj` and `.down_proj` ship
whatever the allocator last left in those pages.

Measured, tiny config, `dsa_only`:

| build | non-finite / huge params |
|---|---|
| without `import prismaquant` | 0 |
| with `import prismaquant` (`init_noop: True`) | 8-10 params at ~1.8e17-2.0e17, incl. `layers.1.mlp.experts.gate_up_proj` (1.99e17) and `.down_proj` (1.80e17); some runs non-finite outright |

`tests/test_glm5_next_streamed_forward_parity.py` already knew the model came out
under-initialized: it re-randomized parameters matching a hand-listed suffix set
`(".fn", ".base", ".scale", ".dt_bias", ".A_log")`. That list does not contain the
expert tensors. Under
`torch.utils.deterministic.fill_uninitialized_memory = True` the miss is exact and
reproducible: 14 uninitialized tensors (dsa/dsa), 18 (kda/kda), 16 (mixed), of
which exactly **2 per variant** fall outside the repair list — always the two
routed-expert tensors.

## Why this is not a wiring bug (ruling out (a))

The 2x2 of `fill_uninitialized_memory` x suffix-repair:

| fill | repair | reference forward | streamed forward | worst divergence | result |
|---|---|---|---|---|---|
| True | False | **NaN** | NaN | `nan` | FAIL |
| True | True | finite | finite | `0.000e+00` | PASS |
| False | * | finite | finite | small | PASS |

`ref_nan=True` is the exoneration: the *reference* upstream forward NaNs too, so
the assertion at `:363` was comparing garbage to garbage. `nan != nan` on both
sides is indistinguishable from a wiring defect, which is exactly how it read.
Corroborating: the CI divergence table reads `nan` at **every** context octave
including `1-1` — no long-range or sparse-index defect can NaN position 1.

## Why it looked like a regression (it is not)

- Both CI jobs installed identical `torch 2.13.0+cpu` and `transformers 5.16.1`;
  the only library difference between the green py3.12 job and the red py3.11 job
  is numpy (2.4.6 vs 2.5.2). No library-version explanation survives.
- The two commits merged between `e05f4b2` and `1ca4c7d` are **exonerated**:
  `c198593` (a tolerance derivation in a test file) and `3921f2b` (re-stamped
  SHA256 constants) can only perturb the heap, which is the whole mechanism.
  They are also why the visible failure count *changed*: on `e05f4b2` both
  interpreters failed the same 3 tests, and those two commits fixed exactly those
  3 — leaving the pre-existing latent flake as the only line left standing.
- Reproduced identically on `e05f4b2`: same 5 failed / 3 passed, same 5 test ids.
  This is latent, not new.

## Reproduction

py3.11, `torch 2.13.0+cpu`, `transformers 5.16.1`, `PYTHONPATH=.`,
`CUDA_VISIBLE_DEVICES=""`. `scratch/pytest_fill.py` is a 3-line wrapper that sets
`torch.utils.deterministic.fill_uninitialized_memory = True` and delegates to
`pytest.main(sys.argv[1:])` — the deterministic stand-in for whatever the py3.11
CI heap happened to contain.

```
# FAILS on 1ca4c7d and on e05f4b2 -> 5 failed, 3 passed
python scratch/pytest_fill.py tests/test_glm5_next_streamed_forward_parity.py -q -p no:cacheprovider

# PASSES on the same commits -> 8 passed
python -m pytest tests/test_glm5_next_streamed_forward_parity.py -q -p no:cacheprovider
```

That divergence **is** the flake. On this branch both commands give `8 passed`.

## The fix

Two halves, no tolerance touched (`worst < 1e-3` is unchanged — NaN is not a
tolerance problem):

1. `prismaquant/__init__.py` — the polyfill now **preserves** the real method as
   `PreTrainedModel._prismaquant_real_initialize_weights` before installing the
   no-op, and exports a `genuine_weight_initialization()` context manager that
   restores it for the duration of a construction. The no-op itself is unchanged,
   so no PrismaQuant load path changes behaviour. (Restoring the real
   `_initialize_weights` and letting `post_init()` drive it is the only correct
   form: calling `_init_weights` manually binds the wrong config and raises
   `AttributeError: 'Glm5NextConfig' object has no attribute 'initializer_range'`.)
2. `tests/test_glm5_next_streamed_forward_parity.py` — the model build, previously
   duplicated in two places, is now one helper that builds inside the context
   manager and ends with a **fail-closed guard**: any non-finite parameter or
   buffer aborts with a message saying both sides of the comparison are garbage,
   rather than letting it surface as `nan` divergence. The suffix re-randomization
   is kept (it perturbs constants `_init_weights` deliberately sets, so a wrong
   wiring cannot agree with a right one by accident) but is no longer load-bearing
   for finiteness.

`tests/test_genuine_weight_initialization.py` (new) covers preservation, swap,
restore-on-exception, and the behavioural check that `_init_weights` actually runs
inside the CM and does not outside it. Mutation-checked: deleting the two
preservation lines from `prismaquant/__init__.py` fails 3 of the 4 new tests.

## Out of scope

On this aarch64 box a 14-file model-building subset shows 7 failures **with and
without** the amplifier, present on `e05f4b2` too, and the two spot-checked files
pass when run alone. That is the pre-existing order-dependent set already recorded
as `suite_has_order_dependent_failures`; unrelated to this NaN, not touched here.

### DIFF
--- a/prismaquant/__init__.py
+++ b/prismaquant/__init__.py
@@ -12,6 +12,8 @@ Current production path:
 Older cross-layer allocators are archived under archive/cross_layer_2026-05-09
 for artifact replay and comparison.
 """
+import contextlib as _contextlib
+
 from .format_registry import FormatSpec, REGISTRY, register_format
 
 # Resolved from installed metadata rather than duplicated here, so
@@ -100,9 +102,21 @@ def _polyfill_transformers() -> None:
         # rotary modules to expose `compute_default_rope_parameters`,
         # which older remote modeling files (MiniMax M2/M2.7) don't
         # provide. No-op it globally at import time.
+        #
+        # The no-op is only sound for callers that overwrite every
+        # parameter afterwards. A caller that *uses* a from-config model
+        # -- a test fixture, an eval harness building a random reference --
+        # gets whatever the allocator last left in the pages backing the
+        # parameters a modeling file allocates with a bare `torch.empty()`
+        # (transformers' own `nn.Linear`/`nn.Embedding` still self-init in
+        # their constructors; the raw `nn.Parameter(torch.empty(...))` ones
+        # do not). Keep the real method reachable so such a caller can
+        # restore it -- see `genuine_weight_initialization` below.
         import transformers.modeling_utils as _mu
         if hasattr(_mu, "PreTrainedModel") and \
                 not getattr(_mu.PreTrainedModel, "_prismaquant_init_noop", False):
+            _mu.PreTrainedModel._prismaquant_real_initialize_weights = (
+                _mu.PreTrainedModel._initialize_weights)
             _mu.PreTrainedModel._initialize_weights = (
                 lambda self, *a, **kw: None)
             _mu.PreTrainedModel._prismaquant_init_noop = True
@@ -161,6 +175,45 @@ def _polyfill_transformers() -> None:
         _register_qwen3()
     except Exception:
         pass
+
+
+@_contextlib.contextmanager
+def genuine_weight_initialization():
+    """Build a model the way transformers would, inside a PrismaQuant process.
+
+    `_polyfill_transformers` no-ops `PreTrainedModel._initialize_weights`
+    for the whole process at import time, because every PrismaQuant load
+    path builds a `from_config` skeleton and then overwrites every
+    parameter from the checkpoint. That makes `from_config` construction
+    silently return **uninitialized** parameters for any tensor the
+    modeling file allocates as a bare `nn.Parameter(torch.empty(...))` --
+    routed-expert weights and hyper-connection tensors are the common
+    cases -- and heap contents are neither reproducible nor guaranteed
+    finite, so a forward on such a model can be fine on one run and `nan`
+    on the next.
+
+    Wrap a from-config construction in this context manager when the model
+    it returns is used as-is:
+
+        with genuine_weight_initialization():
+            model = SomeForCausalLM(config)
+
+    A no-op when the polyfill never applied (older transformers, or an
+    import that raised), so callers need not test for it.
+    """
+    import transformers.modeling_utils as _mu
+
+    real = getattr(
+        _mu.PreTrainedModel, "_prismaquant_real_initialize_weights", None)
+    if real is None:
+        yield
+        return
+    patched = _mu.PreTrainedModel._initialize_weights
+    _mu.PreTrainedModel._initialize_weights = real
+    try:
+        yield
+    finally:
+        _mu.PreTrainedModel._initialize_weights = patched
 
 
 _ensure_triton_cache_writable()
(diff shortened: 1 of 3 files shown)


RobTand:
Verification:

- CI on this branch: **all four checks pass**, including `tests (py3.11, cpu)` (20m42s) and `tests (py3.12, cpu)` (20m37s) — run 33285973067.
- Local full suite (aarch64, py3.11, torch 2.13.0+cpu, transformers 5.16.1): `4 failed, 5481 passed, 154 skipped, 3 xfailed` in 18m41s. The 4 are `test_rtx4090_fp8_burn.py::test_allocate_uses_exact_sealed_bundle_probe_through_child` and three `test_run_pipeline_gates_execute.py` gate tests; they fail **identically on the base commit 1ca4c7d** in a fresh worktree with the same interpreter, in isolation as well as in-suite, and are green on x86 CI. Box-specific and pre-existing; not touched here.
- Targeted: `tests/test_glm5_next_streamed_forward_parity.py` gives `8 passed` both with and without the `fill_uninitialized_memory` amplifier on this branch, against `5 failed, 3 passed` under the amplifier on 1ca4c7d and on e05f4b2.

Note for whoever is on `ci/drop-py311`: this flake is not interpreter-specific. The trigger is heap contents at `from_config` time, so dropping the py3.11 job would hide it rather than fix it — the py3.12 job could surface the same NaN on any future commit that perturbs allocation.

PR #95, opened 2026-08-30
Land the proven rescues: walker R5 name projection, grouped wo_a Fisher, unit-sharded producer, TP=2 tooling

RobTand:
## What this lands

`merge/proven-rescues` is 31 commits ahead of `main` (20 non-merge, 52 files,
+11,039 / −356). It is the accumulated "proven rescues" work plus
`origin/main` (`1ca4c7d`) merged in. Five themes:

**1. Walker R5 — a shared name-projection layer.** `prismaquant/name_projection.py`
becomes the single place that maps a Linear's name between the five namespaces
(recipe / checkpoint / live model / export / vLLM-internal), and all four
consumers are migrated onto it: read-traffic (`f64b861`), footprint (`25b82c0`),
probe (`2a49363`), cost (`ef4455b`). Plus a fail-closed discovery-walker export
gate (`1dca4c2`, `prismaquant/model_walk.py`).

**2. The grouped operand `attn.wo_a` is priced** (`a2e1ae1`). DSv4's
`DeepseekV4GroupedLinear` was invisible to the Fisher probe; this adds a
grouped-BMM accumulator and a declarative spec field
(`probe.grouped_module_class_names`) so a grouped module is priced rather than
silently skipped.

**3. Parallel producer v1 — unit-sharded production-cache renders** (`5a8303b`,
`eb06a72`): `PRISMAQUANT_UNIT_SHARD=i/N`, `prismaquant/unit_sharding.py`,
`tools/merge_unit_shards.py`, and a cluster render sentinel that **refuses a
partial render instead of certifying it** (`6a66cf7`, `fee7a7f`).

**4. TP=2 cluster tooling**: a decode-feasibility harness for the two-Spark
10 GbE cluster (`7e5f888`) and a budget planner (`3a9061d`).

**5. Correctness fixes**: CB warm-state LDLQ identity scoped per format
(`373bf1d`, this is PR #84, which was merged *into* this branch rather than
retargeted, because retargeting would have pulled 28 commits for a 2-file fix);
the probe no longer refuses a profile over an optimization-only field
(`db8d3ae`); read-traffic provenance no longer names the atomic staging dir the
export destroys (`7fee52b`).

## Test evidence (CPU suite, py3.12, `pytest tests/ -q --timeout=300`)

| tree | result |
|---|---|
| `origin/main` @ `1ca4c7d` | **5490 passed**, 143 skipped, 3 xfailed, **0 failed** |
[...]

PR #99, opened 2026-08-30
feat(model-walk): the discovery walker as a fail-closed export gate

RobTand:
Split 1 of 6 from `merge/proven-rescues` (#95). **This is the Walker.** Land this one first — three of the other five stack on it.

## What it does

Closes the remaining open half of the R5 design contract: the discovery walker is now wired as a **fail-closed export gate**, and the walk itself became a cacheable artifact.

`run-pipeline.sh` stage **[3d]** (`python3 -m prismaquant.model_walk`) runs before **every** export lane (GGUF, CB, compressed-tensors) and refuses with `exit 2` on:

- an unclaimed matmul-fed node,
- an unresolved floating multiplicand,
- an unknown walk-failure kind,
- a decided-but-unpriced contradiction.

It decides from **structured** fields only (`artifacts/model_walk.json`, gate schema `prismaquant.model_walk_gate.v1`). Prose `detail`/`refusal_reason` fields explain to humans; nothing branches on them (principle 14).

`PRISMAQUANT_WALK_GATE_OVERRIDE=<reason>` excuses **trace incompleteness only** — DSv4's data-dependent scalar aborts the fake trace — and is stamped into the verdict. Claim failures have no override: they are fixed by pinning with a reason in `ModelProfile.walk_claim_rules()`.

Same change: walks are persisted under `SCHEMA prismaquant.model_walk.v1` with an atomic write and a fail-closed reload (foreign schema, execution mismatch, or claim-rule digest mismatch all refuse), carrying trace-time provenance — model and config identity, package versions, input contract — plus the serialized applied rule list. Byte semantics are pinned as LOGICAL-TOTAL behind a `per_device_bytes(total, tp_degree, policy)` seam for tensor parallelism.

The R5 profile sweep's findings land in the universal base rules: MoE router gates are pinned and packed expert stacks are decided. Six or seven profiles had them unclaimed, and gemma4's router was **decided-but-never-priced** — the wrong polarity inside the claim table itself. The gate turns that contradiction class into a refusal so it cannot recur silently.

## Why it is worth landing

This is the piece that finds eligible tensors. Before it, a matmul-fed weight nobody claimed simply shipped at source precision and nobody was told. `find_decided_but_unpriced` is the part that matters most: it catches the class of bug where the claim table and the pricing path disagree, which is how `wo_a` stayed unpriced (see #103) and how gemma4's router was mis-polarised.

`prismaquant/model_walk.py` is **not** a new module — `origin/main` already imports it from `model_profiles/base.py`, `deepseek_v4.py` and `glm5_next.py`. This is a `+1094/-4` extension of live code, so I checked the blast radius explicitly:

- The **only** symbol those three consumers import is `ClaimRule`. Its dataclass body and `apply_claim_rules` are **byte-identical to main** (verified by diffing the extracted definitions).
- The four removed lines are a field comment, a signature rewrap, and a guard hoist — no removed behaviour.
- `model_profiles/base.py` gains exactly two appended `ClaimRule`s (between the `ndim<=1` exclude and the final Linear-decide rule) plus a `9 -> 11` docstring renumber. Nothing else in the file moves.
- `pipeline.py` (+44) and `run-pipeline.sh` (+41) add the stage; no existing stage changes.

## Dependencies

- **None.** This is the base of the stack.
- Depended on by: `rescue/name-projection` (hard — three `test_name_projection.py` tests build a walk that only claims packed expert stacks after this PR's rule) and `rescue/grouped-woa` (hard — its tests extend `test_model_walk_export_gate.py`, and its frozen digest covers this PR's `base.py`).

## Tests

They **do not reproduce** — #90 and #94 already fixed them — so every failure below is attributable. (Note for the record: the "4 pre-existing aarch64 failures" that were flagged to me do **not** reproduce on this commit — #90 and #94 fixed them. Any failure below is attributable.)

This branch: **2 failed, 5522 passed**, 143 skipped, 3 xfailed, 179 subtests passed (20m20s, `-p no:randomly`, `CUDA_VISIBLE_DEVICES=""`). Both failures are the DSv4 W8A16 freeze gate documented below; **everything else is green**, including `tests/test_run_pipeline_gates_execute.py` and both doc gates.

## ⚠️ Known red: the DSv4 W8A16 freeze gate

`tests/test_dsv4_w8a16_export_handoff.py` fails 2 tests, naming exactly one file:

```
frozen exporter/source closure changed (1 of 15 file(s)):
prismaquant/model_profiles/base.py;
observed=7cff3a4af253777d831094838a35ec56b9cd5e4c1022654431449163be8a848e,
expected=147699331599870a8ba153ae5132f5bda1f32f5ba4e0298a784c3084df05207a
```

**This is the gate working, not a defect in the split.** `base.py` is in the frozen export-source closure, so any change to it must be re-frozen *with review*. I did not re-stamp it: blind re-stamping is precisely what the gate exists to prevent, and the sandbox blocked me from editing that region anyway.

The review already exists. It was written on `merge/proven-rescues` (2026-08-22) and I **re-verified it against these rebased bytes**:

> `walk_claim_rules()` appended two ClaimRules between the `ndim<=1` exclude and the final Linear-decide rule (a router-class pin and an `*Experts` packed-stack decide; docstring renumbered 9→11). Nothing else in the file moved. The method's only consumer is the R5 discovery walker and its tests — a grep over this closure finds no other caller, so exporter, completeness, decode-source, and namespace behaviour cannot observe the change. On the DSv4 walk itself the rules are strictly shadowed or equivalent: `DeepseekV4Profile` prepends its own per-class router pins ahead of `super()`, first-match keeps the original disposition and reason for every node both rules match, and the R5 sweep verified every router-family node on the vendored topology was already pinned; the packed-expert decide matches owner classes containing "expert", but this lane swaps packed classes for per-expert `nn.Linear`s before instantiation, whose weights the Linear rule already claimed with the same disposition. Net: byte-identical claims on the shipped DSv4 topology; additional coverage only on architectures that previously had NO claim at all.

Re-verification on the rebase: `git diff origin/main -- prismaquant/model_profiles/base.py` on this branch is exactly the two appended ClaimRules plus the docstring renumber, and no other hunk.

**To clear it:** set `prismaquant/model_profiles/base.py` to `7cff3a4af253777d831094838a35ec56b9cd5e4c1022654431449163be8a848e` in `_FROZEN_EXPORT_SOURCE_SHA256` and add the review note above. One edit, reviewer's call.

---

Extracted from `merge/proven-rescues` commit `1dca4c2`. `merge/proven-rescues` is left in place as the record.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,38 @@
 # PrismaQuant Architecture
 
-As of: 2026-08-29 · `docs/architecture-currency-20260829` — stamps follow, newest
+As of: 2026-08-29 · `rescue/walker-export-gate` — stamps follow, newest
 first, each recording its own branch and date. Re-stamped (2026-08-29,
+`rescue/walker-export-gate`) for
+**the discovery walker as a fail-closed export gate** (§8.8,
+`prismaquant/model_walk.py`): the R5 design contract's remaining open half —
+wiring the walk as an export gate — is closed; the probe/cost/footprint/
+read-traffic migration onto its edge list stays a separate workstream.
+`run-pipeline.sh` stage **[3d]** (`python3 -m prismaquant.model_walk`) runs
+before EVERY export lane (GGUF, CB, compressed-tensors) and refuses (`exit 2`)
+on an unclaimed matmul-fed node, an unresolved floating multiplicand, an
+unknown walk-failure kind, or a decided-but-unpriced contradiction, deciding
+from STRUCTURED fields only (`artifacts/model_walk.json`, gate schema
+`prismaquant.model_walk_gate.v1`). The explicit override
+(`PRISMAQUANT_WALK_GATE_OVERRIDE=<reason>`) excuses trace incompleteness ONLY
+(DSv4's data-dependent scalar aborts the fake trace) and is stamped; claim
+failures have no override — they are fixed by pinning with reasons in
+`ModelProfile.walk_claim_rules()`. Same commit: the walk became a cacheable
+artifact (`SCHEMA prismaquant.model_walk.v1`; envelope `{schema, provenance,
+result}`, atomic write, fail-closed reload on foreign schema / execution
+mismatch / claim-rule digest mismatch), carrying trace-time provenance
+(model+config identity, versions, input contract) and the serialized applied
+rule list; byte semantics are pinned as LOGICAL-TOTAL with the
+`per_device_bytes(total, tp_degree, policy)` seam for Tensor-Parallel
+(decision unit: whole logical tensor; dispositions are TP-invariant; the
+future group/shard misalignment refusal lands as an unknown-until-implemented
+walk-failure kind that already refuses). The R5 profile sweep's findings land
+here too: universal base rules now pin MoE router gates and decide packed
+expert stacks (six/seven profiles were unclaimed; gemma4's router was
+decided-but-never-priced — wrong polarity inside the claim table itself, now
+pinned like every other router), and the gate surfaces that contradiction
+class structurally.
+
+Earlier stamp (2026-08-29,
 `docs/architecture-currency-20260829`) for a **§8.4 conformance-matrix
 correction on `glm5_next`**, found by a principle-13 sweep of `origin/main`
 rather than by a test — the two doc gates were green across both drifts. (i) The
@@ -19,8 +50,9 @@ quantized formats outside the three module families PR #53906 wires a
 `b4a8846` refresh, so it is the § P13 "currency is not truth" case: this file
 was re-stamped in lockstep while carrying a false statement about a serving
 default. Nothing else in the window crossed the trigger; the reasoning is in the
-commit message. Re-stamped (2026-08-29,
-`claude/trellis-continuous-surface`) for the **trellis seam correction**
+commit message.
+
+Earlier stamp (2026-08-29, `claude/trellis-continuous-surface`) for the **trellis seam correction**
 (§4.9): the seam that landed earlier the same day is now **fail-closed** when
 enabled, and this document's claim that it made trellis rungs "pass the same
 legality, aggregation and byte accounting every other candidate does" is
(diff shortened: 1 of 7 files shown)


RobTand:

Sequence: this is the stack base, lands first, no predecessor. Currently DIRTY/CONFLICTING against main, so rebase before merge:
git fetch origin
git checkout rescue/walker-export-gate
git rebase origin/main
Retarget already main, no retarget needed.

Freeze owed: exactly 1 file, prismaquant/model_profiles/base.py, set to 7cff3a4a in _FROZEN_EXPORT_SOURCE_SHA256 per PR body. Review must confirm the diff vs origin/main is only the two appended ClaimRules between the ndim exclusion and the final Linear-decide rule plus the 9-to-11 docstring renumber, ClaimRule and apply_claim_rules byte-identical for existing consumers, and DSv4 claims byte-identical because DeepseekV4Profile prepends its own router pins and packed-expert weights are already Linear-claimed. Do not stamp blind; that check is the gate.

ARCH stamp: docs/ARCHITECTURE.md provenance block will conflict with each later PR by design. Resolution: keep both stamps, newest first. Zero code conflicts expected.

RobTand:
**Merged locally onto the integration branch `tessera/decouple-gridbook` at `0815a00`** — rebuilt on a base eight commits newer than the PR's, every conflict resolved on the merits. Not pushed; that is Rob's call.

**One conflict, and it was the whole point of the merge.** This branch grew a Tessera export arm after #99 opened, and that arm ends in `exit 0`. Taking HEAD's ordering puts the Tessera lane first, so a Tessera run would exit before the gate ever ran — the gate silently stops covering a lane while still claiming to cover every lane, and nothing fails. Resolved by putting [3d] ahead of Tessera (`:2372`), GGUF (`:2435`) and compressed-tensors (`:2517`) alike.

**Then the ordering test turned out to be blind to exactly this.** `test_run_pipeline_invokes_the_gate_before_every_export_lane` kept a literal list of three invocations. It was stale in both directions at once: it still named `CB_EXPORT_ARGS` after the Gridbook lane was retired, so it failed on a lane that no longer exists — and it had never heard of the Tessera lane, so it would have gone on passing while the gate stopped covering one. Rewritten in `90cbcb8` to derive the lane set from `run-pipeline.sh` itself, and **verified to have teeth rather than assumed**: moving the gate below the Tessera arm now gives `AssertionError: the walk gate must precede the tessera export lane`.

`docs/ARCHITECTURE.md:269` named the covered lanes as '(GGUF, CB, compressed-tensors)'. Two of three were wrong by merge time; corrected to the sanctioned three.

39 passed, 1 skipped across model_walk, docs-staleness and architecture-doc.

RobTand:
## Closing: already landed on the integration branch

Checked rather than assumed. Every distinctive symbol this PR introduces is present on `tessera/decouple-gridbook` (the branch every merge has gone to since):

- `prismaquant/model_walk.py` is **byte-identical** between this branch and the integration branch (same blob sha).
- `prismaquant.model_walk_gate.v1` → `prismaquant/pipeline.py`, `prismaquant/model_walk.py`
- `PRISMAQUANT_WALK_GATE_OVERRIDE` → `pipeline.py`, `model_walk.py`, `run-pipeline.sh`
- `find_decided_but_unpriced` → `model_walk.py`, `tests/test_model_walk_export_gate.py`
- stage [3d]: `run-pipeline.sh` carries 4 `model_walk` call sites.

The work was landed by cherry-pick, so the commit shas differ and GitHub could not close this automatically. The files that still show as differing (`pipeline.py`, `tests/test_model_walk_export_gate.py`) differ because *later* PRs have edited them since 2026-08-30 — that is drift on the integration branch, not content missing from it.

## Why it looked stuck

This PR targets `main`, and **`main` is 187 commits behind `tessera/decouple-gridbook`** (0 commits ahead — it would fast-forward). That is the whole reason this and its two stacked PRs read as open-and-confusing: their base is a branch nothing has merged to in a long time. Raised separately with Rob; nothing to do inside this PR.

Reopen if any part of the gate turns out to be missing — the containment check above is the thing to re-run.

PR #100, opened 2026-08-30
refactor(names): one shared name-projection layer, and all four consumers on it

RobTand:
Split 2 of 6 from `merge/proven-rescues` (#95).

> **Merge order: this stacks on #99 (`rescue/walker-export-gate`), and the dependency is hard, not cosmetic.** Base is set to that branch; GitHub will retarget to `main` when it merges.

## What it does

The R5 consumer analyses each independently found the same thing: probe, cost, footprint and read-traffic were re-deriving the mapping between the pipeline's parameter-name namespaces ad hoc, with private string surgery in four places that could — and did — disagree.

`prismaquant/name_projection.py` is the one projection between a Linear's **live** name (`WalkNode.name`), its allocator/probe **recipe** name, its source-**checkpoint** key, its **exported**-artifact key, and its **vLLM** scheme-dispatch qname. The contract:

- **The profile owns the knowledge; the layer owns the discipline.** Every mapping routes through existing `ModelProfile` accessors. No second mapping lives anywhere, and consumers keep no private `_recipe_name` / `_strip_weight` helpers.
- **Fail closed and loud.** An unmappable or ambiguous name raises `NameProjectionError` with structured fields — `code`, source/target namespace, and what was tried. Unlike `decision_units.fused_group_key`, nothing swallows a profile failure into an identity fallback. The profile's *declared* drops are data instead: `project()` returns a `ProjectedName` whose `outcome == declared_out_of_graph`, so read-traffic classifies `excluded_non_text_graph` by branching on a field.
- **Non-totality is in the type.** Fused siblings and packed experts surface as `ServingGroup(kind, key, members)`; reverse lookups refuse group KEYS rather than returning an arbitrary member.
- **Round-trip property tested where the profile's rules are total** — Qwen3 dense, Qwen3.5 multimodal, DSv4 — plus the gap the round trip *surfaced*: DSv4's spec generates `hc_head.hc_*` source spellings while the real checkpoint stores flat `hc_head_*`, so that family's inverse declines. The layer reports it; it does not guess.
- **TP seam held.** No rank, shard index or degree exists in any signature (pinned by a constructor-kwarg refusal test). Byte accounting stays in `model_walk.per_device_bytes`.

All four consumers are migrated in this PR, because a shared layer with no consumers is not a refactor, it is a second mapping.

## Why it is worth landing

The migrations are not cosmetic. Two of them close silent-wrong-answer paths:

- **read-traffic**: a failing profile accessor used to pass the raw key through, which **re-priced unmappable tensors at p=1**. It now refuses.
- **probe**: the packed-expert shard-scope filter read its block id from a positional `[:3]` qname slice, and the Fisher skip set keyed its embedding clause on a hardcoded `"model.embed_tokens"` substring test. Both now go through the profile's declared accessors.

All four are live pipeline stages: `read_traffic.py`, `footprint.py`, `sensitivity_probe.py`, and the cost path (`allocator_candidates.py`, `decision_units.py`, `measure_quant_cost.py`, `production_render_cost.py`).

**Emitted values are unchanged.** Bytes, class tables, byte authorities, refusals, probe stats keys, inventories and shard regexes are pinned byte-identical test-side. The manifest's `profile=None` convention now builds over the repo's declared generic baseline (`DefaultProfile` — the substitution `resolve_cost_target_name` always made) instead of inlined string surgery.

One deliberate behaviour change, **fail-closed only**: a profile accessor that raises now propagates `NameProjectionError` instead of silently skipping the row.

One exception stays and is **not** a projection: `production_render_cost.canonical_cost_name`'s umbrella-infix half remains a **total** normalizer, because render/cost payloads key costed MTP rows physically and a projection that may decline must never drop them.

Sharing a name projection is not sharing a requirement set. Every consumer's INVENTORY is still its own enumeration, so the edge-list migration proper stays open for all four.

## Dependencies

- **Hard dependency on #99.** I initially built this off `main` and the tests found it: three `test_name_projection.py` tests (`test_packed_expert_stack_groups_by_role`, `test_split_checkpoint_spells_cover_the_packed_aggregate`, `test_coverage_queries_fail_closed`) build a walk over a packed-expert fixture and raise `WalkError` on `main`, because `walk_claim_rules()` does not claim packed expert stacks until #99 adds that rule. `name_projection.py` also cites `load_walk` and `per_device_bytes` in its module docstring, both of which #99 introduces.
- No other PR in this split depends on this one.

## Tests

They **do not reproduce** — #90 and #94 already fixed them — so every failure below is attributable.

This branch: **2 failed, 5568 passed**, 143 skipped, 3 xfailed, 179 subtests passed (20m27s). Both failures are #99's inherited freeze gate, documented below; **everything else is green**.

## ⚠️ Known red: the DSv4 W8A16 freeze gate (inherited, not this PR's)

`tests/test_dsv4_w8a16_export_handoff.py` fails the same 2 tests as #99, naming the
same single file, `prismaquant/model_profiles/base.py`, at the same owed digest
`7cff3a4af253777d831094838a35ec56b9cd5e4c1022654431449163be8a848e`.

**This PR's own commit does not touch the frozen closure at all** — verified: the diff
between #99's tip and this branch's tip contains no file under
`prismaquant/model_profiles/`, and this branch's `base.py` is byte-identical to
#99's. The red is inherited by stacking and **clears the moment #99's
re-freeze lands**; nothing extra is owed here.

---

Extracted from `merge/proven-rescues` commits `7fee52b`, `28b9822`, `f64b861`, `25b82c0`, `2a49363`, `ef4455b`, `db8d3ae`, `41d3333`.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,8 +1,72 @@
 # PrismaQuant Architecture
 
-As of: 2026-08-29 · `rescue/walker-export-gate` — stamps follow, newest
+As of: 2026-08-29 · `rescue/name-projection` — stamps follow, newest
 first, each recording its own branch and date. Re-stamped (2026-08-29,
-`rescue/walker-export-gate`) for
+`rescue/name-projection`) for
+**the cost consumer migrating onto the shared name-projection layer** (§8.8.1):
+`_scan_source_dtype_manifest` lost its private checkpoint→live→recipe builders
+(`_strip_weight_suffix`, `_to_recipe_name`, `_packed_to_recipe_name`,
+`_per_expert_packed_recipe_name`) and now projects every source-kind row
+through `NameProjection.checkpoint_to_live` / `recipe_unit` /
+`packed_parent_of_expert_param` (`prismaquant/allocator_candidates.py`);
+`decision_units._recipe_name` is retired for
+`name_projection.strip_weight_leaf`, as is the inline leaf surgery in
+`production_render_cost.canonical_cost_name` (whose umbrella-infix half stays
+a total normalizer — render/cost payloads key costed MTP rows physically,
+which a declining projection must never drop) and `measure_quant_cost`'s
+act-cache candidate builder. Emitted values are UNCHANGED for existing
+profiles and artifacts — the manifest's profile=None convention now builds
+over the repo's declared generic baseline (`DefaultProfile`, the substitution
+`resolve_cost_target_name` always made) instead of inlined string surgery.
+One deliberate behavior change, fail-closed only: a profile accessor that
+RAISES now propagates `NameProjectionError` instead of silently skipping the
+row (the wo_a shape); MTP rows stay recipe-native verbatim by an explicit,
+commented short-circuit pending a profile declaration of recipe-native
+checkpoint prefixes. Probe, footprint, and read-traffic remain unmigrated.
+Previously re-stamped (2026-08-22, `walker/consumer-probe`) for **the probe
+consumer migrating onto the shared name-projection layer** (§8.8.1): `FisherAccumulator`
+now builds one `NameProjection`, and the probe holds no private name mapping — the
+packed-expert shard-scope filter reads its block id through `NameProjection.block_id`
+(previously a positional `[:3]` qname slice of its own; value-identical on every qname
+shape reachable at that site, pinned by `tests/test_consumer_probe_name_projection.py`),
+and the Fisher skip set keys its embedding clause on the profile's declared
+`ModelProfile.embedding_name()` (previously a hardcoded `"model.embed_tokens"` substring
+test). Probe stats keys, inventories, and shard regexes are untouched, and the regression
+baseline is byte-identical; the walk-edge-list migration itself stays open for all four
+consumers.
+Previously re-stamped (2026-08-22, `walker/consumer-footprint`) for **the
+second R5 consumer migration**: `footprint.py` holds no private name mapping anymore — its
+private `.weight` strips and its packed-expert parser (`packed_expert_alias` + the legacy
+parent fallback, now defined in `name_projection.py` and re-exported) are the shared
+layer's; `source_tensor_bytes_manifest` / `floor_bytes_for_model` accept a prebuilt
+`NameProjection` (keyword-only, mutually exclusive with the raw accessor kwargs), map
+checkpoint keys through `checkpoint_to_live`, keep the profile's DECLARED drops as raw-key
+floor entries (data, not exceptions), and propagate layer refusals instead of swallowing
+them. Emitted bytes are unchanged — projection/accessor parity is pinned test-side.
+Previously re-stamped (2026-08-22, `walker/consumer-readtraffic`) for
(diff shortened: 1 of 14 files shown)


RobTand:

Sequence: lands after 99. Base is now rescue/walker-export-gate; GitHub retargets to main once 99 lands. After 99 merges:
git fetch origin
git checkout rescue/name-projection
git rebase origin/main
If 103 lands first, expect only the ARCHITECTURE.md provenance-block conflict; keep both stamps newest first.

Freeze and verification owed: none of its own. Its 2 failing dsv4 w8a16 handoff tests are the 99 inherited red on base.py at the same digest. Verify with a diff that prismaquant/model_profiles is untouched between the 99 tip and this tip and base.py is byte-identical to 99, then the red clears with the 99 re-freeze. What review must confirm instead: the two silent-wrong-answer fixes stay fail-closed — read-traffic NameProjectionError propagation instead of p-equals-1 reprice, probe grouped scoping via declared accessors not positional slice or hardcoded embed substring — with emitted values pinned byte-identical.

ARCH stamp: same rule, keep both, newest first.

RobTand:
**Merged locally onto `tessera/decouple-gridbook` at `e12a3f4`.** Not pushed.

Two conflicts, both import blocks, and both are the Gridbook retirement surfacing in a branch that predates it:

- `allocator_candidates.py`: the PR imports `trellis_menu`, archived on 2026-09-02 with the rest of the trellis rate surface. This file's only remaining reference to it is a refusal, not a call, so the import does not follow the PR in. `name_projection` does — six uses.
- `read_traffic.py`: the PR imports `CODEBOOK_TENSOR_PREFIX` from `cb_export_config`, which went with the CB lane. The literal had already moved into this file for that reason and keeps its note about where it came from; the PR's `name_projection` imports land beside it.

101 passed across name-projection, read-traffic, the two consumer-projection suites, sensitivity-coverage and the walk export gate.

RobTand:
## Closing: already landed on the integration branch

Checked, not assumed. On `tessera/decouple-gridbook` — the branch every merge has gone to since 2026-08-30 — this PR's files are **byte-identical** (same blob sha):

- `prismaquant/name_projection.py` — IDENTICAL
- `tests/test_name_projection.py` — IDENTICAL
- `prismaquant/model_walk.py` — IDENTICAL

and the contract's distinctive symbols are live: `NameProjectionError` in `name_projection.py`, `footprint.py`, `allocator_candidates.py`; `declared_out_of_graph` in `name_projection.py` and `tests/test_read_traffic.py` — i.e. all four consumers are on the shared layer, which was the condition for this being a refactor rather than a second mapping.

It landed as `6b014b3d` ("refactor(names): one shared name-projection layer, and all four consumers on it") — same subject, different sha, because it was cherry-picked. GitHub cannot close a cherry-picked PR automatically.

`read_traffic.py` still shows as differing: that is later drift on the integration branch, not missing content — `declared_out_of_graph` is present there.

## Why it looked stuck

Base is `rescue/walker-export-gate` → `main`, and **`main` is 187 commits behind `tessera/decouple-gridbook`** (0 ahead; it would fast-forward). Raised with Rob separately.

PR #104, opened 2026-08-30
feat(cluster): unit-sharded production-cache renders, with a byte-faithful merge

RobTand:
> ## 🚧 BLOCKED — draft, do not merge
>
> This PR changes `prismaquant/production_weight_cache.py` (+150/-1), which is inside the **DSv4 W8A16 frozen export-source closure**. It needs Rob's **reviewed** re-freeze. `tests/test_dsv4_w8a16_export_handoff.py` is red on this branch **by design** — that is the gate firing, not a defect in the split.
>
> The change to review:
> ```
> git diff 90cd143 9cd25af -- prismaquant/production_weight_cache.py
> ```
> Three things in it: the `unit_shard` / `render_qnames` parameters, the shard stamp, and the `_LinearActivationCollector` reservoir-draw change (from `5a8303b`).
>
> A reviewed re-freeze to `03bb1aa4c71b59f2cc92c830b561b3911d19120600c18052c95c5c7607ea0894` closes it. **I did not attempt the re-freeze** — blind re-stamping is exactly the band-aid that gate exists to prevent.
>
> Verified on the rebase: the file's digest on this branch is byte-identical to the digest recorded as owed on `merge/proven-rescues`, so rebasing onto current `main` did **not** move the review target.

Split 5 of 6 from `merge/proven-rescues` (#95).

## What it does

Splits a dense production-cache render across boxes. `PRISMAQUANT_UNIT_SHARD=i/N` partitions the render on a deterministic, layer-contiguous, exact-byte-balanced split (`prismaquant/unit_sharding.py`), and `tools/merge_unit_shards.py` reassembles the shards into the same `ProductionWeightCache` under a fail-closed completeness gate. The merged `cache_dir` is byte-identical to an unsharded one.

**Each shard stamps the `(unit, format)` entries it owed**, so the merge gate reads the shard's own debt rather than an operator-supplied layer config. That is the difference between a gate and a hope: an operator-supplied enumeration can be wrong in the same direction as the mistake it is meant to catch.

`tools/cluster_render_sentinel.py` certifies a shard's render before it is merged, refusing a partial render instead of certifying it, and keys its completeness check by the **canonical** format name. It ships here rather than with the other cluster tools (#102) because it imports `prismaquant.unit_sharding` and was born in the same commit.

`production_render_cost`, `production_recache` and packed-expert renders **refuse** the variable rather than pretending to shard.

The merge tool re-derives the render-gate summary over **every** shard's records; inheriting shard 0's counters would label one shard's tally as the whole run's.

## Why it is worth landing

It is the mechanism that makes the second Spark useful for producing rather than only for serving. It also fixes a real reproducibility bug on the way (below).

This is the only PR in the split that touches the render hot path: `production_weight_cache.py`, `build_production_cache.py`, `production_recache.py`, `production_render_cost.py`, `activation_sampling.py`, `streaming_production_cache.py`.

No default changes and `run-pipeline.sh` is untouched. **Unset is a byte-identical no-op**, verified on Qwen3-0.6B.

The substantive behaviour change is inside `_LinearActivationCollector`: row-priority reservoir values are now drawn for **every** hooked Linear rather than only for the ones a run stores. That also fixes resumed builds which did not reproduce a fresh build's sampled rows. (`--include-qnames-file` shrinks the hooked set itself, upstream of this fix, and still changes sampled rows.)

That reservoir change is one of the two pieces the freeze review has to look at, because it is **not** gated behind an unset parameter.

## Dependencies

- **None on the other five.** It can rebase onto `main` alone once unblocked.
- #102 mentions this PR because the third cluster tool lives here.
- One conflict with `main` was resolved: `build_production_cache.py`'s coverage narrowing now applies shard filtering **first**, then `--include-qnames-file`, in the same order the streaming path applies them. This matches the resolution the earlier worker made in `d53bf4c` on the original branch.

## Tests

They **do not reproduce** — #90 and #94 already fixed them — so every failure below is attributable.

This branch: **2 failed, 5546 passed**, 143 skipped, 3 xfailed, 179 subtests passed (19m22s). Both failures are the declared blocker at the top of this PR — the freeze gate naming exactly `prismaquant/production_weight_cache.py` at exactly `03bb1aa4c71b59f2cc92c830b561b3911d19120600c18052c95c5c7607ea0894`, the digest recorded as owed. **Everything else is green.**

New coverage: `tests/test_unit_sharding.py` (825 lines).

---

Extracted from `merge/proven-rescues` commits `5a8303b`, `eb06a72`, `6a66cf7`, `fee7a7f`.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,24 @@
 # PrismaQuant Architecture
 
-As of: 2026-08-29 · `docs/architecture-currency-20260829` — stamps follow, newest
+As of: 2026-08-29 · `rescue/parallel-producer` — stamps follow, newest
 first, each recording its own branch and date. Re-stamped (2026-08-29,
+`rescue/parallel-producer`) for
+**unit-sharded production-cache renders** (§5.4, D33): `PRISMAQUANT_UNIT_SHARD=i/N` splits a
+dense render across boxes on a deterministic layer-contiguous, exact-byte-balanced partition
+(`prismaquant/unit_sharding.py`), and `tools/merge_unit_shards.py` reassembles the shards into
+the same `ProductionWeightCache` under a fail-closed completeness gate. No default changes and
+`run-pipeline.sh` is untouched; unset is a byte-identical no-op, verified on Qwen3-0.6B. The
+substantive behaviour change is inside `_LinearActivationCollector`: row-priority reservoir
+values are now drawn for every hooked Linear rather than only for the ones a run stores, which
+also fixes resumed builds that did not reproduce a fresh build's sampled rows.
+(`--include-qnames-file` shrinks the hooked set itself, upstream of this fix, and still
+changes sampled rows — D33.) Each shard stamps the `(unit, format)` entries it owed, so the
+merge gate reads the shard's own debt instead of an operator-supplied layer config, and the
+merged `cache_dir` is byte-identical to an unsharded one. `production_render_cost`,
+`production_recache`, and packed-expert renders refuse the variable rather than pretending to
+shard.
+
+Earlier stamp (2026-08-29,
 `docs/architecture-currency-20260829`) for a **§8.4 conformance-matrix
 correction on `glm5_next`**, found by a principle-13 sweep of `origin/main`
 rather than by a test — the two doc gates were green across both drifts. (i) The
@@ -19,8 +36,9 @@ quantized formats outside the three module families PR #53906 wires a
 `b4a8846` refresh, so it is the § P13 "currency is not truth" case: this file
 was re-stamped in lockstep while carrying a false statement about a serving
 default. Nothing else in the window crossed the trigger; the reasoning is in the
-commit message. Re-stamped (2026-08-29,
-`claude/trellis-continuous-surface`) for the **trellis seam correction**
+commit message.
+
+Earlier stamp (2026-08-29, `claude/trellis-continuous-surface`) for the **trellis seam correction**
 (§4.9): the seam that landed earlier the same day is now **fail-closed** when
 enabled, and this document's claim that it made trellis rungs "pass the same
 legality, aggregation and byte accounting every other candidate does" is
@@ -3772,6 +3790,51 @@ at phase 50 with `before=("gptq",)` and no relation to each other, and
 (matching `pipeline.py`'s own stage list); `fisher_gptq` (50, archived); `scale_sweep` (60,
 after gptq). The production lever set is §3.3.
 
+**A unit shard is a lifetime mode of that same store too** (`prismaquant/unit_sharding.py`,
+2026-08-23). `PRISMAQUANT_UNIT_SHARD=i/N` makes `build_production_cache` render only shard `i`
+of a deterministic layer-contiguous partition of the units it already enumerates, balanced by
+exact source-tensor bytes with a min-max partition DP. Unset is a byte-identical no-op, and
+`run-pipeline.sh` does not set it. `tools/merge_unit_shards.py` puts the shards back into one
+`ProductionWeightCache` with the same layout — not a second store — and refuses unless every
+unit of the recomputed partition appears exactly once, under its owning shard. There is no
+force flag; missing, duplicate, and out-of-shard units all name the units and stop.
+
+The gate reads what each shard says it owed, not what an operator says it should have. Every
+shard stamps `owed_pairs` — the exact `(unit, format)` entries fixed at the moment its render
+map was built, from the same structure the render loop consumes — so a `--render-layer-config`
+that disagreed with the config a shard actually ran under can no longer under-expect a dropped
+unit into a clean merge. That flag survives only for stamps predating the field. The merged
+artifact is the unsharded artifact: `.pt` files, `render_scores.json` (through the stage's own
+`_write_render_score_sidecar`, whose `{"records": …}` envelope the resume loader requires) and
+`activation_max_abs.json` are written in the same order and shape, the render-gate summary is
+re-derived over every shard's records rather than inherited from shard 0, and the merged pickle
+differs from an unsharded one only by its `cache_dir` path and the `unit_shard_merge` block.
+
+Three properties make a shard's rendered bytes the unsharded run's bytes, and each is enforced
+rather than assumed. Layer atomicity keeps every fused-sibling group in one shard, so the joint
+NVFP4 global (`_compute_nvfp4_joint_global`, a max over the members present in the run) is
+computed over the same members; the stage asserts the grouping against the model profile and
+refuses a straddling group. The hooked activation set, the fused-sibling joint globals, and
+`activation_max_abs` are all computed over the FULL enumeration — only the render narrows. And
+`_LinearActivationCollector` now draws its row-priority reservoir values for **every** hooked
+Linear, not only for the ones this run stores (`production_weight_cache.py`). One generator
+feeds every Linear's reservoir, so drawing only for stored Linears made the surviving rows a
+function of which other Linears happened to be in the run — which also means a **resumed** build
+did not reproduce a fresh build's rows. That is a defect fix, not only shard plumbing; a fresh
+full build is unaffected byte-for-byte. `--include-qnames-file` is a different and still-open
+case: it filters the `qnames` list before the fill, so it shrinks the hooked set itself and the
+fix does not reach it (§12, D33).
+
+`production_render_cost` renders nothing, so it does not shard: it runs once on the merged
+cache and refuses a cache still carrying an unmerged shard stamp. `production_recache` and
+`--recache-layer-config` refuse the variable outright — the recache is a whole-model
+calibration replay, not a per-unit render. Packed-MoE expert renders refuse it too: their
+reservoir has the same shared-generator coupling and their enumeration is separate, so v1
+shards dense units only. `tools/cluster_render_sentinel.py` is the cross-box preflight: it
+renders K deterministically chosen units through this same fill path and byte-compares the
+manifests, refusing a match that was produced without `PRISMAQUANT_DETERMINISTIC=1`. Cross-box
+execution and the 4B repeat of the bit-identity acceptance remain open (§12).
+
 ### 5.5 Named invariants
 
 | Name | One line | Detail |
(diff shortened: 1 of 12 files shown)


RobTand:

Stays DRAFT until Rob completes the freeze review below. Author supplies the owed digest and refuses to self-re-stamp — do not merge, do not re-stamp without reviewed approval.

What the Rob freeze review must cover, item by item:
```
git diff 90cd143 9cd25af -- prismaquant/production_weight_cache.py
```
(1) unit_shard / render_qnames parameters — confirm unset is a byte-identical no-op and the sharded path partitions deterministically; (2) shard stamp on (unit, format) entries — confirm the merge gate reads shard-owed debt, not operator config; (3) _LinearActivationCollector reservoir change from 5a8303b — row-priority draws now for EVERY hooked Linear, not only stored ones; this is UNGATED and also changes resumed-build sampled rows, so it needs correctness review on the hot path plus build_production_cache.py, production_recache.py, production_render_cost.py, activation_sampling.py, streaming_production_cache.py.

Close-out: reviewed re-freeze to 03bb1aa4c71b59f2cc92c830b561b3911d19120600c18052c95c5c7607ea0894 turns tests/test_dsv4_w8a16_export_handoff.py green (currently red BY DESIGN as the gate firing). Until that reviewed stamp lands, keep draft and leave the CONFLICTING state untouched.

RobTand:
## Verdict: PARTLY LIVE — closing unmerged, one hunk salvaged and landed, one finding filed

**The sharding mechanism is superseded, 1:1.** `origin/main` already has unit-sharded production-cache rendering, built, documented and tested: `production_cache_stripes.plan_stripes()` (layer-atomic, asserts fused siblings and packed experts cannot cross a boundary), `build_production_cache.py --include-qnames-file`, `union_production_cache` (`manifest`/`verify-shard`/`union`/`verify`, fail-closed), and `cluster_campaign` for dispatch. Every piece of this PR maps onto one of them: `unit_sharding.partition_units`↔`plan_stripes`, `PRISMAQUANT_UNIT_SHARD`↔`--include-qnames-file`, `merge_unit_shards`↔`union`/`verify`, `cluster_render_sentinel`↔`verify-shard`. A second cache mechanism is a principle-8 defect by construction, so this half cannot merge on its merits regardless of quality.

**One hunk is live, real, and unrelated to sharding — landed at `56c765d`.** `_LinearActivationCollector` hooks the full quantizable enumeration but stores only `qnames_needing_activation`, which is resume-dependent. One seeded generator feeds every Linear's priority reservoir, and the draw sat inside the store branch — so the slice of the random stream a Linear received was a function of which *other* Linears the run happened to store. **A `--resume` build kept different calibration rows than a fresh build**; rows → Hessian → rendered bytes.

I verified this independently rather than accepting it: reverting the fix on the current tree gives `AssertionError: layers.2.o_proj sampled different rows when only a subset was stored`, and restoring it passes, with 80 passed / 2 skipped across the four production-cache suites. The API grew a `new_priorities` parameter with exactly one of `generator`/`new_priorities` required, so no caller can silently fall back.

The freeze digest for `production_weight_cache.py` moves `69e29daf…` → `b7eaa4c0…`. **Not re-stamped** — that is release-keyed and Rob's review. A full-store single pass digests identically pre- and post-fix, so fresh builds are byte-preserved; only `--resume` artifacts moved, which is exactly the class of change the closure exists to gate.

**A finding this PR uncovered that outlives it,** filed separately: the same hook-full/store-narrow asymmetry is a gap in the *existing* stripes path, and the salvaged fix does **not** close it — `plan_stripes` is LPT-binned, so stripes are non-prefix subsets and 4/4 units diverge including stripe 0, identically before and after. Resident path only; streaming reads pre-sampled rows through `ActivationIndex` and has no reservoir. No test asserts byte-equality either way.

**One correction to the record while closing:** the 1.87x two-box figure was never a production-cache render. Those campaign stages are `plan-*` → `materialize-part-*` → `merge-checkpoint-parts`, splitting by safetensors shard index — PrismaSnap source materialization, zero production-cache references. The stripes/union path is built and tested and has **never been run end to end**.

RobTand:
> **Maintainer note (I am the final word here, and I verified the load-bearing claim myself).**
> The salvage's item 1 — the reservoir fix — **is already landed**, as `56c765d` on
> `tessera/decouple-gridbook`, and I diffed it: `activation_sampling.py` and
> `tests/test_activation_reservoir_row_identity.py` are md5-identical to the rescue branch's,
> and `production_weight_cache.py` differs only by the Tessera render hook that exists on that
> branch and not on `origin/main`. So the red-to-green receipt below stands and its subject is
> on a branch, not a proposal. That branch has not been pushed — Robert's call — so this PR
> stays open until it lands.
> I confirmed the mechanism at the source rather than from the report: the draw now happens
> before the `if key not in self.store_qnames: return` guard, for every hooked Linear.
> Item 2 (the striped-vs-unsharded hook-scope gap) I am filing as its own issue against
> `production_cache_stripes` / `union_production_cache`, because it indicts the mechanism
> `origin/main` already ships and has nothing to do with whether this PR lands.
> Item 3 (the exact-bytes contiguous DP as a balance improvement to `plan_stripes`) I am
> recording in that issue and not scheduling.

---

## Verdict: PARTLY LIVE — two hunks in two of twelve files survive; the sharding mechanism is SUPERSEDED

*(+2715/−16 across 12 files reduces to +142/−8 across 3, including a new test.)*

I rebuilt this PR's question from the tree rather than from its title. The short version: **`origin/main` already has unit-sharded, fail-closed, multi-host production-cache rendering.** `unit_sharding.py`, `merge_unit_shards.py` and `cluster_render_sentinel.py` are a second implementation of it, and under principle 8 a second implementation of the one cache mechanism is a defect by construction, not a feature.

But this PR is not empty. Buried under the sharding is **a live reproducibility bug on the shipping `--resume` path**, and a design finding that indicts the mechanism `origin/main` already ships. Those are the salvage.

---

### (a) What `origin/main` already has

The split/merge is built, documented, and tested — read at commit `55d43e3`:

| Concern | Module / entry point |
|---|---|
| Deterministic partition | `prismaquant/production_cache_stripes.py` — `plan_stripes()`; keeps each decoder layer whole, makes `mtp.*`/`lm_head` indivisible auxiliary groups, and **asserts the profile's fused-sibling and packed-expert serving units cannot cross a group boundary** (`production_cache_stripes.py:96-118`). LPT-balanced over `n_params × max(in_features,1)`. Schema `prismaquant.production_cache_stripe_plan.v1`. |
| Worker selection | `build_production_cache.py:651` `--include-qnames-file`, fed by the plan's SHA-bound `stripe-NN.qnames.txt` |
| Shard certification | `union_production_cache.py:1777` `verify-shard` |
| Fail-closed merge | `union_production_cache.py:1783` `union` (1837 lines) — refuses overlaps, failed renders, identity drift, missing/modified/symlinked files, incomplete coverage, unmergeable metadata |
| Post-transfer re-verify | `union_production_cache.py:1788` `verify` |
| Multi-host dispatch | `prismaquant/cluster_campaign.py`, `tools/run_cluster_campaign.py` |
| Docs | `docs/ARCHITECTURE.md:1727` (the "4/4 D×N" row) and `:3817-3860` |
| Tests | `tests/test_production_cache_stripes.py`, `tests/test_union_production_cache.py`, `tests/test_packed_expert_union_identity.py` |

### (b) What this PR adds over that

**Duplicated, 1:1 (SUPERSEDED):**

| This PR | Already on `origin/main` |
|---|---|
| `unit_sharding.partition_units` — layer-atomic, deterministic, refuses to split fused/packed groups | `production_cache_stripes.plan_stripes` — same three properties, `:96-118` |
| `PRISMAQUANT_UNIT_SHARD=i/N` | `--include-qnames-file` + `stripe-NN.qnames.txt` |
| `merge_unit_shards.py merge` (fail-closed completeness) | `union_production_cache union` |
| `merge_unit_shards.py digest` | `union_production_cache verify` + manifest SHA-binding |
| `cluster_render_sentinel.py` | `union_production_cache verify-shard` |
| `unit_shard` refuses a CB render scope | union "deliberately refuses all CB keys" (`ARCHITECTURE.md:3853`) |

The one honest improvement in that column is the balance objective — exact `numel × element_size` under a contiguous-partition DP, versus LPT over a shape proxy. That is worth a patch to `plan_stripes`, not a parallel mechanism.

**Genuinely novel (LIVE):**

1. **The reservoir fix** (`activation_sampling.py` whole diff + `production_weight_cache.py:964-1011`). This is not about sharding at all.

   `_LinearActivationCollector` hooks the full quantizable enumeration but stores activations only for `qnames_needing_activation` (`production_weight_cache.py:5145-5146`), which is **resume-dependent**. One seeded `torch.Generator` feeds every Linear's priority reservoir, and the draw happened only inside the `store_qnames` branch (`:970`, `:990`). So the slice of the random stream a Linear received was a function of which *other* Linears the run happened to store — **a `--resume` build kept different calibration rows than a fresh build, the GPTQ Hessian followed the rows, and the rendered bytes followed the Hessian.**

   Not a reasoned claim — reproduced. See "Measurement" below.

2. **The hook-scope/render-scope contract.** `fill_production_weight_cache(unit_shard=..., render_qnames=...)` narrows only what gets *rendered*, holding the hooked activation set, the fused-sibling NVFP4 joint globals and `activation_max_abs` at the full enumeration. That is what makes a shard's bytes the unsharded run's bytes.

   **`origin/main`'s striping does not have this property on the resident path**, and the PR author says so in his own comment. `--include-qnames-file` narrows the **hooked set itself**: `build_production_cache.py:843` — `qnames = [q for q in qnames if q in allowed]` — before `fill_production_weight_cache` at `:883` hooks exactly that set (`:5145`).

   Measured, not reasoned. Because `plan_stripes` is **LPT-binned, not contiguous** (it sorts groups by work descending and assigns each to the least-loaded bin), a stripe over roughly equal layers is a *non-prefix* subset — layers `{0,2,…}`. Hooking that subset against the full model:

   ```
   full hooked  = 8 Linears
   stripe hooked= 4 Linears  ['layers.0.q_proj','layers.0.o_proj','layers.2.q_proj','layers.2.o_proj']
     layers.0.q_proj          rows match unsharded: False
     layers.0.o_proj          rows match unsharded: False
     layers.2.q_proj          rows match unsharded: False
     layers.2.o_proj          rows match unsharded: False
   RESULT: 4/4 units diverge -> GAP OPEN
   ```

   **Every stripe diverges, including stripe 0** — the generator advances at a different rate between batches once the hooked population changes, so even the first Linear's second batch draws differently. Identical output on this branch and on unmodified `origin/main`: **the salvaged fix does not close this**, which is why it is a separate issue and not a bigger patch here.

   **Scope it honestly — this is the resident path only.** The streaming path has no reservoir of its own: `streaming_production_cache.py` reads pre-sampled rows through `ActivationIndex` (`:43,134,1786`), keyed per qname, and its only `priority` symbols are layer-residency (`_priority_setter`, `:1712`). Narrowing at `:123` there removes modules from the render, not rows from a shared stream. So a *streamed* striped build is unaffected; a *resident* one is not.

   **And nothing tests either way**: grepping `tests/test_union_production_cache.py`, `tests/test_production_cache_stripes.py` and `tests/test_packed_expert_union_identity.py` for `unsharded` / `identical` / `byte_identical` returns nothing, and `ARCHITECTURE.md:3817-3860` claims coverage and identity binding but never byte-equality with an unsharded run. The 6/6 cross-box identity result does not cover it — that used synthetic activations on three tensors of one layer and never exercised the hook-scope path.

   This is a real finding against the *existing* mechanism. It belongs there, not in a new one.

### (c) PrismaBuild

`/home/rob/prismabuild` owns the dispatch layer and nothing else relevant here: `pool.py:340` `PoolQueue` (claim-by-`rename`, NFSv4), `pool.py:194-337` `ResourceLedger` (admission, with the starvation floor at `:105`), `pool.py:672-707` retries, `core.py:1327` `seal_action` / `core.py:2268` CAS provenance. It has **zero** occurrences of `unit_shard`, `production_cache_stripes`, `union_production_cache`, `include_qnames`, or `cluster_campaign` in code; "shard" there means a source safetensors file index (`dispatch_tessera_shards.py:75,94-95`), and "stripe" means lock striping (`slurm.py:1283`). `render` appears only in `tools/fleet/render_identity.py`, a measurement script.

So PrismaBuild has not taken over any part of this PR — but it is also not where the salvage goes. It schedules; it does not decompose a render. Its own `docs/design.md:782-787` files per-tensor sharding as future work.

### One correction to the framing

The **1.87x on a 27B was not a production-cache render.** Campaign `qwen38-27b-prismasnap-20gb-20260825` has stages `plan-{host}` then `merge-plans` then `materialize-part-{host}` then `merge-checkpoint-parts` (`tools/build_prismasnap_27b_campaign.py:766,828,871,917`), and `materialize-part` splits by **safetensors shard index** (`--shards-file`, `:884`) through `tools/prismasnap.py:186,326` — that is PrismaSnap source materialization. Neither the stripes/union path nor this PR has a measured multi-box production-cache render. That does not rescue this PR (`plan_stripes` + `union` is still the built mechanism and this is still a duplicate of it), but "already exists and is measured" should be read as "already exists, and is tested, and has never been run end to end."

---

### Measurement

Salvaged the PR's own CPU-only regression test out of `tests/test_unit_sharding.py:786` and ran it against **unmodified `origin/main`** (`55d43e3`):

```
CUDA_VISIBLE_DEVICES="" PYTHONPATH=. python -m pytest tests/test_activation_reservoir_row_identity.py -q
...
E  AssertionError: layers.2.o_proj sampled different rows when only a subset was stored
1 failed, 15 warnings in 3.16s
```

With the two salvaged hunks applied:

```
1 passed, 15 warnings in 3.05s
```

Targeted suite on the salvage branch:

```
CUDA_VISIBLE_DEVICES="" PYTHONPATH=. python -m pytest \
  tests/test_activation_reservoir_row_identity.py tests/test_production_weight_cache.py \
  tests/test_streaming_production_cache.py tests/test_production_cache_stripes.py \
  tests/test_union_production_cache.py tests/test_packed_expert_union_identity.py \
  tests/test_production_render_cost.py tests/test_mtp_production_cache.py \
  tests/test_cost_activation_row_usage.py tests/test_activation_fair_pricing.py \
  tests/test_dsv4_w8a16_export_handoff.py -q

2 failed, 134 passed, 2 skipped
```

The two failures are the DSv4 W8A16 freeze gate firing, by design.

Whole suite on the salvage branch, same environment:

```
2 failed, 5594 passed, 144 skipped, 3 xfailed, 179 subtests passed in 1013.61s (0:16:53)

FAILED tests/test_dsv4_w8a16_export_handoff.py::test_exact_w8a16_handoff_returns_read_only_receipt
FAILED tests/test_dsv4_w8a16_export_handoff.py::test_tracked_frozen_export_sources_match_reviewed_bytes
```

Same two, and only those two. Nothing else in the tree notices this change.

The second caller of the changed helper, the packed-MoE collector at `production_weight_cache.py:5861`, is deliberately left on `generator=`: it hooks and stores the same set (`experts_qnames`, built at `:6438` from the render scope and passed whole at `:6735`), so it has no draw/store asymmetry. It *would* break under any unit-shard narrowing of `experts_qnames` — which is exactly why this PR refuses to shard packed-expert renders.

### Conflicts

`git merge-tree --write-tree origin/main 7889abc` gives **exactly one conflicting path**:

```
Auto-merging docs/ARCHITECTURE.md
CONFLICT (content): Merge conflict in docs/ARCHITECTURE.md
Auto-merging prismaquant/production_recache.py
Auto-merging prismaquant/production_weight_cache.py
```

That is the provenance-block stamp the PR body predicted — every code file auto-merges. So the `CONFLICTING` label on this PR is a docs-header collision, not a code disagreement, and **the PR would rebase clean**. That is not an argument for landing it; it just means the duplication is the only real objection, and a clean rebase does not answer it.

I resolved no conflicts, because I did not rebase. The one conflicting file (`docs/ARCHITECTURE.md`) and the one file the PR body flagged as needing a merits resolution (`build_production_cache.py` — shard filtering before vs after `--include-qnames-file`) are both **files I dropped entirely**, so both conflicts are mooted rather than resolved. The two hunks I kept touch `activation_sampling.py` and `production_weight_cache.py`, and both apply to `55d43e3` unchanged from the PR's merge base at `8461ae2` (+29/−7 either way on `activation_sampling.py`).

### Freeze digest — not re-stamped

The reduced change moves `prismaquant/production_weight_cache.py` from `69e29daf21ec8fcfb6ee86375dc6a9792750f38e32337c931b26be64bb0c9fc7` to **`b7eaa4c0992b30084ed74f446068460218804f1c43fd23527204ce9fdca29ef2`**. That is **not** the `03bb1aa4…` this PR owed — dropping the `unit_shard`/`render_qnames` parameters moves the review target. I have not re-stamped it.

**The blast radius is narrower than the PR implies, and it is measured.** Pre-fix, a full-store run draws once per hook inside `update_priority_reservoir`; post-fix it draws once per hook just before the store check, same count, same order, same guard (`n_rows > 0 and self.max_rows > 0` mirrors the helper's own early return). So a **fresh single-pass build is byte-preserved by construction** — and by receipt: the same full-store collection digests to `6d1cd869eb308e8713ab165a7a1acd7961010b070ec6d37372026da7c35e8324` on unmodified `origin/main` and on this branch.

Only `--resume` builds moved. Any artifact already built under `--resume` on the old code has bytes the fixed code will not reproduce; artifacts built in one pass are unaffected. That is the scope the freeze review needs.

### On the Muse Spark comment above

No patch, no diff — it restates this PR's own three review items and its own owed digest. It does not touch the duplication question at all, and nothing in it would have surfaced `production_cache_stripes.py` or `union_production_cache.py`. Its item (2) — "confirm the merge gate reads shard-owed debt, not operator config" — is accurate about the PR but reviews a mechanism that should not exist; `union_production_cache` already derives coverage from a SHA-verified stripe plan or the assignment (`ARCHITECTURE.md:3838-3845`). Its item (3) correctly identifies the reservoir change as ungated and the piece needing real scrutiny; that is the one item worth keeping.

### Recommended disposition

**Close this PR.** Land instead:

1. **The reservoir fix**, as a standalone ~50-line commit — prepared, red-to-green demonstrated, digest computed. It is a bug fix on the shipping path and is worth landing on its own merits whatever happens to sharding.
2. **An issue against `production_cache_stripes` / `union_production_cache`**: on the **resident** path `--include-qnames-file` narrows the hooked activation set (`build_production_cache.py:843`), so a striped render is not byte-identical to an unsharded one — measured, 4/4 units, every stripe — and no test asserts it either way. The fix is this PR's contract — narrow `render_formats_by_qname`, hold the hooked enumeration full — applied to the existing mechanism, plus the missing `union(stripes) == unsharded` test. Streaming is unaffected and should be said so explicitly in `ARCHITECTURE.md:3817-3860`.
3. **Optionally**, port the exact-bytes contiguous-partition DP into `plan_stripes` as a balance improvement.

Also worth correcting: the memory note `two-box-sharding-already-exists` recommends the stripes path as-is under "How to apply", and its 1.87x figure is PrismaSnap, not a cache render.

Nothing here is pushed, merged, closed or edited — this is a report.

PR #107, opened 2026-08-30
docs: survey GB10 GPU oversubscription prior art

RobTand:
## Summary

- surveys GPU sharing, Linux memory-pressure control, DL preemption research, and PyTorch allocator controls against the DGX Spark unified-memory constraints
- gives one direct recommendation and adoption cost for each requested A-D decision
- records a bounded GB10 conformance experiment showing that cgroup memory.high and memory.max did not account a touched 2 GiB CUDA tensor
- recommends combined global PSI and MemAvailable trajectory signals, progress-aware victims, and stateful bounded speculation

## Key conclusions

1. Keep proactive queue-directed kill/requeue; memory.high is host-side containment on the measured stack, not a CUDA unified-pool throttle.
2. Add global PSI beside the MemAvailable trajectory: PSI is the better harmful-reclaim signal, while trajectory remains the fast-allocation guard.
3. Split soft and hard reserves, allow one protected speculative overlap, and persist failed resident-generation combinations to prevent livelock.

## Validation

- 20 documentation tests passed: tests/test_docs_staleness.py and tests/test_architecture_doc.py
- all cited external links were checked; the USENIX Pollux page was browser-verified because it rejects plain curl
- documentation-only change; tools/pqwork.py is unchanged

This is intentionally stacked on claude/pqwork-memory-caps so the PR contains only the survey.

### DIFF
--- a/docs/design/gpu_oversubscription_prior_art.md
+++ b/docs/design/gpu_oversubscription_prior_art.md
@@ -1,0 +1,702 @@
+# GPU oversubscription prior art for PrismaQuant on DGX Spark
+
+**Status:** research survey and design recommendation, 2026-08-30.  This
+document critiques the governor in [`tools/pqwork.py`](../../tools/pqwork.py)
+as documented by [`work_queue.md`](work_queue.md); it does not change that
+implementation.
+
+## Executive decision
+
+The useful prior art is not a replacement cluster scheduler.  It is three
+policies to incorporate into the existing receipt-gated queue: Linux PSI as a
+second system signal, progress-aware preemption, and a stateful speculation
+guard.  MIG is unavailable on this GB10, generic GPU-sharing stacks do not
+isolate the one physical memory pool, and cgroup memory limits do not account
+the CUDA portion of that pool on the measured software stack.
+
+The four requested decisions are:
+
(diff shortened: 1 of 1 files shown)


RobTand:
Closing in favour of #111 (v2), which supersedes this on every axis I could measure: it adds Volcano vGPU, Kueue all-or-nothing, HAMi isolation, dynamic fractions and memory swap as sources, and it does carry the gang-eviction phases this PR had — my first diff read those as missing only because v2 hyphenates `gang-eviction`.

One point here was genuinely absent from v2 and has been grafted onto the base branch rather than dropped: KAI's shard-level **minimum runtime before preemption/reclaim** (`docs/operator/scheduling-shards.md`). It earns its place because it is the same knob as `pqwork`'s `--evict-horizon-s`, so the survey now cites prior art for a parameter we already ship.

The four URLs unique to this PR were otherwise version-pinned variants of pages v2 cites at `latest`/`master` (`gpu-operator/22.9.2`, `ray/en/latest`) plus `resource-optimization/fractions`, which v2 replaces with the more specific `dynamic-fractions`. Worth noting for whoever maintains this doc: the pinned forms are the more durable citations, since `latest` will drift out from under the claims made about it.

Issue #175, opened 2026-09-04
CI cannot install the now-required private Tessera package


RobTand:
## Evidence

The main CI workflow installs PrismaQuant but never installs Tessera, while `prismaquant/tessera_formats.py` imports Tessera's grammar unconditionally. CI run https://github.com/RobTand/prismaquant/actions/runs/33881032397 fails import and aborts suite collection with 26 `ModuleNotFoundError: No module named 'tessera'` errors; its parent does the same in https://github.com/RobTand/prismaquant/actions/runs/33801860048. This is P1: a gate that cannot reach the code it should check.

## Agreed resolution

Rob clarified that Tessera will become public after its remaining bug fixes and audit. No cross-repository credential is requested or needed for that plan, and repository visibility is not changed by this work.

PR #176 installs the exact `TESSERA_DEV_PIN_COMMIT`, resolved from its owning source file without importing PrismaQuant, before PrismaQuant in both CPU jobs. A malformed pin stops the step before any checkout ref is emitted. The workflow uses ordinary checkout credentials; failure to access Tessera while it is still private is expected. Once Tessera becomes public, the same exact checkout works for normal and fork PRs without a private-repository secret. Serving-release PENDING sentinels stay unchanged until the release is approved under RobTand/tessera#17.

## Completion

Merge the reviewed workflow and verify both CI jobs after the planned Tessera publication. Do not close this issue on local/static checks alone; successful dependency installation on GitHub is the remaining receipt.

## Why filed rather than fixed originally

The original workflow fix was separable from the then-unknown private dependency distribution policy. Rob has now supplied that policy: publish Tessera after fixes and audit. The former token decision is superseded and `needs-decision` no longer applies.

RobTand:
Implementation is published as PR #176 from `codex/pq-175-ci-tessera`, commit `6a68dfc3b07d9b52e570b85d5affbbdde68fa55b`.

PrismaBuild evidence (dl380g10 CPU population; `CUDA_VISIBLE_DEVICES=''`; these are static CI-contract/doc tests, not a CUDA acceptance run):

- Pre-fix `469aa632bf3359c814255b9e0539f2845e8efb96f5b13feaca03c831b78c2b33`: the two new #175 regressions failed (`2 failed`): no stdlib pin resolver and no exact private-Tessera install contract.
[...]

PR #176, opened 2026-09-04
Install the exact Tessera development pin in CI

RobTand:
Refs #175.

CI currently fails before collection because PrismaQuant imports Tessera's grammar but never installs Tessera. Both jobs now resolve the single `TESSERA_DEV_PIN_COMMIT` literal with a stdlib-only reader, check out that exact commit without persisting credentials, and install Tessera before PrismaQuant. Invalid pins stop the shell step before it can publish an empty ref and fall back to Tessera's default branch.

Tessera will become public after its remaining fixes and audit, as Rob has specified. This workflow uses ordinary checkout access and requires no private-repository secret, including for fork PRs. Checkout failure while Tessera remains private is expected; functional GitHub CI recovery is verified after publication. No repository visibility, serving eligibility, or PENDING release pin changes.

Validation through PrismaBuild on dl380g10, 1 CPU / 4 GiB, `CUDA_VISIBLE_DEVICES=''`:

- Original implementation: two regression failures before the fix (`469aa632bf3359c814255b9e0539f2845e8efb96f5b13feaca03c831b78c2b33`); both passed afterward.
- Shell-failure regression: `b6528206bd859a73121c59e27e7ccf8b5815e40a3c8ee03119f8ae0d497d1a97` produced `2 failed, 4 passed`, with `AssertionError: CI masked a failed pin resolver` in both jobs.
- Post-fix `ee2688b19208a673fe3c8cbb82930fe2d9fc08dd800ce890533d9f76eedeb6eb`: all six CI-contract cases passed; combined touched CI/doc files produced `22 passed, 3 failed`, no skips. Those three failures are the already-recorded missing `compressed_tensors` dependency, also present in the pristine-main comparison `742e6ff00bd859a58f422793c17a3c47b3a80b2db01b3cb31d36114e219f3c8f`.
- Publication-contract regression `cffe356189277b0d1a5c675aa815d2aad9870889f91470da735acda5b9402f5f`: `1 failed, 5 passed` because the former workflow still required `TESSERA_REPO_TOKEN`. Final `1f0f75fdaa8a21d449f51be67dfe43688f6aded9d3f892766dcfc91797473616`: all six CI cases passed, combined `22 passed, 3 failed`, no skips, with the same known dependency failures.

The latest CPU snapshots omit the three tracked external calibration symlinks because PrismaBuild correctly refuses to seal paths outside the repository. None of these static CI/doc tests reads those datasets; the source worktree's links are restored and unchanged. This is no CUDA acceptance claim and no full-suite run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,33 @@ jobs:
           python-version: ${{ matrix.python }}
           cache: pip
 
+      # Resolve the compatibility pin without importing prismaquant: its
+      # package surface is exactly what this job has not installed yet.
+      - name: Resolve Tessera development pin
+        id: tessera-pin
+        run: echo "commit=$(python tools/resolve_tessera_dev_pin.py)" >> "$GITHUB_OUTPUT"
+
+      - name: Require Tessera repository token
+        env:
+          TESSERA_REPO_TOKEN: ${{ secrets.TESSERA_REPO_TOKEN }}
+        run: |
+          if [ -z "${TESSERA_REPO_TOKEN:-}" ]; then
+            echo "::error::TESSERA_REPO_TOKEN is required to check out pinned private Tessera"
+            exit 1
+          fi
+
+      - name: Check out pinned private Tessera
+        uses: actions/checkout@v7.0.1
+        with:
+          repository: RobTand/tessera
+          ref: ${{ steps.tessera-pin.outputs.commit }}
+          token: ${{ secrets.TESSERA_REPO_TOKEN }}
+          persist-credentials: false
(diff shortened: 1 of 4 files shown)


RobTand:
Coordinator source review complete. The resolver reads the single authoritative literal and preserves failure before publishing a checkout ref; both jobs install that exact dependency before importing PrismaQuant. Six focused CI-contract tests are green through PrismaBuild; the three documented dependency-environment failures are not being claimed green. Hosted pin resolution succeeds and private checkout currently fails as expected. Merging preparation only: #175 remains open until the planned public release makes both hosted jobs pass. No credential or visibility changes are required now.

Issue #87, opened 2026-08-25
shipgate: KL/PPL + greedy-smoke certify boundary-broken artifacts — three DSV4-Flash quants within ~3% PPL span a 6× behavioral gap; proposal: add a sampled behavioral component


smb209:
## Summary

The current ship gates (KL/PPL + greedy-smoke) are argmax- and distribution-distance measures, and there is a real defect class that is structurally invisible to both: boundary-token distribution defects that only manifest under sampling. We hit one on DeepSeek-V4-Flash-0731, built a reproducible instrument that detects it, and — the part that makes this an upstream issue rather than a local anecdote — verified that the instrument separates three independently built artifacts whose PPLs are indistinguishable. The proposal at the bottom is a small, cheap behavioral gate alongside the existing ones; we're offering our battery as a starting point.

Filed in the same spirit as gridbook #46/#47 and prismaquant PR #86 — evidence first, reproduction pointers at the end.

## The defect class

Quantized DSV4-Flash artifacts stutter or fail to emit a clean `</think>` on ultra-short numeric prompts ("144÷12", "9²") under sampling. The *answer stays correct* — which is exactly why greedy-smoke and KL/PPL miss it: greedy takes the argmax path where the boundary token still wins, and KL/PPL average over a corpus where a per-token near-tie at one boundary position is noise. But under a token cap the model never reaches its answer (it presents as a hang), and on harder prompts the same defect surfaces as zero-`</think>` runaways. It lives in the boundary-token distribution under sampling, on specific prompt shapes. No gate that scores argmax agreement or distributional distance can see it, at any threshold.

## The evidence: three artifacts, one PPL, a 6× behavioral spread

All three are DSV4-Flash-0731 quants, re-served on the **same stack** (gridbook 0.8.11, single GB10, production config), same frozen scorer, battery pre-registered before the data existed.

| artifact | bpp | calibration | WikiText-2 PPL ↓ | battery pooled ↓ (of 180) | terse stratum (of 60) |
|---|---|---|---|---|---|
| official (unquantized) | — | — | — | 10 | 0 |
| **v4c** (ours) | 2.17 | numeric triplets | **12.47** | **14** | 0 |
| **v2-ayot** (ours) | 2.166 | AYOT triplets | 12.71 | 16 | 1 |
| **rdtand AQUA** (yours) | 2.346 | WikiText 8×512 | 12.87 | **83** | 24 |

[...]

PR #178, opened 2026-09-04
Refuse empty boundary samples and add paired BF16 measurement

RobTand:
The mandatory boundary probe could return a clean pass without sampling any generations. This fixes that online refusal and prepares an opt-in BF16-relative measurement for the remaining policy defect in #87.

The new instrument derives a finishing cap from a frozen prompt/seed schedule and the observed context remainder. Reaching a context or iteration bound is inconclusive. Candidate runs require the exact control, cap, token IDs, source closure, campaign, host boot and serving stack. The client captures live pre/post process manifests and validates the BF16 serve's actual dtype/quantization arguments. Its per-stratum deltas are advisory: it changes no shipping threshold and fills no shipcard slot. The physical A/B remains scheduled after Tessera #113/#5.

A separate commit fixes shared serve-process discovery: an image/model argument containing `vllm` no longer turns the measurement client or shell wrapper into a server. Stale universal-cap prose is corrected in its own commit.

PrismaBuild evidence: pre-fix empty-population/input tests 8 failed; source/token pairing review 5 failed; process classifier 3 failed/5 passed. Final CPU-only Sparky population on executable source `df44f6d2`: 209 passed, 0 failed, 0 skipped, plus compile checks. Action `b5da2968616280679717305628147cec2fdba6fbe484a01e9d41106c30a8b0fd`, receipt `4a9b4d22ba9d92b0d096be0c51a516a177f187e1c22c372ba9034dd115100766`.

An expanded run also found nine pre-existing Tessera pin/schema failures in `test_tessera_serve_fingerprint.py`; running only that file on pristine `483ff9ca` reproduced the same nine. The concurrent release-consumer/v4 work owns them. They are documented, not counted as passing here. All run populations, pre-fix lines, and the calibration-symlink snapshot exclusion are in `docs/results/pq87_boundary_control_2026-09-04.md`.

Refs #87. Does not close it: no physical A/B or replacement shipping-policy promotion is claimed.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,5 +1,19 @@
 # PrismaQuant Architecture
 
+Re-stamped (2026-09-04, `codex/pq-87-control-relative`) for the **opt-in
+paired boundary instrument** (§7.2). `prismaquant/boundary_control.py`
+records a frozen prompt/seed schedule and replays every raw scored outcome.
+The BF16 budget grows until no sampled completion is censored, bounded by
+the live model context minus chat-tokenized input lengths; exhausting that
+bound or the explicit iteration backstop is inconclusive. A candidate uses
+the identical final cap and schedule, with common campaign/host/stack and
+tokenization bindings. `tools/measure_boundary_control.py` captures live
+process/residency manifests before and after, and can also run the historical
+raw-request arm in the same control session. Per-stratum defect deltas are
+**advisory only**: this adds no pass threshold, fills no shipcard slot, and
+does not replace the historical mandatory 64/zero policy. No physical served
+result is claimed by implementation or CPU tests.
+
 Re-stamped (2026-09-04, `codex/pq-87-control-relative`) for the boundary
 probe's input refusal (§7.2). Empty prompt/repetition populations, non-finite
 or nonpositive sampling temperatures, nonpositive token caps and malformed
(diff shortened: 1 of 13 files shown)


RobTand:
Exact native-weight identity follow-up is pushed as 78f16918 after merging main b6d6824e (merge 088264b4); PR retargeted to main. The opt-in instrument reuses shipcard content manifests and weight-stat attestations, retains exact content in each binding, replays its compound artifact ID, and refuses pre/post mutation including same-size replacement and change-and-restore. Generic shipcard identity and shipping defaults are unchanged.

PrismaBuild red cf18a78f7fd2: 6 failed / 0 skipped, including equal IDs after same-size weight replacement and missing mutation refusals. Green bbca6ff0659d: 243 passed / 0 failed / 0 skipped, CPU-only Sparky, 1 CPU, 3 GiB, CUDA_VISIBLE_DEVICES empty, 11 targeted files, all selected modules collected. Three executable modules compiled. This includes all prior test_tessera_serve_fingerprint failures against the SHA-verified Tessera 1221d2a archive. CAS receipt: 1d99c6cace6e6e5cf2c0474d5771a39edfee9d4f949078c11407fd6e7409bc8f. Full provenance and pre-fix lines are appended to docs/results/pq87_boundary_control_2026-09-04.md.

Independent read-only review found no remaining concrete gap in this bounded fix. Physical A/B is still not measured and remains ordered after Tessera. The campaign launcher must bind/freeze the same artifact before server startup; client hashes alone do not prove what an already-running process loaded earlier. No release pin or gate promotion is claimed.

RobTand:
Root review approved the source and the additional exact-weight identity fix. The measured regression was real: a same-size replacement retained the old cheap checkpoint identifier. The new opt-in receipt binds the existing exact weight manifest plus model metadata, replays the compound identity, and verifies pre/post content and stat attestations; launch-time source binding remains explicit rather than claimed retroactively. Final deployed PrismaBuild action bbca6ff0659d4a64bf365943c79df5b903b7d0a1ad405146c3fe5e1aa7b4c6bc passed 243 targeted tests, zero skips/collection errors, CPU-only Sparky1core3GiB; receipt1d99c6cace6e6e5cf2c0474d5771a39edfee9d4f949078c11407fd6e7409bc8f. This merge does not close #87, change the shipping threshold, or claim the pending physical A/B result.

PR #182, opened 2026-09-04
Measure #87 BF16 boundary instrument A/B with frozen-source receipts

### LINKED ISSUE #87: shipgate: KL/PPL + greedy-smoke certify boundary-broken artifacts — three DSV4-Flash quants within ~3% PPL span a 6× behavioral gap; proposal: add a sampled behavioral component

RobTand:
Records the completed same-session Qwen3-0.6B BF16 instrument campaign; does not close #87 or change shipping defaults.

Measured impact (30 samples each): raw64 and chat64 both 30 cap hits; chat control cap hits at 128/256/512/1024/2048 were 25/20/10/2/0. The uncensored 2048 BF16 control and same-session A/A repeat both had zero scored defects in all three strata. Exact pre-launch weight content, process/image/source/token-ID binding and pre/post checks replayed successfully. No quantized-candidate discrimination or default tolerance is claimed.

Physical action a184593f7e7b: 617.24 seconds of bounded physical work, full Sparky gpu2/cpu2/mem32, exact EUGR image, one eager sequence, explicit 1 GiB KV, 4096 context, 28 GiB container ceiling. Both-box Netdata and process I/O retained. Actual original-worker receipt verified. A lowercase Docker absence diagnostic produced a conservative false inconclusive outer status; original evidence is preserved, independent PB8f2d9e2919f2 confirmed container absent, and cleanup parser is fixed in its own commit b6945ce4. Sparky was handed back to MoE only after that check.

Tests: pre-red new API absence and real health/cleanup refusal failures are recorded in docs/results/pq87_bf16_boundary_ab_2026-09-04.md. Final executable-source PB554cc56a0e4f: 26 passed, 0 failed, 0 skipped plus compilation, CPU-only 1CPU/3GiB. No master baseline or whole suite run. All compute used deployed PrismaBuild; no scheduler source/runtime changes.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,5 +1,15 @@
 # PrismaQuant Architecture
 
+Re-stamped (2026-09-04, `codex/pq87-physical-ab`) for the opt-in bounded
+physical boundary controller. `experiments/pq87_physical_ab.py` freezes and
+hashes native BF16 source before starting one read-only-model server, then
+requires the client to observe that same content. It journals every completed
+HTTP response and cap step, uses explicit one-sequence/4096-context/1-GiB-KV
+limits, and bounds the whole physical action to 1200 seconds plus cleanup.
+Timeout or refusal remains inconclusive; cleanup verifies only its uniquely
+named container. CPU controller tests are not a served result, and no shipping
+policy changes. Gate: `tests/test_pq87_physical_ab.py`.
+
 Re-stamped (2026-09-04, `codex/pq-87-control-relative`) for the paired
 instrument's **exact native weight identity** (§7.2). Each boundary binding
 carries the existing model-metadata hash plus the shared safetensors content
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -11,6 +11,11 @@ named container. CPU controller tests are not a served result, and no shipping
 policy changes. Gate: `tests/test_pq87_physical_ab.py`.
 The cleanup reader accepts Docker's case-varying absence diagnostic only when
 it names this exact container; an unrecognized error never proves absence.
+The first BF16 instrument A/B is recorded in
+`docs/results/pq87_bf16_boundary_ab_2026-09-04.md`: 64-token raw/chat were both
+censored on 30/30 samples; the uncensored chat control and same-session repeat
+at the measured 2,048 cap were 0/30. This remains one instrument population,
+not a quantized-candidate promotion or a new default budget.
 
 Re-stamped (2026-09-04, `codex/pq-87-control-relative`) for the paired
 instrument's **exact native weight identity** (§7.2). Each boundary binding
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -9,6 +9,8 @@ limits, and bounds the whole physical action to 1200 seconds plus cleanup.
 Timeout or refusal remains inconclusive; cleanup verifies only its uniquely
 named container. CPU controller tests are not a served result, and no shipping
 policy changes. Gate: `tests/test_pq87_physical_ab.py`.
+The cleanup reader accepts Docker's case-varying absence diagnostic only when
+it names this exact container; an unrecognized error never proves absence.
 
 Re-stamped (2026-09-04, `codex/pq-87-control-relative`) for the paired
 instrument's **exact native weight identity** (§7.2). Each boundary binding
--- a/docs/experiments/pq87_boundary_ab.md
+++ b/docs/experiments/pq87_boundary_ab.md
@@ -1,7 +1,10 @@
 # PrismaQuant #87: boundary instrument A/B
 
-Status: prepared, opt-in; no physical measurement claimed. Run after the
-Tessera #113/#5 GPU campaign, through one exclusive PrismaBuild action.
+Status: opt-in; the first BF16 instrument run is recorded in
+`docs/results/pq87_bf16_boundary_ab_2026-09-04.md`. It used the coordinator's
+approved Sparky idle window between MoE encoding and final serving, through
+one exclusive PrismaBuild action. Further runs require the same scheduling
+coordination; no quantized-artifact discrimination has been measured here.
 
 The first population is the local Qwen3-0.6B BF16 checkpoint. It is a small
 non-DeepSeek thinking model, so it tests the contributor's claim about the
@@ -69,7 +72,8 @@ use one server at a time with explicit KV bytes/max-sequences/context limits,
 and collect the server log, profiler window, power and Netdata series from
 both boxes. Report work per joule if making a speed claim; GPU utilization
 alone is not diagnostic. No launch permission or capacity is implied by this
-document, and this preparation has consumed no GPU slot.
+document. The original preparation consumed no GPU slot; the physical run and
+its completed GPU handoff are recorded separately in the results document.
 
 After the physical result, the remaining decision is which versioned,
 control-relative rule replaces the current mandatory 64/zero boundary gate.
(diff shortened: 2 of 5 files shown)


RobTand:
Coordinator integrated current main in df4a924dd0165ae11940ce36b477b14a061d0fb8; resolved only the architecture provenance header, preserving both records. Exact integrated head passed 26 targeted controller/architecture/doc tests, 0 failures, 0 skips, 0 collection errors, via PrismaBuild c667f8c2f0470165a23e8ce3fbbe5a03c58a423b731ed06ef65ad6956803f894 (Sparky CPU-only, CUDA_VISIBLE_DEVICES empty, 1 CPU/3 GiB; receipt 7108baa7f76fe74bacd0cc2aedfe62b7b76035f78dce86cedf508274718edae5). Prior dl380 attempt 71aacd363794 failed 3 import-dependent doc checks because that interpreter lacks compressed_tensors; it is retained as an environment failure, not counted green. This PR records an instrument A/B and cleanup correction, not a new gate/default or closure of #87.

RobTand:
Hosted run 33916752588 finished red in both jobs at Check out pinned Tessera; neither test/import step ran. This is the known private-checkout defect #175 that Rob explicitly deferred until public release, not a CI pass. The exact-head PrismaBuild targeted result above supplies the actual regression evidence for this merge.

Issue #184, opened 2026-09-04
run-pipeline accepts COST_PATH_OVERRIDE for Tessera but never consumes the prepriced file


RobTand:
## Finding

`prismaquant/run-pipeline.sh` tells a Tessera campaign operator to supply `COST_PATH_OVERRIDE`, but never uses that value to select the allocator cost input or bypass the incremental cost stages.

Read on checkout `0b9426e2b2545798ef5e6e89e92fba9cdb890d88` (the same behavior is present before the scoped-export changes):

- `prismaquant/run-pipeline.sh:70` checks only whether `COST_PATH_OVERRIDE` is nonempty; line 73 advertises rerunning with it after `prismaquant.tessera_campaign` produces a cost file.
- `rg -n COST_PATH_OVERRIDE prismaquant/run-pipeline.sh` finds no consumer beyond that check and explanatory output.
- `prismaquant/run-pipeline.sh:833` begins the `COST_MODE`-selected probe/cost path setup; the branches set `BASE_COST_PATH` and `COST_PATH` under the work directory independently of the override.
- `prismaquant/run-pipeline.sh:1428` still checks the baseline cost stage, and lines 1441–1447 launch `incremental_measure_quant_cost` with `COST_FORMATS` and `BASE_COST_PATH` when it is not reusable. The incremental cost path is exactly the stage the Tessera token preflight says cannot price the token.

With `FORMATS=TESSERA,...` and a nonempty override pointing to an already priced Tessera campaign file, the early check therefore passes without binding the run to that file. Fixing the separate inline `get_format("TESSERA")` refusal does not fix this stage-graph break.

Severity: **P1** — the prepriced-input gate accepts the requested input without ensuring downstream stages use it. This is a code-path finding, not a measured GPU failure or an end-to-end success claim.

## Why filed instead of fixed here

I could work on this, but my active delegated change is the selected-unit runtime-scoped export boundary for Tessera #126. Correct override handling changes the cost-stage graph, currency/provenance validation, cache reuse, and potentially which stages write a supplied file; folding it into the scoped-export diff would swamp that task. A separate worker should repair it in the same initiative, with a failing endpoint test before the fix. No Rob-only decision has been identified.

## Bounded acceptance

- A declared existing prepriced cost file is the actual allocator input; do not merely suppress a token error or copy a path into prose.
- Validate the supplied file through the existing cost-currency/provenance mechanisms, and refuse missing, incompatible, or ambiguous inputs before GPU work.
[...]

Issue #183, opened 2026-09-04
Tessera campaign omits live packed experts; packed MoE capture/export projection remains unsupported


RobTand:
## Finding

The Tessera anchor campaign does not enumerate packed expert parameters. Runtime-scope support for Tessera #126 must not be described as complete PrismaQuant MoE campaign support.

Evidence inspected at PrismaQuant `fe7182175bd77d2119b91f4a086c41a36612148a`:

- `prismaquant/tessera_campaign.py:712-717` adds targets only when `isinstance(module, torch.nn.Linear)`; packed `nn.Parameter` expert stacks are absent from this population.
- `prismaquant/model_profiles/specs/lfm2_moe.json:42-56` declares `Lfm2MoeExperts`, packed `gate_up_proj`/`down_proj`, and their `w1`/`w3`/`w2` source projections.
- `prismaquant/sensitivity_probe.py:1242-1271` already handles the actual rank-three population, reading expert/out/in dimensions and recording `num_experts` plus `_packed_experts_module`. The campaign does not use that enumeration/capture path.
- `prismaquant/model_profiles/lfm2_moe.py:33-45` documents the existing source/live boundary: per-expert 2-D checkpoint tensors become 3-D live packed parameters. Dense attention and shared units cannot be classified as routed merely because the model contains experts.

The in-progress scoped-export follow-up (`codex/pq-tessera-v5-endpoints`, `prismaquant/tessera_export_lane.py:431-437`) intentionally refuses a packed/aggregate assignment without an exact producer projection rather than pretending a slice has the served module's shape. That refusal is honest, but the supported bridge is still missing.

## Impact / severity

**P1:** a campaign can finish without a coverage refusal for its missing packed expert population; its output cannot establish a complete packed-MoE rate surface. A real packed-LFM quantization campaign therefore remains unsupported even after runtime-scoped admission is wired. No wrong served artifact or numeric impact has been measured here; the release pin and scoped export remain fail-closed.

This is source/declaration evidence, not a live-model or served measurement. No GPU experiment was launched for this finding.

## Acceptance

1. Show a failing CPU regression with a synthetic model containing both dense Linears and declared packed experts; the campaign must enumerate the actual quantizable population or refuse missing coverage explicitly.
2. Reuse the existing packed-expert discovery, activation capture, resident/prefetch and cache mechanisms for per-expert/per-projection Hessian/activation inputs. Preserve one calibration identity and priced == written == served.
[...]

PR #181, opened 2026-09-04
Enforce explicit runtime scope throughout Tessera v5 admission

RobTand:
## Scope

Consumer capability for the approved Tessera #126 runtime-scope revision. Every v5 cell requires an exact image digest and explicit eager/compiled scope. One shared parser and matcher retain these fields; complete admission requires platform, per-unit structure, residency, image and execution mode together, with all declared regimes on that same context.

The scope travels through menu admission, released-pin rendering helpers, synthesized format admission, serving-lane provenance, per-unit candidate lookup and campaign caches. Missing context fails closed in attested mode. Opt-in research can retain a writable but explicitly unattested rung. Removing a unit's final candidate raises a named refusal instead of silently dropping it.

No runtime default, wire bytes, quality threshold, MoE promotion or release pin changes. Context-free v4 behavior is preserved; an explicit runtime query on an unscoped legacy table now refuses instead of serializing the requested image beside an unrelated backed claim. The reviewed DEV pin remains `1221d2a` (v4); actual immutable publisher `a4927fe51219e809ab5c3138e4328473be1473f3` is tested separately through explicit context. DEV answer/pin advances once against the final reviewed producer after measured MoE cells land; release pin changes only at the actual tagged release. **This capability PR does not close Tessera #126.** Real CLI/campaign/export ingress is a separate active follow-up; name-only callers remain closed under v5 rather than guessing context.

## PrismaBuild evidence

All workloads run through PrismaBuild, CPU-only on Sparky, one CPU and 4 GiB; CUDA masked. No GPU or served-quality result is claimed.

- Final 33-file targeted population: **512 passed, zero failures, five skips, zero collection errors**; all five skips are verbatim `Tessera encodes need CUDA`. Eight touched-module compile checks passed in the same action. Action `1c9ff1e390476074b99244c1a244aa23bb6eb30d9af6eae7b4f0d9cfdda9a7e0`; receipt `a25a4b028f2160071809368eceb5e5d498f69081ed18cc3b9f3d259e69995168`.
- After merging current main `6252ef70` (only ARCH stamp conflict, both preserved): **175 passed, zero skipped, zero collection errors** across core/legacy/recipe/docs. Action `e10d68772faae48fee940d3dccfdd51c360b4bf46ebbe25ce00c8db1a42c83fd`; receipt `b18c9b671aa439646b68d976adb04abd932c9510e0c0fd7c0795cf579712ca35`.
[...]

PR #185, opened 2026-09-04
Thread explicit Tessera runtime scope through campaign, allocator and export

### LINKED ISSUE #126: artifact_bpp under-prices the TCQ-at-cap rung by 512 bytes a unit: the ALPHABET and DESCENDANT planes are not charged
`prismaquant.tessera_formats.artifact_bpp` prices the E2M1x2 rung at the TCQ cap (`q256 = 896`) as **exactly 4.0 bpp on every shape**, and the exporter writes 512 bytes more than that on every shape. The window rungs below the cap are priced exactly right, so this is one rung, not a scale error.

## Measured

`experiments`-style harness: encode a real tensor with `tessera.export.encode_linear`, read the per-plane byte ranges out of the artifact with `container.plane_ranges`, and compare `exact_bytes` against `artifact_bpp("TESSERA_E2M1_K2", q, shape=(rows, cols))`.

```
       shape  rung  exported B  priced B   delta  exp bpp   priced   planes the artifact carries
   (96, 768) R895       38620     38620       0   4.1905   4.1905   ALPHABET=4096, BODY=32220, SCALE_REFINE=2304
   (96, 768) R896       37376     36864     512   4.0556   4.0000   ALPHABET=256, DESCENDANT=256, BODY=34560, SCALE_REFINE=2304
   (64, 512) R896       16896     16384     512   4.1250   4.0000   ALPHABET=256, DESCENDANT=256, BODY=15360, SCALE_REFINE=1024
   (96, 320) R896       15872     15360     512   4.1333   4.0000   ALPHABET=256, DESCENDANT=256, BODY=14400, SCALE_REFINE=960
   (64, 640) R896       20992     20480     512   4.1000   4.0000   ALPHABET=256, DESCENDANT=256, BODY=19200, SCALE_REFINE=1280
 (512, 2048) R896      524800    524288     512   4.0039   4.0000   ALPHABET=256, DESCENDANT=256, BODY=491520, SCALE_REFINE=32768
(1024, 3072) R896     1573376   1572864     512   4.0013   4.0000   ALPHABET=256, DESCENDANT=256, BODY=1474560, SCALE_REFINE=98304
```

The delta is **exactly 512 bytes on all six shapes**, and the decomposition says what it is: the accountant charges `BODY + SCALE_REFINE` and nothing else, so it omits the TCQ body's `ALPHABET` (256 B) and `DESCENDANT` (256 B) planes. On the window rungs it *does* charge `ALPHABET` — 4096 B for the L=12 table — and the delta is 0, which is what makes this a hole in one branch rather than a missing term everywhere.

## Why it matters

512 bytes is a fixed cost, so the error is inversely proportional to unit size, and it lands hardest exactly where an MoE plan spends most of its units:

| shape | under-report |
|---|---|
| (96, 320) | 0.1333 bpp |
| (64, 512) | 0.1250 bpp |
| (64, 640) | 0.1000 bpp |
| (96, 768) | 0.0556 bpp |
| (512, 2048) | 0.0039 bpp |
| (1024, 3072) | 0.0013 bpp |

A GLM expert unit is in the top half of that table. The byte budget is spent in the currency this function returns, so a plan that fits by the accountant can overrun on disk, and the byte gate cannot see it. It is also live on the DP: at matched bytes the cap rung is being offered ~0.1 bpp cheaper than it is against every window rung it competes with.

This is the second time this cost model has priced a Tessera rung by a rule of its own rather than by the wire's layout (cf. the 6.25% overcharge that was live on the DP and the byte gate). The fix is the same shape: price it through the layout the exporter actually writes, not through a per-family formula.

## Interaction with tessera#43

tessera#43 reports that the q256 rate axis is not monotone in bits at this cap, with drops measured **through this function**. Those magnitudes are inflated by exactly this bug: at (96, 768) the priced drop is 0.1905 bpp and the drop in bytes the exporter actually writes is 0.1350. The non-monotonicity is real either way — the cap rung genuinely weighs less than the rung below it, because the TCQ body sheds a 4096-byte window table and only pays 512 back — but the numbers in that issue should be re-read after this is fixed.

Harness: `/home/rob/tmp/t45_accountant.py`.

RobTand:
Follow-up to #181 for Tessera #126. Connects explicit target intake, topology-derived unit structures, scoped menu/candidate/cache/provenance and selected export admission through real campaign/allocator/export endpoints. Plan and shipcard independently bind exact allocation bytes. Pins, defaults and attestation cells do not move.

PrismaBuild CPU-only Sparky, 1 core/4 GiB: main endpoint population 408 passed/0 failed/5 skipped (all "Tessera encodes need CUDA"), review-fix population 112 passed/0 skipped, final post-main-merge shell/docs 27 passed/0 skipped; no collection errors. Targeted red evidence and immutable receipts are in docs/results/tessera_v5_endpoints_2026-09-04.md.

Separate review commits fix stale-plan/new-build binding, preserve exact runtime in printed serve recipes, and replace stale census instructions with raw scoped input plus allocation binding. Scoped receipt implementation/final producer pin remain separate #126 work; packed bridge #183 and prepriced dispatch #184 are assigned independently. This PR does not close #126 or claim a serve.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,13 @@
 As of: 2026-09-04 · `codex/pq-tessera-v5-endpoints`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-04, `codex/pq-tessera-v5-endpoints`) for **scope-preserving
+serve instructions** (§9.4). The printed Tessera launch recipe now supplies
+the validated exact `IMAGE`, eager/compiled `TESSERA_LANE_EAGER`, residency
+and `TS` producer checkout with shell-safe quoting. Legacy unscoped callers
+retain the existing wrapper image/mode defaults. This changes instructions,
+not serving defaults or attestation.
+
 Re-stamped (2026-09-04, `codex/pq-tessera-v5-endpoints`) for **allocation-bound
 Tessera plan reuse** (§9.4). The plan stage now records the full SHA-256 of
 the exact allocation bytes as a late-bound settings key. Changed allocations
@@ -9014,7 +9021,7 @@ decision rather than operator policy. Strix Halo enters this lane first, serving
 
 ### 9.4 tessera — the Tessera wire on Tessera's own vLLM plugin
 
-**Declared, driveable, and fail-closed on one thing.** This lane existed as a
+**Declared, driveable, and fail-closed.** This lane existed as a
 *bar* before it existed as a shipping path: `prismaquant/lane_specs/tessera.json`
 states the serve, the endpoint, the gate set, the KL evaluator and the executed
 activation contracts. Since 2026-09-03 `run-pipeline.sh` has a real
@@ -9034,7 +9041,7 @@ a partial allocation is planned as-is with every other body Linear spelled BF16,
 broadcast by role and stamped as the extrapolation it is; silence must never become
 a 4-bit rung.
 
-**Four preflight gates, each fail-closed on its own** (`tessera_export_lane.py`,
+**Preflight gates, each fail-closed on its own** (`tessera_export_lane.py`,
 run before any GPU work): the checkpoint's structure must be one the packaged
 contract declares (`require_declared_structure`, read from the artifact's own
 `config.json`, because the `qwen3` profile claims both the dense and the MoE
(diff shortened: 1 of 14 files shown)


RobTand:
Coordinator review complete at 225b04484637dc53c4caaecc48f98653ed7e08d5, including the independently found stale-plan/new-allocation anchor defect and printed runtime loss. Both have separate fixes with actual-driver pre-red and targeted green evidence. Verified final hosted run 33918323823 failed only at Check out pinned Tessera in both jobs; no tests ran there (known #175, explicitly deferred by Rob). Actual regression evidence is the scoped PrismaBuild CPU populations in the report. This is an endpoint integration merge, not MoE promotion, a shipping default change, or #126 closure.

PR #188, opened 2026-09-04
Consume validated prepriced cost overrides without running cost builders

### LINKED ISSUE #184: run-pipeline accepts COST_PATH_OVERRIDE for Tessera but never consumes the prepriced file

RobTand:
Closes #184.

`COST_PATH_OVERRIDE` now validates and consumes the supplied prepriced file, bypasses incompatible cost builders/finalization, and passes the original path to the allocator. A separate local receipt binds lexical/resolved path and SHA-256, rechecked immediately before use. Supplied input is not rewritten. Existing mode/provenance/currency/Hessian owners remain authoritative; defaults and serving policy do not change.

Two encountered defects are fixed in separate owner commits: nonfinite numeric cost signals, and Tessera rows evading currency checks by omitting their stamp. The validator delegates to those homes. A separate fixture-only integration commit gives existing synthetic endpoint costs their correct campaign stamps.

PrismaBuild CPU-only Sparky, 1 core / 4 GiB:

- Initial red: 47 failed, 1 preservation passed.
- Supplemental owner red: 24 failed, 64 passed.
- Combined targeted green: 97 passed, 0 failed/skipped; Python compilation and shell syntax passed.
- Existing scoped endpoint integration: 21 passed, 0 failed/skipped after a valid three-failure fixture red.

Final 15-file integration on merged PR185/main10dd2395: **234 passed, 0 failed/skipped**, no collection errors; shell syntax passed. No GPU, serving, performance or quality claim. Full immutable action keys/receipts are in `docs/results/prepriced_cost_override_2026-09-04.md`.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,18 @@
 As of: 2026-09-04 · `codex/pq184-prepriced-cost-override`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-04, `codex/pq184-prepriced-cost-override`) for the
+**read-only prepriced-input validator** (§4.10; #184). The explicit
+`prepriced_cost` CLI validates existing schema, objective currency,
+research provenance and uniform Tessera Hessian identity, requires
+the exact producer-stamped mode and model reference, and records the
+input bytes' SHA-256. Its separate local receipt retains both the original
+absolute argument path and resolved file; re-verification checks both
+resolution and bytes before allocation. The input is never rewritten,
+model-reference equality is not checkpoint-content attestation, and
+historical unstamped Hessian rows retain their owner-reported no-claim
+status. This helper does not itself wire the shell's cost-stage bypass.
+
 Re-stamped (2026-09-04, `codex/pq184-prepriced-cost-override`) for
 **Tessera currency-stamp completeness** (§4.10), encountered during
 #184 intake work. `cost_currency` recognizes Tessera rows through the
@@ -4692,6 +4704,28 @@ hardcoded family roster. Missing or unknown currency refuses with the
 unit and format named; dropping a stamp cannot opt a price out of the
 gate. Diagnostic error rows are excluded, and legacy stock rows without
 a Tessera format or currency remain outside this gate's jurisdiction.
+
+**Explicit external prepriced input** (`python -m prismaquant.prepriced_cost`).
+The intake CLI takes `--path`, `--cost-mode` and `--model`, reads the supplied
+trusted pickle without rewriting it, and applies those shared gates plus
+`research_cost_acceptance.accepted_cost_provenance` and
+`tessera_menu.assert_uniform_hessian_identity`. The pipeline has no research
+acceptance override, so research-stamped assembly refuses. Missing, unknown
+or mismatched `provenance.cost_mode` also refuses; the helper changes neither
+the input's stamp nor the run's defaults. It requires at least one explicit
+producer-owned model reference (direct metadata/provenance or retained
+baseline/shard metadata), with all present references equal to the exact
+requested string. This is not a checkpoint-content hash claim.
+
+`--report` writes a separate local JSON receipt containing the exact SHA-256,
+actual and declared format rosters separately, usable-row count, model-reference
+evidence and the shared currency/Hessian reports. Existing unstamped Hessian
+rows remain reported as unstamped, not invented matching claims. Report/input
+path aliases and hardlinks refuse. `--verify-report` checks the retained
+original absolute argument path still resolves to the same file and hashes
+the same bytes immediately before allocation; this catches retargeted file
+or parent-directory symlinks as well as content changes. Per-unit coverage,
+rendering identity and serving admission remain allocator/export decisions.
 
 Every other entry on the format menu is a *point*: `NVFP4` is one rate, `FP8_E4M3`
 is one rate, and a menu is the handful of them a launcher lists. A Tessera family
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,17 @@
 As of: 2026-09-04 · `codex/pq184-prepriced-cost-override`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-04, `codex/pq184-prepriced-cost-override`) for **actual
+prepriced cost dispatch** (§4.10; #184). `COST_PATH_OVERRIDE` now runs the
+shared read-only intake before probe/GPU work, passes the original supplied
+path to the allocator, and bypasses the entire local/render/AURA cost-builder
+and finalization stage. Its separate local receipt is reverified immediately
+before allocation, including lexical-path resolution and exact SHA-256.
+Supplied files and their sidecars are not rewritten. No-override generation,
+`COST_MODE` defaults and existing allocator/serving gates remain unchanged;
+the requested mode must match the explicit producer stamp, never be inferred
+by changing a default. Evidence: `docs/results/prepriced_cost_override_2026-09-04.md`.
+
 Re-stamped (2026-09-04, `codex/pq184-prepriced-cost-override`) for the
 **read-only prepriced-input validator** (§4.10; #184). The explicit
 `prepriced_cost` CLI validates existing schema, objective currency,
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,15 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-04 · `codex/pq-tessera-v5-endpoints`. Stamps
+As of: 2026-09-04 · `codex/pq184-prepriced-cost-override`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-04, `codex/pq184-prepriced-cost-override`) for
+**finite cost-schema signals** (§4.10), encountered during #184 intake
+work. `schemas.validate_cost_payload` now refuses non-finite values in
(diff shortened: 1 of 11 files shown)


RobTand:
## Release review (worker `pq-prepriced-census`) — merged with main, reviewed, re-tested

Tip `2ce8c70c`. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

### Merge with `origin/main` (`8a77b0ad`, i.e. after #196)

`52eece09` merges `origin/main`. One conflict, in `docs/ARCHITECTURE.md`'s header stamp
block only: resolved by re-dating the `As of:` line to `2026-09-05 · codex/pq184-prepriced-cost-override`
and keeping **every** stamp from both sides (this branch's four above `claude/pq-tessera-audit`'s
priced-inputs stamp). No stamp was dropped.

`prismaquant/run-pipeline.sh` and `tests/test_tessera_scope_endpoints.py` merged textually
without conflict even though #196 also edited both. I checked the merged result rather than
trusting that:

* #196's driver edits are (a) `: "${TESSERA_HESSIAN:=}"` / `: "${TESSERA_INPUT_SCALES:=}"` at
  line 174-175 — **before** the cost stage — and (b) the `TESSERA_PRICED_INPUT_ARGS` array built
  and consumed **inside** the `EXPORT_CONTAINER == tessera` arm. This PR's edits are the
  preflight at 992-1003 and the `if/else/fi` around the cost stage at 1528-1811. Disjoint.
* **The override bypass does not skip the priced-export-inputs gate #196 added.**
  `require_priced_export_inputs` does not read the cost table; it reads the *allocation's*
  `tessera_hessian` metadata block (`tessera_export_lane.py:565`), which `allocator.py:5458`
  stamps from `assert_uniform_hessian_identity(costs)` over whatever `--costs` it was handed —
  here the supplied prepriced table, whose identity `prepriced_cost.validate_prepriced_cost`
  already ran through the same function at intake. So an H-aware prepriced allocation exported
  without its matching capture still refuses by name, and a weights-only one handed a stray
  capture still refuses in the other direction. Nothing about the bypass reaches that leg.
* I also checked the bypass leaks nothing: of the nine variables assigned inside the wrapped
  block (`AURA_CACHE_FORMATS`, `AURA_COST_STREAMING`, `AURA_EXECUTION_ARGS`,
  `AURA_EXPERT_EXECUTION_ARGS`, `BASE_COST_REUSABLE`, `COST_CACHE_COL_WEIGHT_ARGS`,
  `OMITTED_EXPERTS`, `PROD_RENDER_COST_ARGS`, plus the per-command `COST_MODE=` prefix), none is
  referenced after `fi` at 1811. `COST_PATH` after the block is read-only (allocator, frontier
  materialization, AURA additivity) and no `require_stage_settings` call outside the block names
  it — correct, since that guard is a *reuse* gate for artifacts this run produces, and the
  supplied table is gated by SHA + provenance instead.

### Review of the PR itself (unreviewed codex output)

Claims in the body check out against the code. Specifically:

* The supplied path really is the allocator's `--costs` (`COST_PATH="$COST_PATH_OVERRIDE"`,
  unresolved, so the operator's own path reaches the receipt), the file is never rewritten
  (report writes go to a `tempfile` + `os.replace` in the report's own parent, with same-path /
  symlink / hardlink aliases refused), and the incremental/render/AURA builders are genuinely
  unreachable rather than merely quiet.
* Refusals are named, not defaulted: unknown or missing `provenance.cost_mode`, requested
  `COST_MODE` naming no objective, mode/stamp mismatch (including the legacy
  `production-render` spelling), research-stamped assembly, zero usable rows, sha mismatch,
  and post-preflight retarget of the original argument (file *and* parent symlink) each raise
  with the path in the message. No silent fallback path exists — `main` exits 2 through
  `parser.exit`, and both driver call sites `exit 2` on nonzero.
* No duplicated rule homes: the validator delegates to `schemas.validate_cost_payload`,
  `cost_currency.require_run_currency`, `research_cost_acceptance.accepted_cost_provenance`
  and `tessera_menu.assert_uniform_hessian_identity` rather than restating any of them.
* The two side-fixes are correctly scoped. `cost_currency` decides Tessera membership through
  `tessera_formats.parse_tessera_format_name` (which returns `None` for non-Tessera names and
  raises only for a Tessera-shaped illegal rung) instead of a `startswith` guess, and skips
  `error` rows; `schemas` adds finiteness only to the cost payload's existing numeric fields.
  Neither invents a threshold.
* `tests/test_run_pipeline_prepriced_override.py` executes the *real* driver text (sliced by
  section markers) with only the expensive worker commands stubbed, and distinguishes builder
  (77) from allocator (88) from refusal (2). That is a meaningful test of the stage graph, not
  a mock of it.

### What I changed

`2ce8c70c` — **`docs: name the priced export inputs in the TESSERA-token refusal`** (prose only,
per AGENTS.md principle 14's fix-on-sight exception). A prepriced run is by definition one where
the campaign ran *out of band*, so the `hessian_capture.pt` and `input_scales.safetensors` it
wrote beside its `--cache-dir` are **not** under this run's work directory. Post-#196, the
TESSERA-token refusal still told the operator only to re-run with `COST_PATH_OVERRIDE` — following
it verbatim fails closed at the export preflight, correctly but hours later. The refusal now names
`TESSERA_HESSIAN` / `TESSERA_INPUT_SCALES` and where the campaign writes them, and
`docs/ARCHITECTURE.md` §4.10 records the handoff (with a new header stamp). No gate, default,
stage-graph or byte change.

No other defect found that needed fixing or filing.

### Tests

sparky, CPU-only (`CUDA_VISIBLE_DEVICES=""`), `/home/rob/venvs/pq-release/bin/python`,
tessera **at the pin** `PYTHONPATH=/home/rob/tessera-pin-1221d2a/src`, clean clone at the pushed
tip `2ce8c70c`:

```
python -m pytest -p no:cacheprovider -q --no-header -n 4 <54 files>
1048 passed, 14 skipped, 0 failed in 227.29s
```

The 54-file population is: this PR's own five test files
(`test_prepriced_cost`, `test_run_pipeline_prepriced_override`, `test_cost_currency_missing_stamp`,
`test_cost_schema_finite`, `test_tessera_scope_endpoints`); the tests of every touched module
(`test_cost_currency`, `test_schema_validation`, `test_wave3_selection_and_provenance`, the five
`*pipeline*` driver files, `test_stage_settings_guard`, `test_doc_claims_driver_truth`,
`test_docs_entrypoints`); `test_architecture_doc` + `test_docs_staleness`; and — because the
currency fix newly refuses a *usable* Tessera row with no campaign stamp anywhere
`allocator.require_run_currency` runs — every test file that mentions `cost_currency`,
`require_run_currency`, or a `"TESSERA_*"` format literal (36 more, including
`test_tessera_priced_export_inputs` and `test_tessera_campaign`, i.e. #196's own tests).

**Zero failures, so nothing to classify as inherited or mine.** Also on sparky at the same tip:
`bash -n prismaquant/run-pipeline.sh` and `py_compile` of the three changed modules both pass.

Not a GPU, served, performance or quality claim; nothing here measures one.

PR #189, opened 2026-09-04
Bind Tessera raw census replay to priced runtime and artifact scope

RobTand:
Completes the scoped raw-census consumer leg of Tessera #126, on merged endpoint PR185. #126 remains open for final reviewed producer/pin integration.

Keeps complete producer `route_census/2` observations and exact allocation/sidecar text, replays allocation SHA against independent card build metadata and sidecar hashes against actual artifact files, then checks exact runtime/mode/residency, each selected source projection, owner, driven phase and launch pair against current v5 cells. Old flat receipts cannot silently acquire v5 scope. No pin, default, bytes or promotion threshold changes.

Boundaries are explicit: no packed source/unknown name projection guesses or nonempty executed-shape predicates; producer combined dense compiled traces cannot establish per-regime route proof, despite dense compiled admission being supported. Scoped single-pair compiled MoE is a parser control, not a new serving attestation.

Worker targeted PrismaBuild evidence: 148 passed, 0 failed/skipped; final endpoint/shipcard integration on main10dd2395: 209 passed, 0 failed/skipped. No collection errors; three Python modules compiled. CPU-only Sparky 1 core / 4 GiB. Valid pre-fix failures, actual-shaped fixture corrections and immutable receipts are in `docs/measurements/tessera-scoped-census-binding-2026-09-04.md`. No GPU or served claim.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,18 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-04 · `codex/pq-tessera-v5-endpoints`. Stamps
+As of: 2026-09-04 · `codex/pq-census-v5-binding`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-04, `codex/pq-census-v5-binding`) for **runtime-bound
+route census replay** (§7.1; Tessera #126). The complete producer
+`tessera.serving.route_census/2` record retains exact runtime, execution mode,
+residency, both driven phases, owner maps, backend-qualified launches and
+checkpoint-sidecar seals. Exact allocation text is rehashed against the
+independent card build anchor, and artifact sidecars are independently
+rehash-checked at fill and verification. Current v5 cells, not an embedded
+agreement or a global decoder blacklist, decide launch eligibility. Legacy
+unscoped rows remain compatible only without a scoped build/current v5 table.
+No default, bytes, promotion threshold or reviewed runtime pin moves.
 
 Re-stamped (2026-09-04, `codex/pq-tessera-v5-endpoints`) for **raw scoped
 census handoff instructions** (§9.4). The driver prints the producer's actual
@@ -70,6 +81,7 @@ router/expert or packed-module facts determine unit structure, cross-checked
 against the declared model profile. Missing facts cannot silently become
 dense. The helper delegates value validation to `ServingContext`; it does
 not change a runtime default or attest any new serving cell.
+
 Re-stamped (2026-09-04, `codex/pq87-physical-ab`) for the opt-in bounded
 physical boundary controller. `experiments/pq87_physical_ab.py` freezes and
 hashes native BF16 source before starting one read-only-model server, then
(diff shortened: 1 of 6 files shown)


RobTand:
## Release review (worker `pq-prepriced-census`) — merged with main, reviewed, re-tested

Tip `b0d22381`. `mergeable: MERGEABLE`.

### Merge with `origin/main` (`8a77b0ad`, i.e. after #196)

`a1fdc152` merges `origin/main`. One conflict, in `docs/ARCHITECTURE.md`'s header stamp block
only, as expected: `As of:` re-dated to `2026-09-05 · codex/pq-census-v5-integration`, and every
stamp from both sides kept, this branch's above `claude/pq-tessera-audit`'s. No other file
conflicted and none of #196's files overlap this PR's.

### Review of the PR itself (unreviewed codex output)

The body's claims check out. The scoped path really does re-derive rather than trust: the
allocation text is rehashed against `card.build.layer_config_sha`, the recipe's
`tessera_serving_scope` must equal `card.build.tessera_serving_scope` (an independent anchor,
not a copy of itself), the census's `checkpoint_sidecars` must equal both the retained texts and
the files on disk, and `_verify_route_census_record` recomputes the whole verdict and compares
it to the carried one — the CLI test round-trips the card through JSON before that comparison,
so it is a real replay and not an equality on freshly built objects. The refusal set is
genuinely broad (23 parameterised mutations plus seven binding mutations, each a named
refusal), the unsupported boundaries are refusals rather than guesses (packed/aggregate source
projections, nonidentity owner mappings, nonempty executed-shape predicates, combined dense
compiled traces), and `parse_route_records` now refuses scope fields rather than flattening
them. `TesseraRouteReceiptError` subclasses `ValueError`, so the broad `except` at the bottom of
`check_scoped_route_receipt` re-raises typed refusals unwrapped instead of relabelling them
"malformed" — that isinstance check is load-bearing, not decoration.

Two things I checked that are not in the body:

* **`shipcard.verify` really does reach the new verifier with its independent anchors.**
  `card`/`model_dir` are in scope at the `LANE_SLOT_VERIFIERS[ROUTE_CENSUS_SLOT]` call site, and
  a scoped record verified without `model_dir` refuses (`test_scoped_verification_requires_
  independent_artifact_files`) rather than passing on the carried verdict.
* **`make_route_census_record`'s new signature is safe.** `route_records` became a required
  keyword-only parameter sitting after two that gained defaults; that is legal only because the
  whole signature is keyword-only, which it is, and there is no positional caller.

### What I changed

`b0d22381` — **`docs: say the scoped census leg is dormant at the current pin`** (prose only,
per AGENTS.md principle 14's fix-on-sight exception). The doc listed every gate the scoped
replay applies but not that none of them can be reached under the pin this repository declares,
which reads as a live priced-vs-served comparison. Measured rather than reasoned:

```
PYTHONPATH=/home/rob/tessera-pin-1221d2a/src python -c \
  "from prismaquant import tessera_route_receipt as r; print(r._current_scoped_contract()[0].schema)"
tessera.lane-eligibility.v4
```

against `LANE_ELIGIBILITY_SCHEMA_TESSERA == "tessera.lane-eligibility.v5"`, which
`check_scoped_route_receipt` requires by exact equality; and the pinned producer's
`tools/tessera_route_census.py:439` emits `"tessera.serving.route_census/1"` against the
`SCOPED_CENSUS_SCHEMA == ".../2"` this module requires by exact equality. **So no scoped receipt
can be filled or replayed at the current pin** — `route.census` on a scoped card stays
`UNFILLED`. That is the correct fail-closed direction for the consumer half of Tessera #126
while its producer/pin half is open, and it is exactly what the PR intends; what was missing is
that the consequence is written down, together with the fact that the scoped comparison has so
far only run against tables constructed inside `test_tessera_scoped_route_receipt.py`. No schema
constant, gate, default or byte moved — the re-pin (#192) owns those.

### For whoever re-pins (#192): every version this PR hard-codes

Left alone here deliberately. For the record, the released contract publishes
`tessera.lane-eligibility.**v8**` (contract_version 20), not v7 — `load_eligibility_table`
refuses it today by name.

| Site | Pinned value | After a re-pin |
|---|---|---|
| `tessera_route_receipt.py:197` `SCOPED_CENSUS_SCHEMA` | `"tessera.serving.route_census/2"`, exact equality | release's `tools/tessera_route_census.py:723` still emits `/2` — no change needed |
| `tessera_route_receipt.py` `CENSUS_PHASE_REGIMES` | `{"decode": "decode", "prefill": "batch"}`, plus `set(table.regimes) == {"decode","batch"}` | v8 still publishes exactly `["decode","batch"]` — no change needed, but it refuses if a regime is ever added |
| `check_scoped_route_receipt` | `table.schema == lane.LANE_ELIGIBILITY_SCHEMA_TESSERA` | follows the constant if v5→v8 **replaces** it; needs revisiting if v8 is added *beside* v5 the way v4/v3 are kept as legacy |
| `shipcard.py` `_verify_route_census_record` | same constant, plus the literal `"current v5 cells cannot attest an unbound legacy flat census"` | same, and the message's "v5" is prose that must move with it |
| `shipcard_cli.py:590` (`--census` help), `tessera_route_receipt` module docstring, `docs/ARCHITECTURE.md` §7.1 | "route_census/2", "v5 cells" in prose | update with the constants |

### Tests

sparky, CPU-only (`CUDA_VISIBLE_DEVICES=""`), `/home/rob/venvs/pq-release/bin/python`, tessera
**at the pin** `PYTHONPATH=/home/rob/tessera-pin-1221d2a/src`, clean clone at the pushed tip
`b0d22381`:

```
python -m pytest -p no:cacheprovider -q --no-header -n 4 <26 files>
610 passed, 0 failed in 36.20s
```

Population: this PR's own `test_tessera_scoped_route_receipt`; the tests of every touched module
(`test_tessera_route_receipt`, `test_shipcard`, `test_shipcard_uniform_control`,
`test_shipcard_git_provenance`, `test_ship_boundary_behavior`, `test_publish_artifact`); every
other file that names `shipcard`, `route_census`, `ROUTE_CENSUS_SLOT`, `LANE_SLOT_VERIFIERS` or
`tessera_route_receipt` (`test_lane_gate_recording`, `test_lane_spec_and_gguf_kl`,
`test_artifact_completeness_namespaces`, `test_export_output_safety`, `test_boundary_control`,
`test_measure_served_gold`, `test_profile_export_lanes`, `test_read_traffic`, `test_mxfp8_ue8m0`,
`test_tessera_export_scope`, `test_tessera_plan_input_binding`, `test_tessera_menu_real_table`);
the lane-eligibility tables the receipt reads (`test_lane_eligibility_v4`,
`test_lane_eligibility_v5`, `test_tessera_lane_admission`, `test_tessera_legacy_scope`); and
`test_architecture_doc`, `test_docs_staleness`, `test_doc_claims_driver_truth`,
`test_docs_entrypoints`.

**Zero failures, so nothing to classify as inherited or mine.** No GPU, served, performance or
quality claim; per the finding above, no served claim is even reachable from this branch yet.

PR #187, opened 2026-09-04
Refuse incomplete MoE campaigns and capture explicit packed expert inputs

RobTand:
Related to #183; this deliberately does not close it.

The campaign now refuses a live packed-expert population before it can emit an incomplete dense-only cost payload. The capture helper can independently collect each profile-declared expert projection's actual routed inputs and full-row Hessian, reusing the existing packed collector, routing/SwiGLU implementation, and native projection views. Scoring-row limits never truncate the Hessian; unobserved selected experts refuse.

The main packed-population gate remains closed. Exact producer source-to-stack planning, identity-bound cached-blob packing, and an actual matched-byte served campaign are separate remaining work. No new menu, pin roster, encoder default, serving eligibility, or GPU/performance claim is introduced.

Separate on-path commits:

- Honor the profile's existing BF16 pins in dense campaign target enumeration.
- Correct stale architecture resume-identity prose and an obsolete rule-number citation.
- Exercise both FP32 and BF16 live capture inputs.

Based on main `10dd23957d6cf9ad784f8eac59fb914f66f6d89e`, after PR #185; no duplicate endpoint dependency commits are included. Historical red/green actions, commands, populations, and receipts are in `docs/measurements/tessera-packed-capture-2026-09-04.md`.

CPU evidence through deployed PrismaBuild, Sparky / 1 CPU / 4 GiB / no CUDA:

- Genuine pre-fix main omissions: 2 failures; packed capture: 4 failures; pin intake: 2 failures.
- Capture + existing reservoir/campaign/streaming files: 41 passed, 7 skips (5 `Tessera encodes need CUDA`, 2 `streaming production render is GPU-or-bust`). Streaming equivalence was not exercised.
- Shared discovery/router/docs: 41 passed, no skips.
- Pins + campaign regression: 39 passed, 5 `Tessera encodes need CUDA` skips.
- FP32/BF16 capture + pins/docs: 28 passed, no skips; both dtype cases also failed against the exact pre-capture campaign implementation.

The combined targeted population on the updated integration base is queued as PrismaBuild `8c87b0777dae`; its result will be added before merge. These overlapping historical populations must not be summed into a suite count.

Encountered checkpoint-resume identity defect is separately filed and actively assigned as #186. It is not hidden in this capture diff.

PR #190, opened 2026-09-04
Add opt-in paired no-new-boundary-failures decision

RobTand:
Implements the delegated policy choice for #87 without changing production defaults or shipcard slots. Paired defect-kind comparison rejects new failures without cross-sample offsets; exact control/candidate replay gates every decision, and optional candidate CLI returns 2 on refusal. CPU evidence: 105 passed, zero skipped, seven selected modules collected, compile green in frozen PrismaBuild bfebe89b876b. Pre-fix failures for all 22 added cases and exact receipts are recorded in docs/results/pq87_paired_policy_2026-09-04.md. Quantized, held-out and representative-population validation remains pending and #87 stays open. This PR does not claim a shipping promotion.

PR #191, opened 2026-09-04
Bind Tessera campaign resume to current inputs and exact priced wire

### LINKED ISSUE #186: Tessera campaign resume admits anchors without source/calibration/recipe identity
Severity: P1 — the allocator can read old measured costs under a new campaign's provenance.

Found while wiring the #183 producer/cached-byte identity boundary. At `prismaquant/tessera_campaign.py:923-940` on `bec85eaa`, checkpoint resume deserializes every saved `CampaignAnchor` without checking a source, calibration, recipe, target-scope, or encoder identity. `flush_checkpoint()` persists only `{schema, anchors}`. Later `:1149-1157` (and the following Hessian provenance block) stamps the current model and calibration settings on the payload. There is no identity in the checkpoint that could be compared.

Inputs: reuse the same `--checkpoint` (including the default next to `--out`) for a changed checkpoint/model revision, calibration seed/draw, Hessian mode/settings, encoder recipe, or serving scope while retaining matching qnames/rates. Old rows can be treated as already measured and mixed with new rows; resumed rows outside the current target population are also not filtered at intake. This is source inspection, not a measured served-quality claim.

Required fix: use the existing source/calibration/recipe identity owners to bind the campaign checkpoint and exact target population, validate before admitting any anchor, and verify each cached wire commitment needed by resumed measured rows. Refuse missing/mismatched identity rather than relabeling unverified anchors as the current run. Tests must show a real main-entry resume accepting mismatched input before the fix, then refusing; same-identity resume should remain valid.

Why filed instead of fixed in the current diff: I could implement this with a separate identity-gate pass, but its cross-cutting checkpoint/priced-wire contract is larger than the bounded packed capture/pin fixes currently ready for integration, and the producer identity API is being implemented concurrently by the isolated #183 worker. Reusing that shared identity owner avoids creating a second incompatible manifest grammar. This is not a request to change an encoder default or quality threshold. The coordinator is being asked to dispatch an explicit worker now; it must not be lost behind the MoE campaign.

RobTand:
## Outcome

Fixes #186. Identity-free or mismatched checkpoint anchors can no longer enter a Tessera campaign's current cost payload. A same-input resume reuses its exact measured costs and wire without re-encoding.

This PR is stacked on #187 until that capture/endpoint base lands. It also depends on the isolated Tessera producer API at `6343137` (`codex/ts-cached-producer-api`, receipt tip `fe3e996`); the release pin and frozen MoE serving candidate are untouched. An older producer lacking that API fails closed. **Do not read this as completion of #183**, whose packed main-entry refusal remains closed pending the real producer bridge and served measurements.

## Changes

- Reuse the existing identity-bound cost-stage journal; an optional manifest pathname preserves the current `--checkpoint` file/default, with existing per-unit shards under a sibling `.parts` directory.
- Bind selected weight/score-row/Hessian values, corpus and token identities, menus/scope, explicit campaign settings, resolved encoder settings and both package source seals. Actual score rows are bound even with `--hessian off`.
- Use the producer's existing generic input/byte receipt APIs. Before any resumed cost is admitted, validate unit/rung/menu/Hessian applicability, exact receipt coverage, full-blob bytes/digest and the freshly derived encoding identity. No current identity is stamped onto an old anchor.
- Preserve exact measured costs and bytes for a valid resume. Refuse missing, changed or symlinked wire without silently recomputing it.

Separate commits keep the shared journal adapter, campaign gate and on-sight prose corrections independently reviewable. Off-task prose corrections: the old Hessian comment incorrectly claimed a fixed plane roster; the opening docstring overclaimed export/serve identity from the cached rendering proof. Both now describe the actual owner/proof boundary.

## Evidence

All computation went through deployed PrismaBuild `aa6d3cfa2f77`, CPU-only serial Sparky, 1 CPU / 4 GiB; immutable producer source archive `6343137` was extracted in the worker. No GPU or served/performance result is claimed.

- Shared journal adapter red: two `unexpected keyword argument 'manifest_path'` failures; green: **19 passed / 0 skipped** (`4b5265e28732`).
- Real main-entry expanded red: **17 failed / 0 skipped** (`935b09c3641f`, completed first attempt; redundant retries withdrawn). Only `campaign.py` was restored to exact pre-fix `2c22feac`; fixture models/menus/calibration rows are synthetic, but their 32x256 units are really encoded and priced. Old code accepted changed weights, score rows, corpus/tokens, actual H, H mode, recipe/menu/source/scope, and missing/corrupt/symlinked wire. All failure lines are recorded in the measurement document.
- Fixed eight-file targeted run: **95 passed / 5 skipped**, no uncollected module reported (`b62e9617fbfef24f65ae0b1491a2c74825ad6ab2b1538e1036d0cfe7e75b9ed9`, 112.32s). Five skip reasons verbatim: `Tessera encodes need CUDA`. Receipt `49a11ee95fc88145da43d1e5c34e47b87355e4015e16265fe4294191afd903e3`.
- Producer API author independently reviewed the adapter; no API misuse found.

Detailed append-only evidence: `docs/measurements/tessera-checkpoint-identity-2026-09-04.md`. `docs/ARCHITECTURE.md` changes with the gate. No full-master baseline or uncoordinated full suite was run.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -5507,7 +5507,9 @@ Two properties make the numbers comparable with the rest of the menu:
 * **Rendering identity (§P8).** The tensor that is priced *is* the decoded wire:
   `_encode_and_render` returns `read_unit_artifact(unit.blob)`, and the blob is
   stored beside the render in the cache. Not two code paths that agree — one
-  object.
+  object. This proves the cached wire is the priced wire, not that an export
+  or serving runtime reused it; those paths owe their own exact-byte and
+  served receipts. The packed bridge remains tracked by PrismaQuant #183.
 
 Rows are written in the codebase's own currency: `output_mse` and **no**
 `predicted_dloss` field, because in this tree `output_mse` is a raw MSE while
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,22 @@
 As of: 2026-09-04 · `codex/pq186-campaign-identity`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-04, `codex/pq186-campaign-identity`) for **identity-bound
+Tessera campaign resume** (§4.10; PrismaQuant #186). The existing cost journal
+binds the current selected weights, scoring rows even with Hessians off, actual
+Hessians, corpus/token draw, menus/scope, explicit campaign settings, resolved
+encoder configuration and both producer/package source seals. Its original
+`--checkpoint` file path is the manifest; sibling `.parts` shards hold durable
+checksummed per-unit anchors and the producer's own cached-byte receipts.
+Every resumed anchor must belong to the current unit/menu, agree with the
+producer's Hessian applicability and verify its exact wire bytes against a
+fresh producer input identity before entering the cost payload. Legacy unbound
+JSON, changed inputs, extra/missing receipts and missing/changed wire refuse;
+old anchors never acquire today's identity. This consumes the pending producer
+`tessera.cached_unit` API and does not change the release pin, admit packed
+campaigns, or establish served/GPU qualification. Gate:
+`tests/test_tessera_campaign_resume.py`.
+
 Re-stamped (2026-09-04, `codex/pq186-campaign-identity`) for the shared
 checkpoint journal's **explicit manifest pathname** (§4.10; PrismaQuant #186).
 `cost_stage_checkpoint.prepare_journal` can retain a file-oriented caller's
@@ -5444,6 +5460,19 @@ purpose. Each surface records its anchors, encode seconds and which of the two
 stopped it.
 
 **Cost is an anchor campaign, not an enumeration** (`prismaquant/tessera_campaign.py`).
+
+Resume is an identity check, not a name match. The JSON checkpoint manifest
+uses `cost_stage_checkpoint` and stores checksummed per-unit anchor shards in
+its sibling `.parts` directory. It binds actual selected weights, scoring rows,
+Hessians and calibration draw, resolved menus/scope/settings and the producer
+and PrismaQuant source seals. Each priced wire has a `tessera.cached_unit`
+receipt, verified against freshly derived producer inputs before its saved
+cost is admitted. Changed or unverifiable inputs, old identity-free anchor
+lists, absent/corrupt wire and receipt coverage mismatch refuse without
+recomputing or relabelling the old result. Output/cache paths and the wall-clock
+interruption limit are not numerical inputs; moving them is allowed only if
+the same manifest, unit journal and verified priced wire remain available.
+The producer API is required; an older producer without it refuses.
 
 > **Premise correction.** This work was briefed on the belief that Tessera's
 > "embedded rate axis" would let one deep encode per unit yield exact decodes of
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,18 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-04 · `codex/pq183-packed-capture`. Stamps
+As of: 2026-09-04 · `codex/pq186-campaign-identity`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-04, `codex/pq186-campaign-identity`) for the shared
+checkpoint journal's **explicit manifest pathname** (§4.10; PrismaQuant #186).
+`cost_stage_checkpoint.prepare_journal` can retain a file-oriented caller's
+manifest path while storing per-unit shards in its checkpoint directory. It
+uses the existing manifest/unit schemas, exact identity comparison, checksums,
+and durable writes; existing directory-oriented callers are unchanged. This
+adapter enables the campaign's identity-bound resume without changing its
+`--checkpoint` filename/default or introducing another checkpoint grammar.
+The campaign's intake/wire verification is a separate change, not supplied
+merely by this optional argument.
 
 Re-stamped (2026-09-04, `codex/pq183-packed-campaign`) for **profile-pinned
 campaign exclusions** (§4.10). The dense target walk now consults the existing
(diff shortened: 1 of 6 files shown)


RobTand:
## Brought up to date with #187 / `main` and verified — worker `pq-campaign`

**Tip:** `e0f4895d` · base `codex/pq183-packed-capture` at `0b01bc62` (which contains `origin/main` at `65cfa6d9`: PRs #188, #189, #190, and #196). `gh pr view 191 --json mergeable` → `MERGEABLE`; `git merge-tree origin/main HEAD` is conflict-free, so retargeting to `main` after #187 lands needs no further merge. No authored commit rewritten.

### Merge resolution — `d5e174a6` (merge of `codex/pq183-packed-capture` incl. PR #196)

Conflicts: the resume intake in `prismaquant/tessera_campaign.py` (this PR's identity-bound intake vs #196's `_require_resumable_anchor` gate) and the `docs/ARCHITECTURE.md` header. Resolved on **one rule with one home: the identity-bound resume subsumes #196's "refuse resuming W4A4 anchors priced under another activation contract" guard.**

- `_campaign_checkpoint_identity` now binds the served A-side contract as the scoring input it is: top-level `input_global_scale_policy` (env-resolved via `PRISMAQUANT_NVFP4_INPUT_GSCALE_FP8_RANGE`, so nothing else bound it) and per-unit `units.<unit>.input_global_scale` (from `_static_input_scales`; a function of every calibration row, not derivable from the bounded score rows or the Hessian). A checkpoint priced under another calibration or policy refuses at the journal, by field, before any row is read — probed: `checkpoint identity mismatch at units.model.layers.0.proj.input_global_scale: stored=2.0 current=1.0` and `... at input_global_scale_policy: stored='legacy_6_over_calibration_amax.v1' current='full_e4m3_range_448x6_over_calibration_amax.v1'`.
- `_checkpoint_anchor_identity(..., static_scales=)` asks the per-row half of the same rule — `_require_resumable_anchor` — right beside the `hessian_applied` check and *before* the producer's `encoding_input_identity` / wire receipt. `main()` no longer calls `_require_resumable_anchor` on its own; its one caller is `_checkpoint_anchor_identity`. The fresh-measure path goes through the same function, so a freshly priced row's receipt binds the scale it was priced at.
- `_require_resumable_anchor` keeps #196's two refusals (`pre-served-contract` row without a scale; `mixes two activation calibrations`) and additionally refuses a dynamic-route row that carries a static scale (no producer of this campaign stamps one): a row's activation identity must be exactly what `_measure_anchor` stamps for its rung.
- `docs/ARCHITECTURE.md`: `As of: 2026-09-05 · codex/pq186-campaign-identity`; every stamp from all sides kept, newest first, this PR's on top; one new stamp records the resolution.
- `e0f4895d` re-merges `codex/pq183-packed-capture` after it took `main` at `65cfa6d9` — clean.

Tests added for the composed rule: `tests/test_tessera_campaign_resume.py` gains `input_scale` and `scale_policy` cases of `test_main_refuses_changed_encoding_or_scoring_inputs` (fixture now returns #196's 4-tuple with `max_abs`, and pins the default policy so the policy case is a real change); #196's `test_resume_refuses_w4a4_anchors_priced_under_another_contract` now also asserts the same refusals from `_checkpoint_anchor_identity` and the stray-scale dynamic-route case.

### Verification (sparklina, CPU)

Set = PR's own tests + every test file importing `tessera_campaign` / `cost_stage_checkpoint` that the #187 set did not already cover + the two doc gates:

```
tests/test_tessera_campaign_resume.py tests/test_cost_stage_checkpoint_manifest.py
tests/test_tessera_campaign.py tests/test_tessera_campaign_packed.py tests/test_tessera_campaign_pins.py
tests/test_tessera_scope_endpoints.py tests/test_architecture_doc.py tests/test_docs_staleness.py
tests/test_cb_anchored_shard_merge.py tests/test_artifact_collection_legacy.py
tests/test_activation_hook_scope_gates.py tests/test_anchored_cost.py tests/test_union_production_cache.py
tests/test_streamed_cost_checkpoints.py tests/test_prismasnap_validation.py tests/test_stock_anchored_cost.py
tests/test_tessera_export_lane.py tests/test_cost_currency.py
```
(`CUDA_VISIBLE_DEVICES="" python -m pytest -p no:cacheprovider -q --no-header -n 4 <set>`)

1. **Dev pin `1221d2a`** (`PYTHONPATH=/home/rob/tessera-pin-1221d2a/src`): **17 failed, 235 passed, 5 skipped**. All 17 are `tests/test_tessera_campaign_resume.py` main-entry tests (`test_main_resumes_identical_cost_and_wire_without_reencoding`, `test_main_refuses_changed_hessian_values_under_same_draw`, `test_main_refuses_changed_encoding_or_scoring_inputs[×12]`, `test_main_refuses_missing_or_changed_priced_wire[×3]`) and all fail identically: `ImportError: cannot import name 'cached_unit' from 'tessera'` → `RuntimeError: Tessera campaign checkpoint identity requires the producer's cached_unit input/byte receipt API; refusing unbound resume`. `tessera.cached_unit` does not exist at the pin; this PR's own measurement doc says the dependency "must land before campaign intake is usable".
2. **Pin + producer API** (`PYTHONPATH=/home/rob/pq-audit-caches/pq-campaign/tessera-pin-plus-cached-unit/src`, a verification-only copy of the `1221d2a` tree with `src/tessera/cached_unit.py` from `/home/rob/tessera-release/src` dropped in — every name it imports exists at the pin; nothing under git was touched): **252 passed, 5 skipped, 0 failed**. This is the only way the 17 tests above could be verified; the full release tree cannot run them because `main`'s `lane_eligibility` refuses the release's `tessera.lane-eligibility.v8` table (`LaneEligibilityError`, the #192 re-pin's business), so the fresh priced fixture never encodes there.

Full suite not run, per campaign brief. Inherited `main` baseline failures are outside this set.

### For the merge decision

- With this PR, `main()` calls `_campaign_checkpoint_identity` for **every** campaign, fresh or resumed, so at the current dev pin the Tessera campaign cannot price anything: it refuses by name for lack of `tessera.cached_unit`. That is the fail-closed design the PR states, but it means this PR is only *usable* once the pin moves to a Tessera that ships `cached_unit` (release `6343137`+, i.e. after the #192 re-pin). CI at the pin will show the 17 failures above.
- `test_main_refuses_unbound_checkpoint_before_accepting_anchors` passes at the pin for the wrong reason (its `checkpoint.*identity` regex also matches the missing-API refusal); it passes for the right reason on the overlay.

Issue #165, opened 2026-09-03
Tessera's priced A-side is never compared to the contract cells' activation_contract, so producer and runtime can disagree at no gate


RobTand:
Evidence first. `prismaquant/tessera_menu.py:464` prices the route from the producer table, not the contract:

```
    return RouteAdmission(
        format_name=name,
        payload_family=family.name,
        activation_contract=route.contract,
```

where `route` is `tessera_serving_route(family, recipe, rung)` -- the layout table in `prismaquant/tessera_formats.py:1111-1148` (e.g. E2M1 returns `contract="w4a4-nvfp4-e2m1-group16-ue4m3"`). The same producer value feeds the byte breakdown (`prismaquant/tessera_footprint.py:425`: `"activation_contract": route.contract`), the campaign rows (`prismaquant/tessera_campaign.py:434`), and the synthesized `FormatSpec` A-side (`prismaquant/tessera_render.py:662-669,716-722`).

The runtime publishes its own value per cell: `prismaquant/tessera_runtime_contract.py:839` parses `activation_contract` off every `lane_eligibility.cells[]` row, and it travels in the pinned answer (`:505`). Those vocabularies do not even match (`w4a4-nvfp4-e2m1-group16-ue4m3` vs `e2m1_group16_ue4m3_static`), and I checked every `tessera_serving_route` consumer (`tessera_footprint.py:383`, `tessera_render.py:662`, `tessera_menu.py:402`) -- none compares the priced contract to any cell's.

The only cross-check is prose containment: `tests/test_tessera_lane_admission.py:382` asserts `value in rationale`, i.e. a test gate reads the lane spec's `rationale` string. That makes the rationale load-bearing: the sanctioned fix for a stale test is editing prose, which is the exact edit the field's own `if_this_changes` note forbids. Provenance nothing consumes is a confession log; a test that consumes prose is a gate that reads prose.

Why it matters: producer table and runtime cells can disagree about a family's executed A-side -- the E4M3-over-block-plane kernel-lane boundary in `tessera_formats.py:1129-1149` is the live edge -- while the allocator prices one contract and the serve executes another. That is the NVFP4_CB 2026-08-17 currency-error class (`prismaquant/lane_spec.py:106-113`, an 87 GB misallocation), with no gate to catch it. Distinct from `#127` (an unread currency *label*) and `#136` (an unread `decoder`): this is an unread *value* on the active pricing path.

[...]

PR #196, opened 2026-09-05
Tessera lane: carry the priced inputs from allocator to executor to cost-table identity (#193, #194, #195)

### LINKED ISSUE #193: Tessera pipeline omits the Hessian and static activation scales required by the priced allocation
Inspected canonical PrismaQuant origin/main `10dd23957d6cf9ad784f8eac59fb914f66f6d89e` and Tessera `ea72d23cbf26a453d0d740afac537fb6d6b8afbc`. The assigned PrismaQuant working tree was older (`1eb88c4eea4498226ca869533aadec39af9349d3`); every source anchor below was rechecked using `git show origin/main:<path>`. No repository source was edited. This was an explicitly read-only audit, so the finding is filed rather than fixed. No GPU, served run, full suite, or PrismaBuild execution was performed.

The Tessera export arm forwards only the source/output paths, `--plan-json`, and `--device` to the external exporter (`prismaquant/run-pipeline.sh:2513-2517`). It passes neither `--hessian` nor `--input-scales`, and does not consume the wire blobs saved by the campaign. These omissions have different outcomes:

1. **Wrong rendering for H-aware allocations:** `tessera_campaign.py:670` defaults `--hessian` to `require`; `_measure_anchor` scores the decoded H-aware blob and stores it beside the cached render. The production path also explicitly requires a calibration identity/Hessian (`tessera_render.py:1166`, `:1259`). Tessera `experiments/export_tessera_serving.py:1123` defaults `--hessian` to None and at `:1196-1217` constructs an ActivationSource only when the argument is present. Consequently a pure E4M3/BF16 allocation can be re-encoded weights-only after being priced/rendered H-aware. The allocation names family/rung but does not reproduce those omitted encoder inputs.
2. **E2M1/NVFP4 export unavailable:** Tessera exporter `:1107-1110` declares `--input-scales` required for any NVFP4 route. Its `:1559-1568` loads scales only from that argument and refuses if any NVFP4 module exists without them. Thus any E2M1-containing allocation fails before output rather than producing the selected artifact. No pipeline argument supplies these scales.

The currently PENDING release gate and runtime schema compatibility can prevent reaching this arm; this is not a claim that today's guarded main ships such an artifact. It is a reachable caller-contract defect once those independent admission prerequisites are satisfied. Draft pin PR #192 opens that route but its run-pipeline changes do not supply these inputs. PR #188 changes prepriced-cost intake, and PRs #187/#191 change capture/resume; none patches this exporter argument list. #183 covers packed-MoE projection/capture, whereas this defect is already present for ordinary dense units.

Evidence/test scope: static read of both sides of the invocation, plus existing `tests/test_tessera_campaign.py` H-aware-vs-weights-only regression (`test_the_production_render_is_h_aware_and_the_lever_is_not`, canonical main lines 761-791) showing these are intentionally different renders. Those tests were trusted at face value, not rerun. No numerical effect on a real model was measured.

Acceptance: introduce a closed export input contract carrying the exact calibration/Hessian identity and resolved encoder settings that priced each selected unit, or use the producer's checked cached-wire assembly interface where supported. Supply and bind the actual static NVFP4 input scales, including fused-group unification. Refuse a missing or incompatible input before encoding. Add a CPU CLI-argument regression for both H-aware dense E4M3 and E2M1 paths, and a bounded producer check that exported decoded weights equal the priced render under the same inputs. Keep ship/pin gates closed until their existing conditions pass.

Disposition: filed for a separate implementation agent. The user explicitly requested an audit and issues, with no fixes; no implementation changes were made for this finding.

### LINKED ISSUE #194: Tessera E2M1 costs use an FP32-scale activation quantizer instead of the served static UE4M3 contract
Inspected canonical PrismaQuant origin/main `10dd23957d6cf9ad784f8eac59fb914f66f6d89e` and Tessera `ea72d23cbf26a453d0d740afac537fb6d6b8afbc`. The assigned PrismaQuant working tree was older (`1eb88c4eea4498226ca869533aadec39af9349d3`); every source anchor below was rechecked using `git show origin/main:<path>`. No repository source was edited. This was an explicitly read-only audit, so the finding is filed rather than fixed. No GPU, served run, full suite, or PrismaBuild execution was performed.

A Tessera E2M1 rung's FormatSpec borrows the registered NVFP4 activation callback (`tessera_render.py:765-766`; `tessera_formats.py:1364`). That callback is `_make_rtn("fp4_e2m1", 16)` (`format_registry.py:821`). Its grouped scale is `max_abs/cmax` in FP32 (`:610`), never rounded through UE4M3, and nearest-code midpoint handling chooses the lower code (`:620-623`). The campaign scores with that callback and explicitly passes `activation_max_abs=None` (`tessera_campaign.py:266-275`); no artifact input-global-scale value reaches this function.

Tessera serves a different operation. `src/tessera/serving/nvfp4_route.py:209-210` reads the artifact's `trellis_input_global_scale` and passes it to `native_fp4_quant`; `serving/native_ops.py:79-98` calls vLLM's compiled `scaled_fp4_quant`, a static-global-scale NVFP4 operation with UE4M3 block scales. The exporter obtains the value from actual supplied scale tensors and takes the fused maximum (`experiments/export_tessera_serving.py:1766-1773`). PrismaQuant already owns the corresponding CPU oracle in `nvfp4_activation_contract.py:1110` (scale snap at `:1142-1144`).

Concrete CPU counterexample, with `PRISMAQUANT_DISABLE_RTN_COMPILE=1` and torch `2.10.0+cpu`:

```python
import torch
from prismaquant.format_registry import get_format, nvfp4_activation_qdq_served
x = torch.tensor([[1e-3] * 16], dtype=torch.bfloat16)
q = get_format("NVFP4").activation_quantize_dequantize(x)
s = nvfp4_activation_qdq_served(x, 1.0)
print(float(q[0, 0]), float(s[0, 0]), torch.equal(q, s))
# 0.00099945068359375 0.0 False
```

This probe ran in the assigned checkout; AST comparisons confirmed `_make_rtn` and `nvfp4_activation_qdq_served` are identical to canonical main. G=1 is a legal explicitly supplied scale, not an assumption that every production artifact uses G=1. The CPU result establishes the mismatch of the existing oracles; it is not a packed-byte or native GPU verification.

Impact: the allocator reads a cost measured against a different activation tensor than the runtime executes; both scale-underflow and midpoint cases are lost. This is a residual of closed #165, not the same fixed bug: its new gate at `tessera_menu.py:423` compares only `(act_bits, act_group_size)`, so both different quantizers pass as (4,16). The tests proving that a callback is applied do not prove that it is the served callback. Open PRs #187-#192 do not replace this callback or give the campaign the static scale.

Acceptance: reuse the existing served NVFP4 oracle with the exact per-module/fused input-global-scale values that export writes. Carry that scale policy/value identity into cost and cache provenance; missing scale inputs must refuse rather than use dynamic FP32 scales. Add CPU underflow, E4M3 scale-snap and E2M1 midpoint tests, plus a bounded native comparison on the supported runtime. Reassess affected Tessera E2M1 costs on the same calibration contract; no quality/performance change is claimed here.

Disposition: filed for a separate implementation agent. The user explicitly requested an audit and issues, with no fixes; no implementation changes were made for this finding.

### LINKED ISSUE #195: Tessera cost-table identity guard ignores the required Hessian identity triple
Inspected canonical PrismaQuant origin/main `10dd23957d6cf9ad784f8eac59fb914f66f6d89e` and Tessera `ea72d23cbf26a453d0d740afac537fb6d6b8afbc`. The assigned PrismaQuant working tree was older (`1eb88c4eea4498226ca869533aadec39af9349d3`); every source anchor below was rechecked using `git show origin/main:<path>`. No repository source was edited. This was an explicitly read-only audit, so the finding is filed rather than fixed. No GPU, served run, full suite, or PrismaBuild execution was performed.

`assert_uniform_hessian_identity` compares only the legacy fields `(supplied, text_sha, token_count, kwarg)` at `tessera_menu.py:1512-1514`. It neither validates nor compares Tessera's required `text_sha256`, `fit_ids_sha256`, and `fit_tokens`. A dict lacking all legacy fields is still counted as stamped, and any number of distinct modern identities collapse to `(True, None, None, None)`. The allocator runs this guard at `allocator.py:2340` and publishes only the returned legacy projection at `:5458`, dropping the modern triple again.

The current campaign already writes the modern triple into every row (`tessera_campaign.py:348-358`); `tessera_hessian.HESSIAN_IDENTITY_FIELDS` derives the required field set from Tessera rather than maintaining a second roster. Thus the omission is in the consumer's actual comparison, not an absence of a defined identity contract.

CPU counterexample (actual guard; AST verified identical between the assigned checkout and canonical main):

```python
from prismaquant.tessera_menu import assert_uniform_hessian_identity
a = {"supplied": True, "text_sha256": "a" * 64,
     "fit_ids_sha256": "b" * 64, "fit_tokens": 128}
b = {**a, "fit_ids_sha256": "c" * 64}
print(assert_uniform_hessian_identity({"m.up": {
    "TESSERA_E4M3_K1_R1024": {"hessian_identity": a},
    "TESSERA_E4M3_K1_R1280": {"hessian_identity": b},
}}))
# supplied=True, text_sha=None, token_count=None, kwarg=None,
# stamped_rows=2, unstamped_rows=0
```

Expected: conflicting required identities refuse, or a modern-only identity is explicitly classified as unsupported. Actual: the guard reports both rows uniformly stamped and lets allocation consume the scores. This also permits inconsistent modern fields to survive when legacy aliases happen to agree. No claim is made that the normal single campaign currently omits its legacy aliases; the failure concerns validation/merging at the public cost-table boundary.

Dedup: #186 and ready stacked PR #191 address checkpoint admission and per-anchor source/wire binding. That fix is needed, but PR #191 changes `tessera_campaign.py` and the checkpoint journal; it does not change this separate table-uniformity guard. PR #188 strengthens currency/schema checks but does not change it either. Existing tests around `test_one_cost_table_carries_one_hessian_identity` compare the legacy fields only; payload tests check that modern fields are written without proving the consumer reads them.

Acceptance: define one canonical versioned identity using the shared required fields; compare all of it, validate any carried legacy aliases for consistency, and distinguish legacy/unstamped/unsupported rows explicitly. Preserve the canonical identity in allocation provenance so production/export can consume it. Add tests for two modern-only identities differing one required field, equal legacy aliases with conflicting modern fields, missing required fields, and a valid matching table. Keep the per-rung `applied` distinction separate as the existing mixed-family semantics require.

Disposition: filed for a separate implementation agent. The user explicitly requested an audit and issues, with no fixes; no implementation changes were made for this finding.

RobTand:
Audit burn-down, Tessera lane: the three issues are one story — the pricing inputs the campaign measures under (Hessian identity, static activation-scale contract) were not carried from allocator to executor to cost-table identity. The stack fixes them coherently around one shared object: the canonical Hessian identity triple (#195) is what the export gate binds a capture against (#193), and the static scales the served contract executes (#194) are the values the campaign now writes for the exporter (#193). One commit per issue, plus one follow-on guard.

Test environments: local CPU (`/home/rob/venvs/pq-cu130`, tessera importable from a clone of the reference checkout at the dev pin `1221d2a4`) and sparky (`/home/rob/audit-sync/prismaquant`, `/home/rob/dq-runs/venvs/prismaquant-cu130`, CUDA_VISIBLE_DEVICES=""). No full suite was run; targeted files only.

---

## #195 — cost-table identity guard ignores the required Hessian identity triple (P1)

Fixes #195

**Defect.** `tessera_menu.assert_uniform_hessian_identity` keyed rows only on the legacy `(supplied, text_sha, token_count, kwarg)` projection. Every current campaign row carries Tessera's required triple (`text_sha256` / `fit_tokens` / `fit_ids_sha256`), so any number of distinct modern identities collapsed to one key: a cost table merged from two Hessian draws passed as uniformly stamped and the DP traded rows priced on different bytes. The allocator then published only the legacy projection into `layer_config.json`.

**Fix.** The guard reads the required roster from `tessera_hessian.HESSIAN_IDENTITY_FIELDS` (derived from `tessera.export.HESSIAN_IDENTITY`, never retyped), keys modern rows on the triple AND the legacy aliases (agreeing aliases cannot launder disagreeing triples), keeps pre-triple rows as their own `legacy` class that never merges with a modern row, refuses a partial triple by name, and returns the canonical triple (plus `identity_schema`, `legacy_rows`) — which the allocator's existing stamp carries verbatim into `__prismaquant__.tessera_hessian`, where #193's gate reads it.

**Pre-fix failure line** (`test_the_guard_compares_the_required_hessian_identity_triple`):
```
with pytest.raises(ValueError, match="mixes Hessian identities"):
E   Failed: DID NOT RAISE <class 'ValueError'>
```
(The issue's counterexample returned `supplied=True text_sha=None token_count=None kwarg=None, stamped_rows=2` for two identities differing in `fit_ids_sha256` — reproduced verbatim.)

**Tests run.** `tests/test_tessera_campaign.py` (new: triple comparison, alias laundering, partial triple, legacy/modern split, canonical return), `tests/test_tessera_menu.py`, `tests/test_tessera_scope_endpoints.py`.

---

## #194 — E2M1 costs used an FP32-scale activation quantizer instead of the served static UE4M3 contract (P1)

Fixes #194

**Defect.** `tessera_campaign._measure_anchor` scored every E2M1 anchor with the NVFP4 FormatSpec callback — `_make_rtn("fp4_e2m1", 16)`, a dynamic per-group FP32-scale RTN, explicitly with `activation_max_abs=None` — while Tessera serving reads the artifact's `trellis_input_global_scale` and calls vLLM's compiled `scaled_fp4_quant` (static global scale, UE4M3-stored block scales). Scale-underflow and midpoint rounding diverge; the post-#165 agreement gate compares only `(act_bits, act_group_size)`, so both quantizers pass as (4, 16).

**Fix.** `_measure_anchor` resolves the rung's serving route; any NVFP4-materialising route is scored through the owned served oracle `format_registry.nvfp4_activation_qdq_served` at the unit's calibrated static `input_global_scale` — calibrated from max|x| over **every** calibration row of the campaign's own forward passes, unified across fused siblings (`unify_fused_sibling_max_abs`) and converted by the owned policy helpers (`resolve_input_global_scale_policy` / `input_global_scale_from_max_abs`). A missing scale refuses (`ActivationScaleContractError`, re-raised out of the anchor loop like a Hessian refusal) rather than falling back to the dynamic quantizer. The scale value rides on every W4A4 row (measured and interpolated) and the policy + per-unit values ride in payload provenance.

**Pre-fix failure line** (served-contract binding against the old `_measure_anchor`, encode stubbed to identity):
```
AssertionError: E2M1 anchor priced under the dynamic FP32-scale RTN: dloss=0 != served-contract score 0.000346137
```

**Re-pricing implication.** Every existing Tessera cost table's **E2M1 rows are stale** (E4M3/BF16 rows unaffected — their routes are dynamic/identity at serve). Measured on a real 32×256 `TESSERA_E2M1_K2_R896` render with 128×256 activations under the resolved default policy: Gaussian activations served/dynamic = **1.011** (+1.1%), wide-dynamic-range rows **0.966** (−3.4%), and the underflow corner is unbounded (dynamic ≈ 0 vs served 3.46e-4 on a uniform 1e-3 input at G=1). Direction is case-dependent, so affected tables must be **re-measured on the same calibration contract**, not adjusted. No recorded campaign result files were modified.

**Residual (noted, not changed).** `tessera_runtime_contract.cell_activation_projection` still projects cell contracts to `(act_bits, act_group_size)` only; with pricing now on the served static oracle the priced and executed A-sides agree, but a third projection axis (static-vs-dynamic + scale dtype) would let the agreement gate refuse a future drift by name. Left for the contract's owner.

**Tests run.** `tests/test_tessera_campaign.py` (new: served-contract binding, refusal-before-encode, dynamic-route exemption, row scale propagation, fused-sibling sharing, whole-stream amax), `tests/test_tessera_scope_endpoints.py`, `tests/test_cost_currency.py`.

---

## #193 — Tessera pipeline omits the Hessian and static activation scales required by the priced allocation (P0)

Fixes #193

**Defect.** `run-pipeline.sh`'s Tessera arm forwarded only `--plan-json` and `--device` to `experiments/export_tessera_serving.py`. The exporter builds an `ActivationSource` only when `--hessian` is present (and raises nothing without one), so an H-aware allocation — the campaign's default — was re-encoded weights-only: different bytes at the same format names, silently. And the exporter hard-requires `--input-scales` for any NVFP4 route, so an E2M1-containing allocation died inside the exporter after the plan translation instead of producing the selected artifact. Nothing in the pipeline produced either input.

**Fix (three pieces, one contract).**
1. *Requirements are declared:* the allocation already carries `__prismaquant__.tessera_hessian` (now with the canonical triple, #195); the W4A4 scale requirement is structural, derived from the selected units' routes.
2. *Execution fails closed:* `tessera_export_lane` gains `--hessian`/`--input-scales` and a fifth gate, `require_priced_export_inputs`, run from `preflight` before the scope gate and before the plan translation. It refuses: an H-aware allocation without a capture; a capture whose identity triple does not match the allocation's stamp; a held-out capture; a stray capture on a weights-only allocation (the same drift, other direction); an allocation that declares no pricing state (ambiguous → closed, AGENTS.md P2); and a W4A4 selection whose scales file does not cover every selected unit (header check, no tensor load).
3. *Inputs are supplied:* the campaign writes `hessian_capture.pt` (the exact un-normalised per-unit XᵀX it priced with, `ActivationSource.from_capture`-shaped, `hessian_role: "fit"`, JSON provenance sidecar) and `input_scales.safetensors` (the #194 scales as `<unit>.input_global_scale` F32 scalars) beside its cache, before the anchor loop; `run-pipeline.sh` threads `TESSERA_HESSIAN`/`TESSERA_INPUT_SCALES` through one array to both the preflight and the exporter invocation.

`docs/ARCHITECTURE.md` re-stamped and `CHANGELOG.md` updated in the stack per the maintenance contract; `tests/test_docs_staleness.py` + `tests/test_architecture_doc.py` pass.

**Pre-fix failure lines** (`tests/test_tessera_priced_export_inputs.py`):
```
E   assert '"${TESSERA_PRICED_INPUT_ARGS[@]}"' in 'experiments/export_tessera_serving.py" \\n    "$MODEL_PATH" "${WORK_DIR}/exported" \\n    --plan-json "$TESSERA_PLAN" \\n    --device "$EXPORT_DEVICE" \\n    '
E   AttributeError: module 'prismaquant.tessera_export_lane' has no attribute 'require_priced_export_inputs'
```

**Tests run.** `tests/test_tessera_priced_export_inputs.py` (new, 17 tests), `tests/test_tessera_export_scope.py`, `tests/test_tessera_export_lane.py`, `tests/test_tessera_plan_input_binding.py` (real shell-arm execution).

---

## Follow-on guard (separate commit, refs #194)

Campaign `--checkpoint` resume merged saved anchors into this run's table with no currency check on the activation axis: a pre-#194 checkpoint's W4A4 rows (no `input_global_scale`) or rows from another calibration would silently rebuild the mixed table. `_require_resumable_anchor` refuses both by name; dynamic-route rows resume untouched.

## #175 (optional item) — commented, not changed

The merged #176 workflow is correct at `master`: it resolves the exact dev pin and installs Tessera `--no-deps` in both CPU jobs. The five most recent CI runs fail only at `Check out pinned Tessera` with `remote: Repository not found` — the expected refusal while Tessera is private. Remaining options (wait for publication per the agreed plan / interim fine-grained PAT / never soft-fail) need Rob's decision, so they are laid out in a comment on #175 instead of a change: https://github.com/RobTand/prismaquant/issues/175#issuecomment-5550002762

## Side findings (not changed here)

- The Tessera exporter's fused-scale join takes the **max** over member `input_global_scale`s (`export_tessera_serving.py:1771`) while PrismaQuant's own contract joins conservatively on the **min** scale / max amax (`unify_fused_sibling_input_global_scales`). For scales computed from one module input (as the campaign does) the two coincide; for scales from independently calibrated stock exports they diverge, and the max-scale join clips the widest sibling. Tessera-repo territory — flagged for that lane.
- `cell_activation_projection`'s two-axis projection (see #194 residual above).
- **Sparky's shared test env is ahead of the Tessera dev pin.** `/home/rob/dq-runs/venvs/prismaquant-cu130` imports tessera editable from `/home/rob/tessera` (`d52e1ac`), which packages a `tessera.lane-eligibility.v8` contract that master's `lane_eligibility` reader (v3–v5) refuses — **42 tessera-lane tests fail there at baseline master `10dd2395`**, before any change of mine. `/home/rob/tessera-audit` (`2b0b933`) has the same skew locally. All authoritative runs here used a clone of the tessera checkout at the dev pin `1221d2a4`, where baseline and the branch are both clean. Whoever coordinates the pin bump may want this on their radar.

## Test evidence

- Local, tessera at the dev pin `1221d2a4` (CPU, `CUDA_VISIBLE_DEVICES=""`): broad sweep of all 15 touched-module test files (`test_tessera_campaign/menu/menu_context/menu_real_table/scope_endpoints/export_lane/export_scope/priced_export_inputs/plan_input_binding/encoder_recipe.py`, `test_cost_currency.py`, `test_run_pipeline_defaults.py`, `test_run_pipeline_gates_execute.py`, `test_doc_claims_driver_truth.py`, `test_lane_gate_recording.py`): **291 passed, 5 skipped (CUDA-marked), 0 failed**; docs gates (`test_docs_staleness.py`, `test_architecture_doc.py`): **19 passed**. (The final rebase only added CHANGELOG text on top of the tested tree — `git diff` between them is CHANGELOG-only.)
- Sparky (`/home/rob/audit-sync/prismaquant`, shared env with the v8 skew above): branch **46 failed / 150 passed / 5 skipped** vs baseline master **42 failed / 125 passed / 5 skipped** on the same file set. The failure sets differ only by 4 of this PR's new #194 tests, each failing with the identical pre-existing `LaneEligibilityError: ... got 'tessera.lane-eligibility.v8'` — environment skew, not regression; all 4 pass on the pinned tessera.

https://claude.ai/code/session_01RnYbLUvQb2xCH3o99VsfFo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@
 
 ### Fixed
 
+- **Tessera W4A4 anchors are priced under the served static UE4M3 activation
+  contract** (#194; `tessera_campaign._measure_anchor`). The campaign scored
+  every E2M1 anchor with NVFP4's registry callback — a dynamic per-group
+  FP32-scale RTN — while Tessera's plugin executes vLLM's static-global-scale
+  `scaled_fp4_quant` against the artifact's `trellis_input_global_scale`,
+  with UE4M3-stored block scales; underflow and midpoint cases diverge (a
+  1e-3 block flushes to exact 0 under the served contract at G=1 and survives
+  under the FP32 scale). W4A4 anchors are now scored through the owned served
+  oracle (`nvfp4_activation_qdq_served`) at the unit's calibrated,
+  fused-sibling-unified static scale — calibrated over every calibration row
+  from the campaign's own forward passes under the resolved NVFP4 policy — a
+  missing scale refuses (`ActivationScaleContractError`) instead of falling
+  back to the dynamic quantiser, and the scale value/policy identity travels
+  on every W4A4 row and in the payload provenance. **Re-pricing:** existing
+  Tessera cost tables' E2M1 rows were priced under the dynamic A side and are
+  stale; re-measure on the same calibration contract (E4M3/BF16 rows are
+  unaffected).
 - **The Tessera cost-table identity guard compares the required Hessian
   identity triple** (#195; `tessera_menu.assert_uniform_hessian_identity`).
   The guard keyed rows only on the legacy `(supplied, text_sha, token_count,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@
 
 ### Fixed
 
+- **The Tessera export arm fails closed on missing priced inputs, and the
+  campaign supplies them** (#193; `run-pipeline.sh`,
+  `tessera_export_lane.require_priced_export_inputs`,
+  `tessera_campaign.write_export_inputs`). The arm forwarded only
+  `--plan-json`/`--device` to Tessera's exporter: an H-aware allocation (the
+  campaign's default) was re-encoded weights-only — the exporter builds an
+  `ActivationSource` only when `--hessian` is present and raises nothing
+  without one — and any E2M1 selection died inside the exporter, which
+  hard-requires `--input-scales` for NVFP4 routes. The campaign now writes
+  `hessian_capture.pt` (the exact un-normalised per-unit XᵀX plus the
+  identity triple, with a JSON provenance sidecar) and
+  `input_scales.safetensors` beside its cache; the driver threads
+  `TESSERA_HESSIAN`/`TESSERA_INPUT_SCALES` to both the lane preflight and the
+  exporter; and a fifth lane gate refuses, before the plan translation, an
+  H-aware allocation without its identity-matched capture, a weights-only
+  allocation handed a stray one, an undeclared allocation, and a W4A4
+  selection whose scales file does not cover every selected unit.
 - **Tessera W4A4 anchors are priced under the served static UE4M3 activation
   contract** (#194; `tessera_campaign._measure_anchor`). The campaign scored
   every E2M1 anchor with NVFP4's registry callback — a dynamic per-group
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### Fixed
 
+- **Campaign resume refuses W4A4 anchors priced under another activation
+  contract** (follow-on to #194; `tessera_campaign._require_resumable_anchor`).
+  A pre-#194 checkpoint's W4A4 rows carry no `input_global_scale` (dynamic
+  pricing) and rows from another calibration carry a different one; merging
+  either silently rebuilds the mixed-currency table on the activation axis
+  that the Hessian identity guard refuses on its own.
 - **The Tessera export arm fails closed on missing priced inputs, and the
   campaign supplies them** (#193; `run-pipeline.sh`,
   `tessera_export_lane.require_priced_export_inputs`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@
 
 ### Fixed
 
+- **The Tessera cost-table identity guard compares the required Hessian
+  identity triple** (#195; `tessera_menu.assert_uniform_hessian_identity`).
+  The guard keyed rows only on the legacy `(supplied, text_sha, token_count,
+  kwarg)` projection, so any number of distinct modern identities —
+  `text_sha256` / `fit_tokens` / `fit_ids_sha256`, read from
+  `tessera_hessian.HESSIAN_IDENTITY_FIELDS`, which every current campaign row
+  carries — collapsed to one key and a table merged from two Hessian draws
+  allocated as one. Modern rows now key on the triple AND the legacy aliases,
+  pre-triple rows are their own `legacy` class that never merges with a
+  modern row, a partial triple refuses by name, and the returned stamp
+  carries the canonical triple into `__prismaquant__.tessera_hessian` so
+  export can bind a capture against the allocation.
 - **The group knapsack's per-member rung licence is read from the Tessera
   contract, not from its own docstring** (#132, RobTand/tessera#37;
   `allocator_candidates.tessera_group_composites`,
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,31 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-04 · `codex/pq-tessera-v5-endpoints`. Stamps
+As of: 2026-09-05 · `claude/pq-tessera-audit`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-05, `claude/pq-tessera-audit`) for **the Tessera
+priced-inputs contract** (§5, §9.4; RobTand/prismaquant#193/#194/#195, one
+story). The campaign now writes the export leg's inputs beside its cache —
+`hessian_capture.pt` (the exact un-normalised per-unit XᵀX plus the identity
+triple, the shape `ActivationSource.from_capture` loads, with a JSON
+provenance sidecar) and `input_scales.safetensors` (one static
+`input_global_scale` per unit, fused-sibling unified) — and the driver's
+Tessera arm threads `TESSERA_HESSIAN` / `TESSERA_INPUT_SCALES` through one
+array to both the lane preflight and Tessera's exporter, which previously
+received only `--plan-json`/`--device` and so re-encoded H-aware allocations
+weights-only and could not export an E2M1 selection at all. A fifth lane gate
+(`tessera_export_lane.require_priced_export_inputs`) fails closed before the
+plan translation: an H-aware allocation without its identity-matched capture,
+a weights-only allocation handed a stray one, an undeclared allocation, and a
+W4A4 selection whose scales file does not cover every selected unit all
+refuse by name. Feeding it, the campaign prices every W4A4 anchor under the
+**served** static UE4M3 contract (`nvfp4_activation_qdq_served` at the unit's
+calibrated scale; a missing scale refuses rather than falling back to the
+dynamic FP32-scale RTN — existing E2M1 cost rows are stale and need
+re-measuring), and `assert_uniform_hessian_identity` compares the full
+required identity triple, so two draws can no longer collapse to one legacy
+key. Tests: `test_tessera_priced_export_inputs.py` plus the campaign-file
+additions.
 
 Re-stamped (2026-09-04, `codex/pq-tessera-v5-endpoints`) for **raw scoped
 census handoff instructions** (§9.4). The driver prints the producer's actual
@@ -5256,8 +5280,15 @@ the same format name, and the seam refuses rather than downgrade silently:
   `fit_ids_sha256`, source, split role, seed) beside it.
 * **The provenance has a consumer.** `tessera_menu.assert_uniform_hessian_
   identity` refuses a cost table whose Tessera rows carry two different Hessian
-  identities, and the allocator calls it at load and stamps the result into
-  `__prismaquant__.tessera_hessian`. Unstamped rows are counted as
+  identities — compared on the **full required triple** (`text_sha256` /
+  `fit_tokens` / `fit_ids_sha256`, read from
+  `tessera_hessian.HESSIAN_IDENTITY_FIELDS`) beside the legacy aliases, so
+  agreeing aliases cannot launder disagreeing triples (#195); pre-triple rows
+  are their own `legacy` class that never merges with a modern row, and a
+  partial triple refuses by name — and the allocator calls it at load and
+  stamps the result, canonical triple included, into
+  `__prismaquant__.tessera_hessian`, which is the identity the export gate
+  binds a `--hessian` capture against. Unstamped rows are counted as
   *unstamped*, not assumed to match. Provenance that nothing consumes is a
   confession log; this one is a gate.
 
@@ -5425,11 +5456,17 @@ predicts worst — and gated on a measured LOO error, not a taste constant.
 Two properties make the numbers comparable with the rest of the menu:
 
 * **Priced as served.** The render is scored under the route's own activation
-  contract — NVFP4's W4A4 quantiser on E2M1 rungs, per-token FP8 on E4M3 — taken
-  *by reference* from the serving format's registry row, so the A leg is not a
-  second implementation of it. The same weight rate therefore costs differently on
+  contract — the **served static UE4M3** W4A4 quantiser on E2M1 rungs
+  (`nvfp4_activation_qdq_served` at the unit's calibrated, fused-unified
+  `input_global_scale`, the same operation `scaled_fp4_quant` executes against
+  `trellis_input_global_scale`; a missing scale refuses,
+  `ActivationScaleContractError`, rather than falling back to the registry's
+  dynamic FP32-scale RTN — #194), per-token dynamic FP8 on E4M3, taken *by
+  reference* from the owned oracles, so the A leg is not a second
+  implementation of either. The same weight rate therefore costs differently on
   the two routes, which is asserted directly by
-  `test_the_same_weight_rate_costs_differently_on_the_two_routes`.
+  `test_the_same_weight_rate_costs_differently_on_the_two_routes`, and each
+  W4A4 row carries the `input_global_scale` it was scored under.
 * **Rendering identity (§P8).** The tensor that is priced *is* the decoded wire:
   `_encode_and_render` returns `read_unit_artifact(unit.blob)`, and the blob is
   stored beside the render in the cache. Not two code paths that agree — one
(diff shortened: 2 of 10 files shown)


RobTand:
**Ruling on the fused-scale join divergence flagged here (tessera max-join vs this PR's min-scale/max-amax): this PR's side is correct; tessera's exporter was wrong and is fixed in RobTand/tessera#275.**

Ground truth, established from the served semantics rather than either join: tessera's route hands `trellis_input_global_scale` unmodified to vLLM's `scaled_fp4_quant` (`src/tessera/serving/nvfp4_route.py:221-222` -> `native_ops.py:97`), and the pinned vLLM build stores each group-16 block scale as `e4m3(block_amax/6 * g)` clamped at 448 (`nvfp4_emulation_utils.py:174-183`) — exactly what this PR's `nvfp4_activation_qdq_served` oracle mirrors at `nvfp4_activation_contract.py:1142`. So `input_global_scale` is **capacity/amax, inverse in the activation range**: too large clips the peaks silently (the stored UE4M3 scale saturates), too small only spends precision. The fused module's one GEMM quantises one input tensor spanning every member's range, so the correct join is min scale = max amax — this PR's rule, and also the rule tessera's own calibrator already stated (`experiments/rotation_activation_stats.py:108,125`). Receipt through this repo's oracle (members calibrated at amax 8 and 4, full-E4M3 policy, served value 8.0): min-join serves 8.0 back (max err 0.656, ordinary rounding); max-join serves **4.0** (max err 4.0 — clipped to the smaller member's calibrated range).

Two notes:

1. **The join is degenerate on every sanctioned donor.** vLLM fuses q/k/v and gate/up because they read the same input tensor, so jointly calibrated members (this repo's `unify_fused_sibling_*`, the jointly-optimised MSE grid, tessera's calibrator) carry identical scales and max == min. Only independently calibrated member scales expose the divergence — tessera's exporter had copied vLLM's stock `.max()` reduction (`compressed_tensors_w4a4_nvfp4.py:126`), which in vLLM is a warning-guarded no-op over already-unified checkpoints, not a join rule.

2. **Materially divergent member scales are now refused at export, not joined.** Beyond one bf16 ULP (`torch.finfo(torch.bfloat16).eps` = 2^-7 — the route casts the A tensor to bf16 before the quantiser sees it, so scales from ONE calibration agree to within one step of that lattice), the spread is two calibrations (mixed draws/policies — the hazard this repo already refuses as "mixed" on its side), and tessera's exporter refuses by name where the bytes are decided rather than serving a distribution nobody measured. This repo's silent min-join in `unify_fused_sibling_input_global_scales` remains correct **where it runs**: at calibration time, max over member max-abs on the same draw IS the joint measurement, completed, not two calibrations laundered.

No prismaquant change needed.

Issue #202, opened 2026-09-05
[P2] 22 call sites still swallow the dead-vendored-override refusal one level up (#201 follow-up)


RobTand:
## Finding

#201 fixed the owner: `model_profiles/registry._resolve` now runs
`_refuse_dead_vendored_override` **outside** its per-candidate
`except Exception: continue` and raises `DeadVendoredOverrideError(RuntimeError)`,
so a checkpoint whose vendored-modelling override is known dead refuses by name
instead of being handed `DefaultProfile`. `detect_profile_with_warning` re-raises
that class rather than logging a fallback, and three gate/validator call sites
were fixed in the same PR (`artifact_completeness._detect_profile_quietly`,
`incremental_probe._detect_profile_for_shards`,
`validate_native_export._resolve_validation_target_profile`).

**22 further call sites across 11 modules still swallow it one level up.** Each
wraps `detect_profile` / `profile_from_config` / `profile_from_model` in a broad
`except Exception` and continues with a substituted profile, so at those call
sites the refusal is still converted back into a silent wrong answer — the run
proceeds against UPSTREAM modelling code under a `None` / `DefaultProfile` /
skipped branch, which is the #19 failure mode the gate exists to prevent.

Census (AST scan for a call to one of the three detection entrypoints
lexically inside a `try` whose handler catches `Exception`/`BaseException`;
`prismaquant/` only, `archive/` excluded), with what each handler substitutes:

| file:line | call | handler substitutes |
|---|---|---|
| `prismaquant/aqua_activation_cost.py:661` | `detect_profile` | logs, falls back to generic name matching |
| `prismaquant/build_rtn_cache.py:115` | `detect_profile` | `keep_composite = False` |
| `prismaquant/build_rtn_cache.py:287` | `profile_from_model` | `profile = None` |
| `prismaquant/build_rtn_cache.py:500` | `detect_profile` | `profile = None` |
| `prismaquant/incremental_measure_quant_cost.py:585` | `profile_from_model` | `profile = None` |
| `prismaquant/incremental_measure_quant_cost.py:1032` | `profile_from_model` | `profile = None` |
| `prismaquant/layer_streaming.py:1015` | `profile_from_model` | `return None` |
| `prismaquant/layer_streaming.py:1058` | `profile_from_model` | `return None` |
| `prismaquant/layer_streaming.py:1140` | `profile_from_model` | `return 0` |
[...]

Issue #197, opened 2026-09-05
test_selected_export_units_resolve_their_own_structure passes alone and fails in the full suite: model-profile registry state leaks between tests


RobTand:
## Finding

`tests/test_tessera_export_scope.py::test_selected_export_units_resolve_their_own_structure`
**passes in isolation and fails in a full-suite run** on `main` (`8a77b0ad`), CPU-only, tessera
at the declared dev pin `1221d2a`. It is therefore an order-dependent test-isolation defect, not
a defect in the code the test covers.

Full-suite run (`-n 4`, worker `gw4`), `/home/rob/prismaquant-baseline`:

```
FAILED tests/test_tessera_export_scope.py::test_selected_export_units_resolve_their_own_structure
prismaquant.tessera_export_lane.TesseraExportLaneError: cannot attest selected Tessera export
scope: model.layers.0.mlp.experts.0.down_proj: explicit Tessera scope needs a declared model profile
prismaquant/tessera_export_lane.py:476
```

Same commit, same interpreter, same tessera, the test alone:

```
cd <pristine origin/main worktree>
CUDA_VISIBLE_DEVICES="" PYTHONPATH=/home/rob/tessera-pin-1221d2a/src \
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider -q --no-header \
  tests/test_tessera_export_scope.py::test_selected_export_units_resolve_their_own_structure
1 passed
```

The refusal comes from `prismaquant/tessera_serving_scope.py:73`
(`unit_structure_from_profile`), which raises when `profile is None or
profile.structure_spec() is None`. Under `detect_profile` the same checkpoint fixture resolves a
profile *with* a structure spec when the test runs alone, so something earlier in the worker's
session changes what `prismaquant/model_profiles/registry.py` resolves or what its resolved
profile carries. That module holds process-global mutable state — `_REGISTERED` (a list
`register_profile` inserts into), `_REGISTRY_GENERATION`, `_DETECTION_ORDER_CACHE`, and
`profile._declare_model_path(model_path)` mutating the resolved profile object at
`registry.py:163` — which is the shape of defect that produces exactly this symptom.

## What breaks, and under what inputs

Only the test suite, as far as I can show: a green targeted run does not predict a green
full-suite run for anything that goes through `detect_profile`. I did **not** establish that any
production path is affected, and I did not identify the polluting test — the obvious candidate,
[...]

Issue #19, opened 2026-07-29
register_qwen3() silently stops overriding on newer transformers (broken by 5.13.1; good on 5.6.0)


RobTand:
Found by the new CI (#18) on a clean environment.

## What happens

`prismaquant.vendored.register_qwen3()` routes Qwen3 causal-LM loads to the vendored model with:

```python
AutoModelForCausalLM.register(Qwen3Config, Qwen3ForCausalLM, exist_ok=True)
```

On newer transformers that registration no longer wins over the natively-supported `model_type`, so `AutoModelForCausalLM.from_config(Qwen3Config(...))` returns the **upstream** class.

## Version boundary — measured, not assumed

| transformers | result |
|---|---|
| 5.6.0 (reference box) | override applies — passes |
| 5.13.1 | override does NOT apply — fails |
| 5.14.1 | override does NOT apply — fails |

I first guessed this was 5.14-specific and pinned CI to `<5.14`. That pin installed 5.13.1 and **still failed**, which is how the boundary came to be measured rather than assumed. It sits somewhere in 5.7…5.13.1 and is unbisected.

## Why it matters more than one red test

The failure is **silent**. `register_qwen3()` returns cleanly and sets `_QWEN3_REGISTERED = True`; nothing raises. A probe on an affected version quietly runs upstream Qwen3 modelling code instead of the vendored path — on the architecture family behind most shipped artifacts — and the only symptom is whatever numerical difference that introduces.

**The silent no-op is the real defect, independent of which version is supported.**

## Current state

CI does not pin transformers (pinning solved nothing and a pin to 5.6.0 would rot). `tests/test_vendored_qwen3.py::test_qwen3_auto_model_uses_vendored_causal_lm` is `xfail(strict=False)` above 5.7 with this issue referenced, so it still passes on the reference box and reports as expected-failure on modern transformers.

## What a fix needs

1. Bisect the exact boundary.
2. Determine how modern transformers resolves `from_config` for a registered override on a natively-supported `model_type`.
3. Make `register_qwen3()` either genuinely override, **or fail loudly** when it cannot.
4. Remove the xfail.

`register_deepseek_v4()` registers a *new* `model_type` rather than overriding an existing one, so it is probably unaffected — unverified.

RobTand:
Fixed in `29f3cff`, shipped in **v0.3.0**. Both halves: the override genuinely applies on modern transformers, *and* a registration that cannot take effect now stops the run.

[...]

PR #192, opened 2026-09-05
Pin Tessera by exact commit and contract digest, and read lane-eligibility v6

RobTand:
## Pin Tessera by exact commit and contract digest; read lane-eligibility v6 → v9

Rob: *"can we just pin prismaquant to latest version of tessera? then we won't have to keep cutting releases."*

"Latest" is read here as **an exact commit plus the packaged contract's raw digest** — never a floating ref. Everything the tag-based design bought is kept; only the tag requirement is dropped.

### The pin, as merged

| | value |
|---|---|
| `TESSERA_DEV_PIN_COMMIT` / `TESSERA_SERVING_RUNTIME_PINNED_COMMIT` / pin JSON `commit` | `8ed1d9a78b3f0c7036dcbe14d7df3a89f398812a` |
| `TESSERA_DEV_PIN_CONTRACT_SHA256` / pin JSON `contract_sha256` | `719daa02da1564b56a141ca2702ae29d4fda553460978efbb6510ddcd1824927` |
| contract | `contract_version` 22, lane schema `tessera.lane-eligibility.v9` |
| version | `0.1.0`, `version_is_release: false` (no tag names the commit) |
| pin schema | `prismaquant.tessera_serving_runtime_pin.v2` |

`8ed1d9a` is Tessera master's tip (the merge of its #332), verified rather than taken on report — fetched into a scratch repo and hashed at the tip, never a working tree:

```bash
TS=$(mktemp -d) && git -C "$TS" init -q
git -C "$TS" fetch -q https://github.com/RobTand/tessera master
git -C "$TS" rev-parse FETCH_HEAD
git -C "$TS" cat-file -p FETCH_HEAD:src/tessera/serving/runtime_contract.json | sha256sum
```

### 1. What immutability rests on

| | before | after |
|---|---|---|
| pin schema | `…pin.v1` | `…pin.v2` |
| identity | `PENDING_TESSERA_RELEASE_COMMIT` / `…_VERSION`, flipped by a tag | an exact `commit` + `contract_sha256` |
| what a gate enforces | `version_is_release == true` | the **installed** `tessera/serving/runtime_contract.json` hashes to `contract_sha256` |
| `version_is_release` | **the admission gate** | **advisory** — recorded, published in provenance, gates nothing |

[...]

PR #203, opened 2026-09-05
Let the dead-vendored-override refusal escape profile detection (#201)

### LINKED ISSUE #201: [P2] _resolve swallows _refuse_dead_vendored_override: a dead vendored override yields DefaultProfile silently instead of refusing by name
## Finding

`prismaquant/model_profiles/registry.py::_resolve` calls `_refuse_dead_vendored_override(model_type)` **inside** the per-profile loop body that is wrapped in `try: ... except Exception: continue` (`registry.py:~300-323`). The refusal raises `RuntimeError` — "vendored modelling override ... did not take effect, so a load here would silently run UPSTREAM modelling code" — and the enclosing `except Exception: continue` swallows it. Detection then falls through to `DefaultProfile(architectures=archs)` (`registry.py:~324`).

So the gate that exists to refuse "the silent-wrong-answer this project refuses to ship" (its own docstring, `registry.py:98-110`, issue #19) has no effect: a checkpoint whose vendored override is dead is handed `DefaultProfile` silently instead of a refusal by name. Downstream, some paths refuse later for an unrelated-sounding reason (`tessera_serving_scope.unit_structure_from_profile`: "explicit Tessera scope needs a declared model profile", because `DefaultProfile.structure_spec()` is None), and other paths — anything that accepts `DefaultProfile` — proceed against upstream modelling code with no signal at all.

Found while closing #197 (PR #200): the order-dependent test failure was this swallow with a leaked `OVERRIDE_ERRORS["qwen3"]` entry; #200 fixed the test-state leak at the owner (an autouse restore in `tests/conftest.py`) and did not touch this production path. Verified in one interpreter: insert `OVERRIDE_ERRORS["qwen3"] = "x"` → `detect_profile(qwen3 config)` answers `DefaultProfile`, no exception.

## Severity

P2 (a refusal with no teeth; the wrong answer is a silently substituted profile, which changes what gets probed/quantised — not a served byte directly, but principle: refuse where the decision is made, by name, no silent fallback).

## Fix

Let the dead-override refusal escape `_resolve`: raise it outside the swallow (compute the match inside the `try`, run `_refuse_dead_vendored_override` after the `try/except`, or re-raise that specific error class from the `except`), with its own error type so callers can tell "no profile matched" from "a profile matched and its vendored path is dead". Test: a config that resolves to `Qwen3Profile` with `OVERRIDE_ERRORS["qwen3"]` set (via `monkeypatch.setitem`) must raise by name, not return `DefaultProfile`; shown failing first. ARCHITECTURE.md re-stamped in the same commit (profile detection behaviour changes). Check `_resolve`'s callers for anyone relying on the fall-through.

Filed by the coordinator from the #197 worker's report; dispatched for fixing immediately.

RobTand:
Fixes #201. Refs #202.

## The defect

`model_profiles/registry._resolve` called `_refuse_dead_vendored_override(model_type)` from
**inside** its per-candidate `try: ... except Exception: continue`. So the gate added by #19 —
the one whose own docstring calls a wrong modelling path "exactly the silent-wrong-answer this
project refuses to ship" — had no teeth. Its `RuntimeError` was caught by the walk's own
tolerance, the profile that matched was demoted, and detection fell through to
`DefaultProfile(architectures=archs)` with no exception anywhere.

Reproduced in one interpreter on `468e7534` before touching anything:

```
clean:              Qwen3Profile
with dead override: DefaultProfile name= default
```

Downstream, some paths then refuse for an unrelated-sounding reason
(`tessera_serving_scope.unit_structure_from_profile`: "explicit Tessera scope needs a declared
model profile", because `DefaultProfile.structure_spec()` is None) and everything that accepts
`DefaultProfile` proceeds against UPSTREAM modelling code with no signal at all.

## The fix (commit 1 — the owner)

The loop was doing two jobs and they are now separated.

**Choosing a candidate stays tolerant.** A profile whose `matches()` raises, or that cannot be
instantiated, is not a match and the walk continues — that is what the `except Exception` is for,
and a new test pins that it still works.

**The refusal is a statement about the candidate that WON**, so it runs after the `except`, not
inside it. `_resolve` now computes the match inside the `try` and runs the gate outside it.

It raises **`DeadVendoredOverrideError(RuntimeError)`**, its own class, because the two failures a
caller can see out of detection are not interchangeable:

| what happened | answer | legitimate? |
|---|---|---|
| nothing matched this checkpoint | `DefaultProfile` | yes, for a not-yet-registered architecture — and guarded downstream by `allocator.validate_default_profile_format_menu` |
| a profile matched, its vendored path is dead | refusal, by name | never an answer — the load would run upstream modelling code under a profile that promises the vendored copy |

Subclassing `RuntimeError` keeps the gate's original contract, so existing `except RuntimeError`
handlers and the four pre-existing gate tests are unchanged.

**One rule, one home.** The refusal itself did not move; only where it is allowed to propagate.
`register_vendored_modeling()`'s own swallow is untouched — its documented reason still holds, and
this gate is precisely how the failure it hides is recovered.

## The same swallow one level up

An AST scan (a call to `detect_profile` / `profile_from_config` / `profile_from_model`
lexically inside a `try` whose handler catches `Exception`/`BaseException`) found **26 further
sites across 15 modules** under `prismaquant/`.

**Fixed in this PR — 4:**

- `registry.detect_profile_with_warning` (commit 1, same file as the owner). It exists so
  production entrypoints keep running on a not-yet-registered model with a logged fallback; a
  printed warning is not a refusal, and the warning it printed here was
  `"architecture unregistered or config unreadable"`, which is false.
- `artifact_completeness._detect_profile_quietly` (commit 2) — a completeness *gate* that
  answered `None`, running name-blind over an artifact the build could name.
- `incremental_probe._detect_profile_for_shards` (commit 2) — answered `DefaultProfile`, then
  sharded and probed against upstream modelling code.
- `validate_native_export._resolve_validation_target_profile` (commit 2) — a *validator* that
  answered `None` and picked its serving target from `default=`.

All four document their tolerance as "an architecture this build does not know", which is a
different statement from "a known architecture on a dead path". Each keeps every other tolerance;
only this one class is re-raised.

**Already correct — 1:** `sample_parallel_probe.py:568` re-raises as
`SampleParallelProbeError(...) from exc`. Refuses by name, keeps the cause. No change.

**Filed, not fixed — 22 across 11 modules:** censused with `file:line` and what each handler
substitutes in **#202**. Several are genuinely optional-hint paths (`layer_streaming.py:1140`
returns a count, `streaming_model.py:112` returns a bool) where `None` is the right answer for an
*absent* profile and the wrong one for a *dead* one — so each needs reading against its own
contract, not one mechanical edit. 22 two-line edits across 11 modules is several times the diff
of the fix they hang off; AGENTS.md principle 14, "large enough to swamp the diff of the task you
came for". The issue records that the filer could have fixed it and why they did not.

## Test, shown failing first

`tests/test_vendored_override_gate.py` is the gate's own file — its module docstring is already
"the profile registry must refuse a model_type whose vendored override died". Eight cases added
there, six of them the new rule (the other two pin the tolerance that must survive).

Commit 1, pre-fix — **3 failed, 6 passed**:

```
FAILED test_detect_profile_refuses_a_dead_override_instead_of_defaulting
  - Failed: DID NOT RAISE DeadVendoredOverrideError
FAILED test_profile_from_config_refuses_a_dead_override
FAILED test_detect_profile_with_warning_does_not_swallow_the_refusal
```

post-fix **9 passed**. Commit 2, pre-fix — **3 failed, 9 passed** (the three call sites, same
`DID NOT RAISE` line); post-fix **12 passed**.

The call-site tests live in the registry gate's file rather than each module's, because the rule
being pinned is the registry's — where this refusal is allowed to stop — and it has one home.

## Docs

`docs/ARCHITECTURE.md` re-stamped in **commit 1**, the same commit as the behaviour change
(principle 12): a new dated stamp at the head, and §8.1's dead-override paragraph rewritten —
it described a gate that did not fire. It now records where the refusal is allowed to stop, why
the class exists, the four fixed call sites, and the 22 filed in #202.

`tests/conftest.py`'s autouse-fixture docstring (from #200) explained #197 partly *in terms of*
this swallow — "that raise happens inside `_resolve`'s `except Exception: continue`, so a stray
entry does not fail loudly". True when written, false after this commit, so corrected in place
(AGENTS.md 14: prose is fixed on sight, never filed). It now records that a leaked
`OVERRIDE_ERRORS` entry fails the next test loudly and by name — which is what made #197 hard to
see in the first place.

## Blast radius

None on a healthy tree: `OVERRIDE_ERRORS` is empty at import on the reference box and only ever
gains an entry when a vendored override actually fails to take effect. Verified:

```
OVERRIDE_ERRORS at import: {}
```

What changes is that when it *is* non-empty, detection refuses by name instead of silently
substituting a profile.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,32 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-05 · `codex/pq186-campaign-identity`. Stamps
+As of: 2026-09-05 · `claude/pq-201-dead-override`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-05, `claude/pq-201-dead-override`) for **a
+dead-vendored-override refusal that actually refuses** (§8.1; PrismaQuant
+#201). The gate added by #19 was calling `_refuse_dead_vendored_override`
+*inside* `_resolve`'s per-candidate `except Exception: continue`, so its raise
+never left detection: it demoted the profile that matched and `detect_profile`
+answered `DefaultProfile` — the silent substitution the gate's own docstring
+says it refuses. The gate now runs after that `except`, on the candidate that
+won, and raises `DeadVendoredOverrideError` (a `RuntimeError` subclass, so the
+gate's original contract is unchanged) so a caller can tell "nothing matched
+this checkpoint" — a legitimate `DefaultProfile`, guarded downstream by
+`allocator.validate_default_profile_format_menu` — from "a profile matched and
+its vendored modelling path is dead", which has no legitimate answer.
+`detect_profile_with_warning` re-raises it rather than logging a fallback: its
+tolerance is for an unknown architecture, and its warning would otherwise have
+claimed "architecture unregistered or config unreadable", which is false. The
+candidate walk keeps its tolerance for a profile that cannot be asked or built,
+and the `register_vendored_modeling()` swallow is unchanged. No default, format
+menu, serving lane or exported byte moves; what changes is that a checkpoint on
+a dead vendored path now refuses by name instead of being probed under
+`DefaultProfile`. Three call sites that were the same swallow one level up —
+the completeness gate, the incremental probe's shard detection and the native
+export validator — re-raise it too; the other 22 broad-`except` swallows around
+detection, across 11 modules, are censused with `file:line` in #202 rather than
+swept here. Gate: `tests/test_vendored_override_gate.py`.
 
 Re-stamped (2026-09-05, `codex/pq186-campaign-identity`) for **one resume
 identity, including the served A-side contract** (§4.10; merge of the
@@ -8634,6 +8659,38 @@ way. The fix registers a PrismaQuant-owned subclass of the native config through
 `AutoConfig.register` (public API, no internals patched), engages only when the direct route is
 verified dead, and verifies every registration by a config-only resolution before setting the
 "done" flag. Boundary measured, not assumed: healthy through 5.12.1, broken from 5.13.0.
+
+**Where that refusal is allowed to stop** (#201, 2026-09-05). For its first year the gate had no
+teeth: `_resolve` called it from *inside* the per-candidate `try: ... except Exception: continue`,
+so the raise was caught by the walk's own tolerance, the matched profile was demoted, and
+detection fell through to `DefaultProfile(architectures=archs)` with no exception anywhere — the
+same silent-wrong-answer shape #19 was filed about, one level up. The two things that loop does
+are now separated. *Choosing* a candidate stays tolerant: a profile whose `matches()` raises, or
+that cannot be instantiated, is not a match and the walk continues. *The refusal is a statement
+about the candidate that won*, so `_refuse_dead_vendored_override(model_type)` runs after the
+`except`, not inside it, and raises `DeadVendoredOverrideError(RuntimeError)`. Its own class
+because the two failures a caller can see are not interchangeable: "nothing matched" is a
+legitimate `DefaultProfile` for a not-yet-registered architecture, while "a profile matched and
+its vendored path is dead" means the load would run upstream modelling code under a profile that
+promises the vendored copy. `detect_profile_with_warning` — the entrypoint that exists to tolerate
+anything — re-raises this one class and logs the rest, because a printed warning is not a refusal
+and the warning it printed ("architecture unregistered or config unreadable") was not even true.
+This is also why #197's leaked `OVERRIDE_ERRORS["qwen3"]` was invisible: a stray entry silently
+re-routed every later qwen3 detection in the process instead of failing. Gate:
+`tests/test_vendored_override_gate.py`. Three call sites were the same swallow one level up and
+re-raise the class for the same reason: `artifact_completeness._detect_profile_quietly` (a
+completeness *gate* that answered `None`), `incremental_probe._detect_profile_for_shards` (which
+answered `DefaultProfile`) and `validate_native_export._resolve_validation_target_profile` (a
+*validator* that answered `None`). Each documents tolerance for an architecture this build does
+not know, which is a different statement from a known architecture on a dead path.
+**Not fixed here:** 22 further broad-`except` swallows around
+`detect_profile`/`profile_from_config`/`profile_from_model`, across 11 modules, can still convert
+this refusal into `None`, `DefaultProfile`, `False` or a skipped branch at their own call sites.
+Several of those are genuinely optional hints where `None` is the right answer for an *absent*
+profile and the wrong one for a *dead* one, so they need reading one at a time against their own
+contract; censused with `file:line` in #202 rather than swept inside this diff. One site already
+has the right shape and is excluded: `sample_parallel_probe.py:568` re-raises as
+`SampleParallelProbeError(...) from exc`.
 
 The `DefaultProfile` fallback is *guarded, not silent*: `allocator.py:1550-1554` calls
 `validate_default_profile_format_menu(...)` (`:961-988`), which refuses a multi-format menu
--- a/prismaquant/artifact_completeness.py
+++ b/prismaquant/artifact_completeness.py
@@ -235,12 +235,21 @@ def _checkpoint_spellings(unit: str, profile) -> set[str]:
 
 def _detect_profile_quietly(artifact_dir: Path):
     """The artifact's own profile, or None. Never fatal: the gate must still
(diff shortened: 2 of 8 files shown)


RobTand:
## Commands and counts

Box `sparklina`, CPU only, tessera at the pin. Interpreter is the one `run-tests.sh` uses
(`/home/rob/dq-runs/venvs/prismaquant-cu130/bin/python`); `pytest-xdist` is not installed in it,
so everything below is **serial** (the brief allowed serial or `-n 2`).

Common prefix for every run:

```
cd /home/rob/prismaquant-audit-wt/pq-201
export CUDA_VISIBLE_DEVICES="" TMPDIR="$PWD/.claude/tmp" PYTHONDONTWRITEBYTECODE=1
export OMP_NUM_THREADS=1 MKL_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1
export PYTHONPATH=.:/home/rob/tessera-pin-1221d2a/src
PY=/home/rob/dq-runs/venvs/prismaquant-cu130/bin/python
```

### 1. The defect, reproduced on `468e7534` before any edit

```
$PY - <<'PY'
import json, tempfile; from pathlib import Path
import prismaquant.vendored as vendored
from prismaquant.model_profiles import detect_profile
d = Path(tempfile.mkdtemp())
(d/"config.json").write_text(json.dumps({"model_type":"qwen3","architectures":["Qwen3ForCausalLM"]}))
print("clean:", type(detect_profile(str(d))).__name__)
vendored.OVERRIDE_ERRORS["qwen3"] = "synthetic: override no-op'd"
print("with dead override:", type(detect_profile(str(d))).__name__)
PY
```

```
clean:              Qwen3Profile
with dead override: DefaultProfile      <- no exception; the gate never fired
```

### 2. The gate's own file, failing first then passing

```
$PY -m pytest -q --no-header -p no:cacheprovider tests/test_vendored_override_gate.py
```

| | result |
|---|---|
| commit 1 pre-fix | **3 failed, 6 passed** — `Failed: DID NOT RAISE DeadVendoredOverrideError` |
| commit 1 post-fix | **9 passed** |
| commit 2 pre-fix | **3 failed, 9 passed** — same line, the three call sites |
| commit 2 post-fix | **12 passed** |

### 3. Tests for the three modules commit 2 touches

```
$PY -m pytest -q --no-header -p no:cacheprovider \
  tests/test_artifact_completeness_namespaces.py tests/test_artifact_completeness_routed_per_role.py \
  tests/test_incremental_probe.py tests/test_incremental_probe_kv_shared_guard.py \
  tests/test_validate_native_export.py tests/test_probe_shard_reuse.py \
  tests/test_sample_parallel_probe.py tests/test_probe_linear_exclude.py
```

**84 passed, 1 skipped in 11.19s.**

### 4. Everything that imports `model_profiles`, plus both doc tests

`registry.py` is imported very widely, so the sweep is every test file naming `model_profiles`
(62 files) rather than a hand-picked list, plus the two doc gates:

```
$PY -m pytest -q --no-header -p no:cacheprovider \
  $(grep -ln "model_profiles" tests/*.py) \
```

**17 failed, 1880 passed, 82 skipped, 1 xfailed, 163 subtests passed in 104.68s.**

The 17 are the known-inherited `test_tessera_campaign_resume.py` main-entry tests the brief names
— `tessera.cached_unit` does not exist at the pin, fixed by the re-pin (#192), untouched by this
PR:

```
>   from tessera import cached_unit
E   ImportError: cannot import name 'cached_unit' from 'tessera'
      (/home/rob/tessera-pin-1221d2a/src/tessera/__init__.py)
E   RuntimeError: Tessera campaign checkpoint identity requires the producer's
      cached_unit input/byte receipt API; refusing unbound resume
```

Count and file match the inherited set exactly. `tests/test_architecture_doc.py` and
`tests/test_docs_staleness.py` both pass.

### 5. The call-site census behind #202

The 26 sites are found mechanically, not by grep — an AST walk for a call to `detect_profile` /
`profile_from_config` / `profile_from_model` lexically inside a `try` whose handler catches
`Exception` or `BaseException`, over `prismaquant/` with `archive/` excluded. Script and full
table are in #202. Of the 26: **4 fixed here**, **1 already correct**
(`sample_parallel_probe.py:568`, which re-raises by name with the cause), **22 filed**.

### Blast radius

`OVERRIDE_ERRORS` is `{}` at import on this box, so nothing changes on a healthy tree — the entry
only ever appears when a vendored override actually fails to take effect. What changes is that
detection then refuses by name instead of quietly answering `DefaultProfile`.

Issue #205, opened 2026-09-05
[P1] Use the served Tessera activation contract in assignment-KL hooks and cache scoring


RobTand:
The campaign now prices Tessera E2M1 activations through the served static-scale oracle, but the production activation hooks and cache render scorer still evaluate the dynamic FP32-scale quantizer. The explicit served-emulation switch does not fix Tessera because its condition checks the format name for literal `NVFP4`.

Initially audited PrismaQuant `65cfa6d9ee53cec7880c5320fd44c395b9406a3b` and rechecked current main `468e7534bad23e4c02454f44eb5517e9a6a39f4d` after merges #187/#191/#200; all source anchors below still apply at `468e7534bad23e4c02454f44eb5517e9a6a39f4d`. The affected hook, scorer, spec and KL functions are AST-identical across this delta; the production-cache change only adds an optional packed capture row consumer below those functions. The original numerical proof remains on the frozen `65cfa6d9ee53cec7880c5320fd44c395b9406a3b` snapshot, with direct source-equivalence evidence in `proofs/prismaquant-delta-function-comparison.json`. This is a residual integration path of #194 after #196 fixed `_measure_anchor`; the new campaign callback/refusal and full-stream calibration tests hold. It is not a request to reopen the original campaign implementation or to change the established stock-NVFP4 screen default.

Current path:

- `tessera_render.py:765-766` still borrows the generic NVFP4 FormatSpec activation callback for a Tessera E2M1 spec. That callback is the dynamic FP32-scale RTN that #194 distinguished from served UE4M3 scales.
- `PerturbedActivationCache` retains that Tessera spec from the assignment (`perturbed_x_cache.py:360-372`), then its real module pre-hook passes it to `_activation_qdq` (`:1236-1238`). The served branch at `:125-136` requires `canonical_format_name(act_spec.name) == "NVFP4"`. A `TESSERA_E2M1_K2_R896` spec therefore falls through to the dynamic callback at `:138`, even with `PRISMAQUANT_NVFP4_ACT_EMULATE_SERVED_SCALES=1` and a known calibration maximum. `kl_measurement.measure_assignment_kl` constructs these hooks at `kl_measurement.py:1094-1102`.
[...]

PR #206, opened 2026-09-05
tests: bound the cluster-campaign waits by producer liveness, not the clock

### LINKED ISSUE #199: test_resume_after_coordinator_death_recovers_owned_helper_receipt: an 8s wall-clock deadline mostly spent on interpreter start, so it flakes with box load
## Finding

`tests/test_cluster_campaign_runner.py::test_resume_after_coordinator_death_recovers_owned_helper_receipt`
asserts against a **fixed 8.0 s wall-clock deadline** whose budget is dominated by a cold
Python interpreter start, so the margin shrinks with box load rather than with anything
the test is about. It was seen failing under `-n 4` in the #197 baseline run; it passes
serially.

The deadline:

```
tests/test_cluster_campaign_runner.py:599    deadline = time.monotonic() + 8.0
tests/test_cluster_campaign_runner.py:615    assert observed_running, (
tests/test_cluster_campaign_runner.py:616        "coordinator never recorded slow-stage as running with a pid; "
```

Between `Popen` (line 585) and the first state the assertion can accept, the test has to
pay for a cold `python -m prismaquant.cluster_campaign` — interpreter start plus the
module's import graph — before the coordinator can write its first state file.

## What I measured

Box: sparklina, CPU-only (`CUDA_VISIBLE_DEVICES=""`), `/home/rob/venvs/pq-release/bin/python`,
`PYTHONPATH=/home/rob/tessera-pin-1221d2a/src`, worktree on `origin/main` @ `5ab603e6`.

Instrumented the test's own poll loop with the deadline lifted to 60 s, so the number is
the real latency rather than a pass/fail:

| box load (1 min) | `observed_running` reached at | margin against the 8.0 s deadline |
|---|---|---|
| ~10 (idle-ish), serial | 2.96 / 3.04 / 3.23 s | 4.8 – 5.0 s |
| ~10, four concurrent copies (mimics `-n 4`) | 3.09 – 3.40 s | 4.6 – 4.9 s |
| **105 – 113** (suite running under other agents) | **5.28 / 5.31 / 5.56 / 5.87 s** | **2.1 – 2.7 s** |

Attribution — a bare import of the coordinator module, nothing else:

```
import prismaquant.cluster_campaign:  3.36 / 3.58 / 3.69 s   (load ~10)
import prismaquant.cluster_campaign:  4.71 / 4.84 / 4.98 s   (load ~108)
```

So 59–62% of the entire 8.0 s budget is spent importing, before the coordinator does any
campaign work at all, and the consumed fraction rose from ~38% to ~73% as load went from
~10 to ~108. The deadline is sized against an unloaded box.

Pass/fail runs I did get:

- serial, the test alone, 5/5 passed (8.18 – 9.31 s each)
- `-n 4`, `tests/test_cluster_campaign_runner.py` alone, 3/3 runs, `11 passed` each

I did **not** reproduce a red run. What I am reporting is the measured margin and where
it goes, not a captured failure.

## What it is not

- **Not the #197 registry leak.** This test never reaches `detect_profile`; it touches
  none of `model_profiles/registry.py`, `vendored/OVERRIDE_ERRORS`, or
  `tessera_serving_scope.py`.
- **Not an xdist worker-shared-path defect.** Every path derives from `tmp_path`:
  `work_root` is `tmp_path/"sparky-work"` (`_local_host`, line 35-40) and the stage lock is
  `work_root/.prismaquant-cluster/<manifest identity_sha256>/locks/<stage id>.lock`
  (`prismaquant/cluster_campaign.py:1222-1228`). `tmp_path` is per-worker under xdist and
  `identity_sha256` differs per test, so two workers cannot collide on the lock.

## Second fixed deadline in the same test

`coordinator.wait(timeout=5)` (line 620) raises `subprocess.TimeoutExpired` — an ERROR,
not a clean failure — if the coordinator does not die within 5 s of `terminate()`. It did
not fire in any run I took, but it is the same kind of constant and worth moving with the
first.

## Severity

**P2** — observability. A gate whose colour tracks box load rather than the code under
test is a gate whose result nobody can read. It cannot ship a wrong artifact (it is a
test), and no number a decision reads comes out of it.

## What would fix it

Wait on the coordinator's own readiness rather than a constant: poll until the state file
records `running` with a pid **or the coordinator process has exited** (`Popen.poll()` is
not None), with a deadline generous enough to be about the campaign runner and not about
interpreter start — a minute, say. That turns "slow box" into "still waiting" and keeps a
genuinely wedged coordinator a fast, clear failure instead of an 8-second race. Same
treatment for the `wait(timeout=5)` on line 620.

## Why filed instead of fixed

Filed under AGENTS.md principle 14. **I could have fixed it and chose not to**: it is a
test-only change of a few lines and I had the worktree open. My brief for #197 scoped this
one explicitly — characterise it, and fix it only if it turned out to be the same
process-global leak or an xdist shared-path defect, otherwise file what I measured. It is
neither, so this is the file. Happy to take the fix on request; it does not need a
measurement I am not set up to take.

Found while fixing #197, on branch `claude/pq-197-registry-leak`.

RobTand:
Fixes #199.

`test_resume_after_coordinator_death_recovers_owned_helper_receipt` waited for the
coordinator's first state file against `time.monotonic() + 8.0`. Almost none of that
budget is campaign work: it is a cold `python -m prismaquant.cluster_campaign` —
interpreter start plus this package's import graph — so the margin tracked box load
rather than anything the test asserts.

## What I measured before the fix

sparklina, CPU only, `/home/rob/venvs/pq-release/bin/python`,
`PYTHONPATH=/home/rob/tessera-pin-1221d2a/src`. The deadline was lifted to 120 s so the
number is latency rather than pass/fail. Load above ~10 is CPU spinners I started and
killed by exact PID.

| 1-min load | time to `observed_running` | margin vs the 8.0 s deadline |
|---|---|---|
| 3.1 | 2.69 s | +5.31 s |
| 4.7 / 7.2 / 9.4 | 2.92 / 2.88 / 2.76 s | +5.08 / +5.12 / +5.24 s |
| 41.1 | 10.85 s | **−2.85 s** |
| 76.1 | 17.34 s | **−9.34 s** |
| 90.0 | 19.33 s | **−11.33 s** |

Attribution: a bare `python -c "import prismaquant.cluster_campaign"` is 3.25 / 3.26 /
3.31 s at load ~5, so the import alone is ~41% of the 8.0 s budget before the
coordinator does any campaign work.

**The filer could not capture a red run; this branch has one.** With 100 spinners of
mine running, the pre-fix tree goes red under `-n 2` — and not only in the test #199
names:

```
$ pytest tests/test_cluster_campaign_runner.py -p no:randomly -n 2 -q
E AssertionError: coordinator never recorded slow-stage as running with a pid; state=(absent)
E AssertionError: assert False + where False = is_file()   # child.pid
E prismaquant.cluster_campaign.CampaignTerminalFailure: terminal failed stages: ['materialize-plan']
3 failed, 8 passed in 52.71s
```

So the finding is the file's rule, not one constant.

## The shape I picked, and why

The issue offered two shapes: derive the budget from a timed cold import, or wait on a
bound that is not load-sensitive. **I took the second.** Deriving a budget still leaves a
multiple to guess, and it charges every test an extra multi-second import to measure
something it then only uses as a bound. Liveness needs no constant at all: every wait
here is "a child process will make an observable thing appear", so each now waits *while
its producer can still satisfy it* — `Popen.poll()`, or a PID probe for the stage child
that deliberately outlives its worker — and gives up the moment it cannot.

A slow box now makes this file slower; it cannot make it red. A producer that dies
without doing the work — the failure these tests are actually about — still fails at once
and by name.

One helper, `_wait_until`, and one named constant, `_WEDGE_BACKSTOP_SECONDS`. That
constant is a wedge detector, not a budget: 120 s sits far above the 19.3 s worst case
measured above and below CI's `--timeout=300` per test, so a genuinely wedged producer
fails by name here instead of being killed by the harness.

## Commits

**1. Bound the waits by producer liveness** — replaces five wall-clock constants: the
`+8.0` and `+30.0` deadlines and `coordinator.wait(timeout=5)` in the resume test, and
the `+8`, the `+5` and two `worker.wait(timeout=5)` calls in
`test_abrupt_worker_death_leaves_stage_lock_owned_by_child`. The `+5` receipt wait there
was additionally an ERROR rather than a failure when it tripped: it fell through to
`receipt_path.read_bytes()` and raised `FileNotFoundError`.

**2. Retire the remaining wall-clock budgets** — three more of the same shape, which are
what fails next once the waits are fixed:

- `_stage`'s default `timeout_seconds=10` is what terminal-failed `materialize-plan`
  above. No test in this file asserts that a stage times out, so it is a wedge detector
  too; it now uses the same named constant, and the four per-stage overrides (20, 60,
  10, 10) are gone, so the file has one number in one place rather than five guesses.
- The `waiting-remote` child carried its own `time.monotonic()+12` deadline and
  `sys.exit(8)` — a second, shorter guess underneath the stage timeout the worker
  already enforces by terminating the child's process group. A gate stage that was
  merely still starting up became a terminal stage failure.
- `test_abrupt_worker_death…` gave itself whatever was left of the child's
  `time.sleep(2.5)` to kill the worker and prove the lock was still held — a wall-clock
  budget minus a cold interpreter start. The child now holds the lock until the test
  writes a release file, so the window is exact instead of probabilistic, and the
  `finally` writes that file on every path out so a failure cannot strand an orphan (its
  worker is dead by then, so nothing else would bound it).

Found by fixing the above: with the waits no longer expiring, `-n 2` under load got as
far as reading `child.pid` and raised `ValueError: invalid literal for int() with base
10: ''`. A file becomes visible in its directory before the writer's bytes land, so
waiting on `Path.is_file` hands the next line a half-written file. The three waits that
read a file's contents now wait on `_nonempty`.

## After

Same 100-spinner load, `-n 2`: **`11 passed in 103.77s` at 1-min load 89**, against
`3 failed, 8 passed in 52.71s` at load 72 before. Slower under load, and no longer wrong
about it — the trade this file was making backwards.

## What I deliberately did not change

- `_worker_lock(timeout_seconds=1)` at the "lock is still held" assertion. The child is
  blocked on the release file by construction there, so a prompt refusal **is** the
  assertion, not a budget. The later acquisition, which waits for the child to exit,
  moved to the backstop.
- `connect_timeout_seconds: 5` on the fake SSH host. The fake `ssh` `os.execv`s
  immediately and never has a connect phase; the flag is argv the fake ignores.

No `pytest.mark.flaky`, no retries, no bigger number: the two numbers that remain are one
named backstop and one assertion.

## Docs

Test-only. `docs/ARCHITECTURE.md` has no test-hygiene section, and neither
`test_architecture_doc.py` nor `test_docs_staleness.py` covers this file or documents a
wait/deadline convention, so no provenance re-stamp is owed under AGENTS.md principle 12
(which scopes to pipeline defaults, the stage graph, the format menu, the plugin
contract, serving-lane defaults and ship gates). Both doc tests run green anyway: `19
passed`.

## Third commit: the race the deadline was hiding

Removing the deadline that used to expire first exposed a race that was always
underneath it, and CI caught it (run 33981895346, ubuntu-24.04, full suite serially, no
synthetic load):

```
FAILED …::test_resume_after_coordinator_death_recovers_owned_helper_receipt
AssertionError: the orphaned helper never wrote its receipt … -- nothing happened in
120s, so the producer is wedged
```

The race is ordering, not budget. `_launch_stage` records the attempt's pid via
`bind_process` as soon as it has *spawned* the worker, and only afterwards writes the
worker's request to its stdin. The recorded pid is the worker's, so
`status == "running" and pid is not None` becomes true while the worker is still blocked
reading a request it has not been given. Terminating the coordinator in that window
closes the write end of the pipe, the worker sees EOF, exits without ever launching a
stage child, and no receipt is ever written.

A second race came from the same `time.sleep(1.0)`: once a child did exist, the test had
whatever was left of that sleep to kill the coordinator, or the coordinator would finish
the campaign normally and leave the resume nothing to adopt.

Both die with the sleep. The helper now announces itself, the test waits for that
announcement before terminating the coordinator (bounded by the coordinator, which has to
stay alive to get the worker that far), and then releases the helper and waits for its
receipt. No wall-clock quantity is left in the sequence, and "the coordinator never got a
stage running" now fails by name instead of at the backstop.

## Scope note

One test in this file (`test_scheduler_refills_a_freed_host_while_another_host_is_active`)
can still miss its stage timeout at a 1-minute load around 220 on a 20-core box — an
11x oversubscription that is not a population this suite is expected to serve, and well
past the ~108 the issue reported. Not addressed here.

### DIFF
--- a/tests/test_cluster_campaign_runner.py
+++ b/tests/test_cluster_campaign_runner.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Callable
 import copy
 import hashlib
 import json
@@ -22,6 +23,75 @@ _WRITE_BYTES = (
     "Path(sys.argv[1]).parent.mkdir(parents=True, exist_ok=True); "
     "Path(sys.argv[1]).write_bytes(sys.argv[2].encode('utf-8'))"
 )
+
+
+# ---------------------------------------------------------------------------
+# Cross-process waits.
+#
+# Every wait in this file is "a child process will make an observable thing
+# appear" -- a state file, a pid file, a receipt.  What that costs is dominated
+# by interpreter start plus this package's import graph, not by the work the
+# child was asked to do.  Measured on sparklina (2026-09-05, CPU only): a cold
+# `python -m prismaquant.cluster_campaign` reached its first state file in
+# 2.7s on an idle box and in 10.9 / 17.3 / 19.3s at 1-minute load 41 / 76 / 90.
+#
+# A wall-clock constant sized against the first number is a coin flip under the
+# second, which is what issue #199 measured and what turned three tests in this
+# file red under `-n 2` on a loaded box.  So no wait here is sized in seconds.
+# Each waits while the producer can still satisfy it and gives up the moment it
+# cannot.  A slow box makes this file slower; it cannot make it red.  A
+# producer that dies without doing the work, which is the failure these tests
+# are actually about, still fails at once and by name.
+# ---------------------------------------------------------------------------
+
+_WEDGE_BACKSTOP_SECONDS = 120
+"""Last-resort bound on every cross-process wait and stage in this file.
+
+Not a budget for the work.  It answers only "the producer is wedged and nothing
+will ever satisfy this", so it sits far above anything a healthy run reaches --
+the slowest wait measured under load was 19.3s.  A healthy run never observes
+it; a run that does has found a real defect, and reports it as one.
+"""
+
+
+def _wait_until(
+    condition: Callable[[], bool],
+    *,
+    unmet: str,
+    alive: Callable[[], bool] | None = None,
+    diagnose: Callable[[], str] | None = None,
+    poll_interval: float = 0.02,
+) -> None:
+    """Wait for ``condition`` for as long as its producer can still satisfy it.
+
+    ``alive`` reports whether the process expected to satisfy ``condition`` is
+    still running.  When it stops being true the condition gets one last look --
+    the producer may have done the work in the window before it exited -- and
+    then the wait fails, because nothing can satisfy it any more.  Omit
+    ``alive`` only where the test holds no handle on the producer, such as a
+    deliberately orphaned helper; the backstop is then the sole bound.
+
+    ``unmet`` is the failure sentence and ``diagnose`` appends state to it.
+    """
+    backstop = time.monotonic() + _WEDGE_BACKSTOP_SECONDS
+
+    def _fail(why: str) -> None:
+        detail = f"; {diagnose()}" if diagnose is not None else ""
+        raise AssertionError(f"{unmet} -- {why}{detail}")
+
+    while True:
+        if condition():
+            return
+        if alive is not None and not alive():
+            if condition():
+                return
+            _fail("its producer exited first")
+        if time.monotonic() >= backstop:
+            _fail(
+                f"nothing happened in {_WEDGE_BACKSTOP_SECONDS}s, so the "
+                "producer is wedged"
+            )
+        time.sleep(poll_interval)
 
 
 def _digest(text: str) -> str:
@@ -596,28 +666,37 @@ def test_resume_after_coordinator_death_recovers_owned_helper_receipt(tmp_path):
         stdout=subprocess.DEVNULL,
         stderr=subprocess.DEVNULL,
     )
-    deadline = time.monotonic() + 8.0
-    observed_running = False
-    while time.monotonic() < deadline:
-        if state_path.is_file():
-            try:
-                raw = json.loads(state_path.read_text(encoding="utf-8"))
-                attempt = raw["stages"]["slow-stage"]["attempts"][-1]
-                observed_running = (
-                    raw["stages"]["slow-stage"]["status"] == "running"
-                    and attempt["pid"] is not None
-                )
-            except (json.JSONDecodeError, KeyError, IndexError):
-                observed_running = False
-            if observed_running:
-                break
-        time.sleep(0.02)
-    assert observed_running, (
-        "coordinator never recorded slow-stage as running with a pid; "
-        f"state={state_path.read_text(encoding='utf-8') if state_path.is_file() else '(absent)'}"
+    def _recorded_running_with_pid() -> bool:
+        if not state_path.is_file():
+            return False
+        try:
+            raw = json.loads(state_path.read_text(encoding="utf-8"))
+            attempt = raw["stages"]["slow-stage"]["attempts"][-1]
+        except (json.JSONDecodeError, KeyError, IndexError):
+            return False
+        return (
+            raw["stages"]["slow-stage"]["status"] == "running"
+            and attempt["pid"] is not None
+        )
+
+    # Nearly all of this latency is the coordinator's cold start rather than
+    # its campaign work, so the wait is on the coordinator, not on a clock.
+    _wait_until(
+        _recorded_running_with_pid,
+        unmet="coordinator never recorded slow-stage as running with a pid",
+        alive=lambda: coordinator.poll() is None,
+        diagnose=lambda: "state="
+        + (
+            state_path.read_text(encoding="utf-8")
+            if state_path.is_file()
+            else "(absent)"
+        ),
     )
     coordinator.terminate()
-    coordinator.wait(timeout=5)
+    _wait_until(
+        lambda: coordinator.poll() is not None,
+        unmet="coordinator did not exit after terminate()",
+    )
 
     # Wait for the orphaned helper to finish, on its receipt rather than on
     # the clock. What this test asserts is that a resume ADOPTS a completed
@@ -625,13 +704,15 @@ def test_resume_after_coordinator_death_recovers_owned_helper_receipt(tmp_path):
     # completed. Probing the instant the coordinator dies asserts something
     # else: that the helper wins a race against the resume. It usually did,
     # and lost under load, which is how this arrived in CI as a flake.
-    helper_deadline = time.monotonic() + 30.0
-    while time.monotonic() < helper_deadline and not receipt_path.is_file():
-        time.sleep(0.02)
-    assert receipt_path.is_file(), (
-        "the orphaned helper never wrote its receipt, so there is nothing "
-        "for the resume to adopt -- the coordinator's termination most "
-        "likely took the helper with it"
+    # The helper was deliberately orphaned, so this test holds no handle on it:
+    # the backstop is the only bound available here.
+    _wait_until(
+        receipt_path.is_file,
+        unmet=(
+            "the orphaned helper never wrote its receipt, so there is nothing "
+            "for the resume to adopt -- the coordinator's termination most "
+            "likely took the helper with it"
+        ),
     )
 
     state = campaign.run_campaign_v2(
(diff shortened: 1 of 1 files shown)


RobTand:
## Commands and counts

All on sparklina, CPU only, `PYTHONPATH=/home/rob/tessera-pin-1221d2a/src`,
`CUDA_VISIBLE_DEVICES=""`, `/home/rob/venvs/pq-release/bin/python`, xdist capped at
`-n 2`. Every CPU spinner referenced below was started and killed by exact PID by
`/home/rob/pq-audit-caches/pq-199/loadrun.sh`; no `pkill`.

### Pre-fix, red (base `bf4b589c`, 100 spinners I own)

```
$ pytest tests/test_cluster_campaign_runner.py -p no:randomly -n 2 -q
3 failed, 8 passed, 28 warnings in 52.71s          # 1-min load 72
FAILED …::test_resume_after_coordinator_death_recovers_owned_helper_receipt
FAILED …::test_abrupt_worker_death_leaves_stage_lock_owned_by_child
FAILED …::test_sealed_stage_wraps_dynamic_output_and_withholds_receipt_on_failure
```

After commit 1 only, same load: `1 failed, 10 passed in 72.53s` — both deadline
failures gone, the stage-budget one left, which is what commit 2 fixes.

### Post-fix, same 100-spinner load

```
$ pytest tests/test_cluster_campaign_runner.py -p no:randomly -n 2 -q
11 passed, 28 warnings in 103.77s                  # 1-min load 89
```

### Required runs (merged tree, tip `4ed38966`)

Serial, 3 runs:

```
$ pytest tests/test_cluster_campaign_runner.py -p no:randomly -q
11 passed, 14 warnings in 20.24s     # load 40.08
11 passed, 14 warnings in 19.74s     # load 31.24
11 passed, 14 warnings in 20.03s     # load 23.29
```

`-n 2`, 3 runs:

```
$ pytest tests/test_cluster_campaign_runner.py -p no:randomly -n 2 -q
11 passed, 28 warnings in 16.47s     # load 16.55
11 passed, 28 warnings in 17.06s     # load 14.49
11 passed, 28 warnings in 17.42s     # load 11.36
```

Docs (test-only change, but run to confirm nothing moved):

```
$ pytest tests/test_architecture_doc.py tests/test_docs_staleness.py -p no:randomly -q
19 passed, 14 warnings in 3.28s
```

### Attribution measurement

```
$ python -c 'import prismaquant.cluster_campaign'    # cold, subprocess-timed
3.25 / 3.26 / 3.31 s at 1-min load ~5
```

Latency to `observed_running` with the deadline lifted to 120 s, so the number is
latency rather than pass/fail: 2.69 s at load 3.1; 2.76–2.92 s at load 4.7–9.4;
10.85 s at load 41.1; 17.34 s at load 76.1; 19.33 s at load 90.0. Against the shipped
8.0 s deadline that is +5.31 s of margin idle and −2.85 to −11.33 s under load.

Totals: 2 commits plus two merges of `origin/main` (the second to pick up the CI
tests-job timeout raise in #205, so this PR gets a full-length run); 1 file changed
(`tests/test_cluster_campaign_runner.py`); 9 wall-clock constants removed, replaced by
1 named backstop and 1 deliberate assertion timeout. No production code touched.

### The fix did not make these tests unable to go red

Load-insensitivity is worthless if it also removed the failure. Negative checks against
the shipped helper (`/home/rob/pq-audit-caches/pq-199/verify_failfast.py`):

```
PASS  producer exits without doing the work: failed in 0.02s
      -- the thing never appeared -- its producer exited first; diagnostic-was-called
PASS  producer alive but wedged: failed in 2.00s (cap shortened for the check)
      -- nothing happened in 2s, so the producer is wedged
PASS  _nonempty(empty file) = False   # the ValueError race
PASS  _nonempty(file with bytes) = True
PASS  _nonempty(missing file) = False
PASS  last-look race: producer wrote then exited -> returned in 0.03s
```

A coordinator that never records the stage now fails in **0.02 s** instead of after the
old 8 s deadline — the failure got faster and keeps its name, while only the healthy-path
patience grew.

### Final state — tip `f562e664`

Third commit added (`7236f826`, the launch-ordering race CI caught), then `origin/main`
merged in (PQ #209, `cdaf8931`). Verified on the merged tree, serially, CPU only:

```
$ pytest tests/test_cluster_campaign_runner.py -p no:randomly -q
11 passed, 14 warnings in 19.40s
```

Branch diff is one file, `tests/test_cluster_campaign_runner.py`; no production code and
no docs, so no ARCHITECTURE re-stamp is owed.

**Heads-up on this PR's CI, which is red for a reason that is not this branch.** The
`tests` job reports 18 failures: 1 was the race fixed in `7236f826`, and the other **17
are pre-existing on `main`** — all in `tests/test_tessera_campaign_resume.py`, all
`RuntimeError: Tessera campaign checkpoint identity requires the producer's cached_unit
input/byte receipt API`. Reproduced on a clean worktree of pristine `origin/main` with no
changes of mine: `17 failed, 2 passed`. `tools/resolve_tessera_dev_pin.py` resolves to
`1221d2a4207a`, and that commit's `src/tessera/` has no `cached_unit`, while
`prismaquant/tessera_campaign.py:558` imports it and refuses without it. Not touched
here — a pin move is not this PR's call.

PR #207, opened 2026-09-05
pq-decisions: #175 receipt (no token needed), #87 reopened + driver replays the candidate decision

RobTand:
Closes out the two items parked "for Rob" (worker `pq-decisions`, 2026-09-05). Three commits, each independently droppable.

## #175 — CI and the pinned Tessera: no token bridge, the receipt already exists

The brief's plan was an optional-token bridge (`secrets.TESSERA_REPO_TOKEN || github.token`). It is not needed: **RobTand/tessera went public on 2026-09-05** (between the 03:04Z run that still failed at checkout and the 06:45Z run that succeeded), and the unchanged workflow then installed the pin with the default checkout credential in both jobs and ran green end-to-end.

Receipt, run [33950707718](https://github.com/RobTand/prismaquant/actions/runs/33950707718) (`main` at `8a77b0ad`, push, 06:45Z): `Check out pinned Tessera` fetched `1221d2a4207a6baeffbe9726bce13125fc1649ae` from `https://github.com/RobTand/tessera`, `Install pinned Tessera` → `Successfully installed tessera-quant-0.1.0`, both jobs; suite 4698 passed / 136 skipped / 3 xfailed in 21m20s. That is the issue's own completion criterion ("successful dependency installation on GitHub"). The newest complete run, [33976858139](https://github.com/RobTand/prismaquant/actions/runs/33976858139), has the same two steps green; its 17 failures are all `test_tessera_campaign_resume.py` main-entry cases inherited at the dev pin (`tessera.cached_unit` absent), fixed by the re-pin in #192.

* `05a4683a` — the workflow comment and ARCHITECTURE.md §8.6 still said "checkout failure is expected while Tessera remains private"; corrected on sight (AGENTS.md 14, prose), stamp added. No secret is added or wanted: it would be the cross-repo credential the issue avoided, and it would not reach fork PRs anyway.
* `769c8b75` — **off-task, tripped over reading run durations**: the `tests` job took 15–29.5 min across the ten completed jobs of 2026-09-05 (job 33973973959 finished at **29m30s against the 30 min limit** and passed). The header's own 2026-08-03 rule said a suite doubling "should raise `timeout-minutes` in the same PR"; the suite is 1.75x larger. Raised to 60 (≈2x on the worst observed job), header re-measured. Nothing pins the old value.

#175 is closed with a comment citing the run IDs and step names (coordinator's instruction); Rob has nothing to click.

## #87 — pending GPU validation and `replay_no_new_failures`

[...]

PR #209, opened 2026-09-05
[P2] Stop swallowing the dead-vendored-override refusal at the remaining 22 call sites (#202)

### LINKED ISSUE #202: [P2] 22 call sites still swallow the dead-vendored-override refusal one level up (#201 follow-up)

RobTand:
Closes the census filed with #201. `_resolve` now raises `DeadVendoredOverrideError` and three gate call sites stopped re-swallowing it (PR #203); this branch takes the remaining **22 call sites across 11 modules**, each of which wrapped detection in a broad `except Exception` and continued with a substituted profile — converting the refusal straight back into the silent wrong answer one level up.

Read one at a time against its own local contract, as the issue asked. **The answer was the same at all 22: re-raise.** That includes both sites the census singled out as possible legitimate swallows; details below, because both turned out to be among the strongest cases rather than the weakest.

## The 22, with the decision per row

Enclosing function named, since several sites sit far from their `def`.

| # | file:line | function | call | substituted | decision |
|---|---|---|---|---|---|
| 1 | `aqua_activation_cost.py:661` | `main` | `detect_profile` | logs, then generic name matching | re-raise |
| 2 | `build_rtn_cache.py:115` | `stage_multimodal` | `detect_profile` | `keep_composite = False` | re-raise |
| 3 | `build_rtn_cache.py:287` | `iter_quantizable_tensors` | `profile_from_model` | `profile = None` | re-raise |
| 4 | `build_rtn_cache.py:500` | `main` | `detect_profile` | `profile = None` | re-raise |
| 5 | `incremental_measure_quant_cost.py:585` | `_run_body_cost_shard` | `profile_from_model` | `profile = None` | re-raise |
| 6 | `incremental_measure_quant_cost.py:1032` | `_run_visual_cost_shard` | `profile_from_model` | `profile = None` | re-raise |
| 7 | `layer_streaming.py:1015` | `_build_concat_merger` | `profile_from_model` | `return None` | re-raise |
| 8 | `layer_streaming.py:1058` | `_build_expert_packer` | `profile_from_model` | `return None` | re-raise |
| 9 | `layer_streaming.py:1140` | `fill_packed_experts_from_source` | `profile_from_model` | `return 0` | **re-raise** (flagged as optional; see below) |
| 10 | `layer_streaming.py:2509` | `_head_prefixes` | `profile_from_model` | hardcoded `hc_head` guess | re-raise |
| 11 | `measure_quant_cost.py:3386` | `run_cost_pass` | `profile_from_model` | `model_profile = None` | re-raise |
| 12 | `perturbed_x_cache.py:1366` | `stage_text_only_under_work_root` | `detect_profile` | `profile = None` | re-raise |
| 13 | `production_weight_cache.py:6571` | `fill_production_weight_cache` | `profile_from_model` | `model_profile = None` | re-raise |
| 14 | `sensitivity_probe.py:132` | `stage_text_only` | `detect_profile` | `profile = None` | re-raise |
| 15 | `sensitivity_probe.py:1388` | `discover_moe_structure` | `profile_from_model` | `profile = None` | re-raise |
| 16 | `sensitivity_probe.py:1515` | `discover_moe_routers` | `profile_from_model` | `profile = None` | re-raise |
| 17 | `sensitivity_probe.py:2006` | `FisherAccumulator.__init__` | `profile_from_model` | `model_profile = None` | re-raise |
| 18 | `sensitivity_probe.py:3592` | `run_streaming_multimodal_visual_probe_pass` | `profile_from_model` | `DefaultProfile()` | re-raise |
| 19 | `streaming_model.py:112` | `_bypass_hf_fp8_module_rewrite` | `profile_from_config` | `return False` | **re-raise** (flagged as optional; see below) |
| 20 | `streaming_model.py:218` | `_init_rotary_inplace` | `profile_from_model` | `pass`, falls through | re-raise |
| 21 | `streaming_production_cache.py:1776` | `fill_production_weight_cache_streaming` | `profile_from_model` | `profile = None` | re-raise |
| 22 | `weight_session.py:90` | `WeightSession.__init__` | `profile_from_model` | `profile = None` | re-raise |

### The two "optional hint" sites the issue asked about

The issue said these two might legitimately share an answer between "no profile" and "dead profile". Reading them, both go the other way — and both are clearer cases than the plain `profile = None` sites.

**`layer_streaming.py:1140`** returns a count, and every other `return 0` in it means "nothing to do". But this function exists *because* the packed params would otherwise stay zero-initialized and, in its own docstring, "silently break every activation-scale calibration that depends on the routed-expert output". Returning `0` on a dead override **is** that breakage, reported as a clean no-op. `0` is only honest when the profile could be consulted and had nothing to say.

**`streaming_model.py:112`** returns a bool, but control only reaches it for a checkpoint that is *already* native FP8 with block scales — so `False` is a decision, not a neutral default. Its docstring says the bypass is "a static architecture property, so it is declared in the model profile ... rather than pattern-matched on the model name". A dead override is precisely the case where that declaration exists and cannot be read, so `False` runs transformers' FP8 module rewrite on the architecture whose profile would have forbidden it — the MiniMax-M2 `FP8Experts` / `experts.0.w1` failure the docstring names.

Neither keeps a swallow, so there is no "why we kept it" comment to write.

### The severest rows

`production_weight_cache.py:6571` and `streaming_production_cache.py:1776` render the production weight cache — the bytes an export ships. Both also lose `pinned_names()` when the profile goes to `None`, so components the allocator is *not* allowed to quantize (AGENTS.md 7) would be rendered too. One swallowed refusal changed both what was protected and what the renderer knew.

## One rule, one home — and why no shared helper

No `detect_profile_or_refuse_dead` was added, deliberately. The rule — what "dead" means, and that it must never be answered — keeps its single home in `_refuse_dead_vendored_override` / `DeadVendoredOverrideError`; every site imports the class from the registry and adds nothing of its own.

A helper would not have removed duplication here, because **it would have had to own each site's fallback too**, and those differ across all 22: `None`, `DefaultProfile()`, `keep_composite = False`, `return 0`, `return False`, a hardcoded prefix guess, `pass`, a log line. A helper parameterised on the fallback is longer than the two lines it replaces, and a helper that swallows on the site's behalf would have moved *tolerance policy* into the registry — the one module whose docstring refuses silent fallbacks. Three sites also wrap more than the detection call in the same `try` (`build_rtn_cache:115`, `layer_streaming:2509`, `streaming_model:218`), which no helper can cover without changing what else is tolerated.

So the edit is the uniform two lines the issue itself proposed, matching the shape PR #203 established at its three gate sites. **The "no profile matched" path is unchanged at every one of the 22.**

One consequence worth flagging: each site's import is **hoisted just outside its `try`**, because a handler cannot name a class whose import is inside the block it guards. That makes an `ImportError` on `prismaquant.model_profiles` fatal at these sites rather than silently tolerated. Same trade PR #203 took and documented; these are function-local imports for cycle-avoidance at module load, not because the in-tree package might be absent.

## Tests

18 behavioural regressions plus **one tree-wide census**, all in `tests/test_vendored_override_gate.py` — beside #201's, not scattered per module, because the rule is the registry's and it has one home (the precedent PR #203 set).

The census walks `prismaquant/` and fails for any call to `detect_profile` / `detect_profile_with_warning` / `profile_from_config` / `profile_from_model` inside a `try` whose broad handler would answer instead of refusing. Two shapes pass: `except DeadVendoredOverrideError: raise` ahead of the broad handler, or a broad handler that always re-raises — which is what `sample_parallel_probe.py:568` already did (`SampleParallelProbeError(...) from exc`), and why it was correctly excluded from the census.

It earns its place twice. Four sites cannot be reached without a GPU, a loaded checkpoint or a built streaming context — `aqua_activation_cost.py:661` (inside `main()`, past argparse and two artifact loads), `build_rtn_cache.py:500`, `sensitivity_probe.py:3592`, `streaming_production_cache.py:1776` — so a behavioural test for those would exercise scaffolding, not the fix. And it fails for the 23rd such call site when someone writes it, which no per-site test does.

Run independently before any fix, it reproduced the issue's census table **line for line** — same 22 `file:line` pairs, same entrypoints:

```
FAILED test_no_call_site_swallows_the_dead_override_refusal
  - AssertionError: assert [...] == []
    Left contains 22 more items, first extra item:
    'prismaquant/aqua_activation_cost.py:661 (detect_profile)'
```

Every commit records its own pre-fix failure lines. One is not what it looks like and says so: `weight_session.py:90`'s refusal already arrives once `build_rtn_cache.iter_quantizable_tensors` stops swallowing, so its test passes against the branch tip either way — it fails on the genuine pre-#202 tree, and the site is fixed anyway rather than left covered-by-accident.

## Commits

One per module, twelve plus a test-hygiene commit and the merge:

```
aqua_activation_cost · build_rtn_cache · incremental_measure_quant_cost
layer_streaming · measure_quant_cost · perturbed_x_cache
production_weight_cache · sensitivity_probe · streaming_model
streaming_production_cache · weight_session
+ tree-wide census pin & ARCHITECTURE re-stamp   (Fixes #202)
+ keep the cost-shard regressions inside tmp_path
```

## `docs/ARCHITECTURE.md`

Re-stamped in the same commit as the census pin. §8.1's **body** claimed the 22 "can still convert this refusal", which is now false, so that paragraph is rewritten: what the census found, that both flagged optional-hint sites went the same way as the rest, that no shared helper was added and why, and where the rule is now pinned. #201's **stamp** is left exactly as written — stamps are dated records of their own change, and it described #201's scope accurately.

## Risk

**No behaviour change on a healthy tree.** `OVERRIDE_ERRORS` is empty unless a vendored override actually failed to take effect, so every one of these `except` clauses is unreachable in a normal run. No default, format menu, serving lane, ship gate or exported byte moves. What changes is that a checkpoint on a dead vendored modelling path now refuses at the call site instead of being quietly answered.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,30 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-05 · `claude/pq-201-dead-override`. Stamps
+As of: 2026-09-05 · `claude/pq-202-detection-swallows`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-05, `claude/pq-202-detection-swallows`) for **the
+dead-vendored-override refusal reaching every call site** (§8.1; PrismaQuant
+#202, follow-up to #201). #201 made `_resolve` raise
+`DeadVendoredOverrideError` and stopped three gate call sites re-swallowing
+it; the census filed with that issue found 22 more, across 11 modules, each
+wrapping detection in a broad `except Exception` and continuing with a
+substituted profile — so the refusal was converted back into the silent wrong
+answer one level up. All 22 now re-raise this one class ahead of their broad
+handler. Each was read against its own contract, and the answer was the same
+at all 22, including the two the census flagged as possibly legitimate
+swallows: `layer_streaming.py:1140` returns a count, but returning `0` on a
+dead override reports exactly the zero-initialized-experts breakage the
+function exists to prevent; `streaming_model.py:112` returns a bool, but is
+only reached for a native-FP8 block-scaled checkpoint, where `False` runs the
+HF module rewrite the profile would have forbidden. No shared helper was
+added — the rule keeps its single home in the registry, and a helper would
+have had to own each site's differing fallback. No default, format menu,
+serving lane or exported byte moves, and the *absent*-profile path is
+unchanged everywhere; what changes is that a dead vendored path now refuses at
+the call site instead of being answered. The rule is pinned tree-wide by an
+executable census that fails for any future detection call inside a broad
+`except`. Gate: `tests/test_vendored_override_gate.py`.
 
 Re-stamped (2026-09-05, `claude/pq-201-dead-override`) for **a
 dead-vendored-override refusal that actually refuses** (§8.1; PrismaQuant
@@ -8683,14 +8706,30 @@ completeness *gate* that answered `None`), `incremental_probe._detect_profile_fo
 answered `DefaultProfile`) and `validate_native_export._resolve_validation_target_profile` (a
 *validator* that answered `None`). Each documents tolerance for an architecture this build does
 not know, which is a different statement from a known architecture on a dead path.
-**Not fixed here:** 22 further broad-`except` swallows around
-`detect_profile`/`profile_from_config`/`profile_from_model`, across 11 modules, can still convert
-this refusal into `None`, `DefaultProfile`, `False` or a skipped branch at their own call sites.
-Several of those are genuinely optional hints where `None` is the right answer for an *absent*
-profile and the wrong one for a *dead* one, so they need reading one at a time against their own
-contract; censused with `file:line` in #202 rather than swept inside this diff. One site already
-has the right shape and is excluded: `sample_parallel_probe.py:568` re-raises as
-`SampleParallelProbeError(...) from exc`.
+The remaining 22 broad-`except` swallows around
+`detect_profile`/`profile_from_config`/`profile_from_model`, across 11 modules, were closed by
+#202 (2026-09-05). Each was read against its own contract rather than swept: the answer was the
+same at all 22, including the two the census singled out as possible legitimate swallows
+(`layer_streaming.py:1140`, which returns a count, and `streaming_model.py:112`, which returns a
+bool). Both turned out to be the strongest cases rather than the weakest — the first exists
+*because* the packed params would otherwise stay zero-initialized, so a clean `0` reports the
+silent breakage it was written to prevent; the second is only reached for a native-FP8
+block-scaled checkpoint, so `False` is a decision, not a neutral default. Every site keeps its
+other tolerances: the `except DeadVendoredOverrideError: raise` clause sits ahead of the broad
+handler and nothing about the *absent*-profile path changed anywhere.
+
+**Where this rule lives.** No shared "detect or refuse" helper was added, deliberately. The rule —
+what "dead" means and that it must never be answered — has one home, `_refuse_dead_vendored_override`
+and `DeadVendoredOverrideError` in the registry; a helper would have had to own each call site's
+*fallback* too, and those differ (`None`, `DefaultProfile()`, `keep_composite = False`, `return 0`,
+`return False`, a hardcoded prefix guess, `pass`), so it would have moved tolerance policy into the
+registry rather than removing duplication. Each site's import is hoisted just outside its `try` so
+the new `except` clause can always be evaluated. The rule is pinned tree-wide by an executable
+census in `tests/test_vendored_override_gate.py`, which walks `prismaquant/` and fails for any
+detection call inside a broad-`except` `try` that does not either re-raise this class first or
+re-raise unconditionally — so a 23rd such call site fails the suite when it is written.
+`sample_parallel_probe.py:568` already had the second shape (`SampleParallelProbeError(...) from
+exc`) and passes unchanged.
 
 The `DefaultProfile` fallback is *guarded, not silent*: `allocator.py:1550-1554` calls
 `validate_default_profile_format_menu(...)` (`:961-988`), which refuses a multi-format menu
--- a/prismaquant/aqua_activation_cost.py
+++ b/prismaquant/aqua_activation_cost.py
@@ -655,10 +655,21 @@ def main() -> int:
     # no profile claims must not become a hard failure for those models. When
     # one IS detected it supplies the checkpoint aliases the generic index
     # cannot derive (see build_weight_resolver).
+    #
+    # "No profile claims this path" is the tolerated case. A dead vendored
+    # override is not it (#202): the architecture IS claimed, and dropping to
+    # generic name matching would price AQUA activation cost through an index
+    # built without the aliases that profile exists to supply. The import is
+    # hoisted out of the `try` so the handler below can always be evaluated.
     profile = None
+    from .model_profiles.registry import (
+        DeadVendoredOverrideError,
+        detect_profile,
+    )
     try:
-        from .model_profiles.registry import detect_profile
         profile = detect_profile(args.model_path)
+    except DeadVendoredOverrideError:
+        raise
     except Exception as exc:
         log(f"no model profile for {args.model_path} ({exc}); "
             f"falling back to generic name matching")
--- a/prismaquant/build_rtn_cache.py
+++ b/prismaquant/build_rtn_cache.py
@@ -110,10 +110,17 @@ def stage_multimodal(model_path: str):
     # skeleton from it (streaming_model._build_streaming_context).
     keep_composite = False
     if "vision_config" in cfg or "text_config" in cfg:
-        from .model_profiles import detect_profile
+        from .model_profiles import DeadVendoredOverrideError, detect_profile
         try:
             keep_composite = detect_profile(
                 model_path).requires_multimodal_skeleton()
+        except DeadVendoredOverrideError:
+            # Not the tolerated case (#202). `keep_composite = False` is the
+            # right answer for a checkpoint no profile claims; for one whose
+            # vendored modelling path is dead it silently picks the text-only
+            # flatten -- the exact staging the comment above says such a family
+            # "cannot survive" -- and the run continues against upstream code.
+            raise
         except Exception:
             keep_composite = False
 
@@ -282,9 +289,18 @@ def iter_quantizable_tensors(model, profile=None):
     Caller uses module.{param_attr}.data to read, and can REPLACE the data
     attribute to free storage. Handles nn.Linear and 3D packed MoE experts."""
     if profile is None:
+        from prismaquant.model_profiles import (
+            DeadVendoredOverrideError,
+            profile_from_model,
+        )
         try:
-            from prismaquant.model_profiles import profile_from_model
             profile = profile_from_model(model)
+        except DeadVendoredOverrideError:
+            # The profile owns which packed-MoE parameter names are
+            # quantizable. `None` drops to the legacy class-name fallback
+            # below, which silently yields a DIFFERENT set of tensors to
+            # quantize -- a wrong answer, not a missing one (#202).
+            raise
         except Exception:
             profile = None
     for name, mod in model.named_modules():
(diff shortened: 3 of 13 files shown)


RobTand:
## Commands and counts

Environment: sparklina, CPU only (`CUDA_VISIBLE_DEVICES=""`), `PYTHONPATH=/home/rob/tessera-pin-1221d2a/src`, `/home/rob/venvs/pq-cu130/bin/python`, xdist `-n 2`.

**Touched modules' test files + the two doc tests**, against the merged tree:

```
python -m pytest -q -n 2 \
  tests/test_vendored_override_gate.py tests/test_vendored_qwen3.py \
  tests/test_aqua_activation_cost.py tests/test_build_rtn_cache.py \
  tests/test_incremental_measure_quant_cost.py tests/test_incremental_shards.py \
  tests/test_layer_streaming.py tests/test_layer_streaming_fp8.py \
  tests/test_layer_streaming_mxfp4.py \
  tests/test_measure_quant_cost_packed_experts.py tests/test_perturbed_x_cache.py \
  tests/test_production_weight_cache.py tests/test_sensitivity_coverage.py \
  tests/test_streaming_production_cache.py tests/test_weight_session.py \
  tests/test_profile_detection_and_calibration.py \
  tests/test_model_profile_conformance.py \
  tests/test_qwen3_profile.py tests/test_incremental_probe.py \

560 passed, 74 skipped, 1 xfailed, 0 failed   (12.79s)
```

**Gate file alone**, which is where all 19 new tests live:

```
tests/test_vendored_override_gate.py tests/test_architecture_doc.py tests/test_docs_staleness.py
50 passed
```

`test_vendored_override_gate.py`: **12 → 31 tests** (18 behavioural regressions + 1 tree-wide census).

## Census counts

The executable census (`_swallowing_detection_sites()`) over `prismaquant/`, `archive/` excluded:

| | before | after |
|---|---|---|
| detection calls inside a broad-`except` `try` | 27 | 27 |
| …already refusing (#201's 3 gates + registry) | 4 | 4 |
| …refusing by re-raise (`sample_parallel_probe:568`) | 1 | 1 |
| **…swallowing the refusal** | **22** | **0** |

The 22 it listed pre-fix matched the issue's table line for line — same `file:line` pairs, same entrypoints — derived independently from the AST rather than copied from the issue.

## Per-module pre-fix failures

Each commit carries its own; the behavioural ones in full:

```
build_rtn_cache                  2 failed → 14 passed
incremental_measure_quant_cost   2 failed → 16 passed
layer_streaming                  4 failed → 20 passed
measure_quant_cost               1 failed → 21 passed
perturbed_x_cache                1 failed → 22 passed
production_weight_cache          1 failed → 23 passed
sensitivity_probe                4 failed → 27 passed
streaming_model                  2 failed → 29 passed
weight_session                   fails only on the genuine pre-#202 tree (see commit)
census pin                       22 sites listed → 0
```

Four sites (`aqua_activation_cost:661`, `build_rtn_cache:500`, `sensitivity_probe:3592`, `streaming_production_cache:1776`) are pinned by the census rather than behaviourally — each needs a GPU, a loaded checkpoint or a built streaming context to reach, so a behavioural test would exercise scaffolding rather than the fix. Named in their commits and in the census docstring.

## Not fixed here

`sensitivity_probe.stage_text_only` and `perturbed_x_cache.stage_text_only_under_work_root` are near-identical, down to the same hardcoded strip-key list in both — so that default has two homes. Both are fixed for #202, but merging them is a refactor well outside this diff. Noted in the `perturbed_x_cache` commit and reported rather than swept in.

RobTand:
## CI: merged current `main` (63ac2845), zero new failures

Rebuilt against `main` after PR #207 landed. The merge conflicted only on the ARCHITECTURE stamp block — both sides had added stamps at the top — resolved by keeping both, this branch's newest, with the `As of:` header naming it. My §8.1 body rewrite was untouched by the conflict (#207 edited §7.2 and §8.6).

`main`'s own CI is currently red, and for a reason this branch neither caused nor can fix — the known-inherited `tessera.cached_unit` gap at the pin:

| run | result |
|---|---|
| `main` @ 63ac2845 (#207 merge, [33981118226](https://github.com/RobTand/prismaquant/actions/runs/33981118226)) | **17 failed**, 4896 passed, 136 skipped, 3 xfailed, 174 subtests — 35m20s |
| this PR @ 00f25f8a ([33981728779](https://github.com/RobTand/prismaquant/actions/runs/33981728779)) | **17 failed**, **4915 passed**, 136 skipped, 3 xfailed, 174 subtests — 31m25s |

All 17 are `tests/test_tessera_campaign_resume.py` main-entry tests failing with `RuntimeError: Tessera campaign checkpoint identity requires the producer's cached_unit input/byte receipt API; refusing unbound resume`. The two failure sets are **byte-identical** — diffed test-id by test-id, nothing added, nothing accidentally fixed:

```
main:  17    this PR: 17
diff main-fails.txt mine-fails.txt   ->   (no output)
```

**+19 passing, and that is exactly this branch's 19 new tests** (4915 − 4896 = 19: 18 behavioural regressions + 1 tree-wide census). `import surface` passes.

Local re-run on the merged tree, same commands as the comment above: **560 passed, 74 skipped, 1 xfailed**, and the two doc gates green. The census re-run after merging #207 still reports **0** swallowing sites, so nothing #207 brought in reopened one.

PR #213, opened 2026-09-05
Price the served activation contract from the spec, not the format name (#205)

RobTand:
Fixes #205 (P1).

## What was wrong

A Tessera W4A4 rung is served by vLLM's `scaled_fp4_quant` with a **static** per-unit `input_global_scale` and UE4M3 block scales. The campaign prices its anchors that way (#196), but the two other consumers that price the same rung decided "which activation quantizer does this spec serve" with `canonical_format_name(spec.name) == "NVFP4"`:

- `perturbed_x_cache._activation_qdq` (assignment-KL hooks)
- `production_weight_cache._render_score_record` (cache scorer) via `_format_uses_static_activation_clip`

A Tessera name is not `"NVFP4"`, so both priced the rung through NVFP4's dynamic FP32-scale RTN — a quantizer the runtime never executes for it. The fill also only computed `activation_max_abs` when a stock NVFP4 format was in `render_base_fmt_set`, so the Tessera record never retained its calibration maximum (`activation_max_abs: null`). Codex's proof: at G=1 a 1e-3 block serves as exactly 0; the hook priced it at 1e-3 and the cache scored the row 0.0 where the served oracle scores it nonzero.

## The fix — one rule, one home

`FormatSpec` gains a last positional field `static_activation_contract` (`nvfp4_activation_contract.StaticActivationContract`: `execution`, `group_size`, `measured_as_served`, the owner's G rule `input_global_scale_from_max_abs`, `require_input_global_scale` refusing by name with `ActivationScaleContractError`, and `quantize_dequantize` = the owner's `nvfp4_activation_qdq_served`).

- `NVFP4` registry row carries it with `measured_as_served=False`: its **screen default stays clip + dynamic RTN**, the served emulation stays the `PRISMAQUANT_NVFP4_ACT_EMULATE_SERVED_SCALES` opt-in (behaviour identical, bit for bit).
- `tessera_render.synthesize_tessera_spec` takes the source row's contract and re-stamps `measured_as_served=True` (the plugin has no dynamic path). The no-G `activation_quantize_dequantize` stays NVFP4's callable by reference (the campaign's dynamic control). The campaign callback does not move.
- `_activation_qdq`: served-measurement → served oracle at the unit's G, no pre-clip, refuses by name without a calibrated maximum; screen default → historical behaviour; no contract → unchanged.
- `_render_score_record`: same rule; Tessera record now retains `activation_max_abs` and gains `input_global_scale` (always present, `None` for dynamically priced rows).
[...]

PR #224, opened 2026-09-05
Resolve a format name once, in the registry; make the production-cache fill gate total (#218)

### LINKED ISSUE #218: [P1] Production cache fill crashes with KeyError on INT4_W4A16_g128: the new static-activation gate upper-cases a mixed-case registry name and looks it up
## What

PR #213 replaced a total string test with a partial one on the production-weight-cache fill path, and the new one raises `KeyError` for a format PrismaQuant itself ships.

`prismaquant/production_weight_cache.py:1547-1562`:

```python
def _format_uses_static_activation_clip(fmt: str) -> bool:
    ...
    spec = fr.get_format(fr.canonical_format_name(str(fmt).strip().upper()))
    return _static_activation_contract_of(spec) is not None
```

`prismaquant/production_weight_cache.py:1565-1574`:

```python
def _formats_need_static_activation_max(formats) -> bool:
    ...
    return any(_format_uses_static_activation_clip(fmt) for fmt in formats)
```

and its caller, `fill_production_weight_cache` at `:6789-6797`:

```python
    render_base_fmt_set = {_render_base_format(fmt) for fmt in fmt_set}
    ...
    needs_static_activation_max = _formats_need_static_activation_max(
        render_base_fmt_set)
```

`_render_base_format` (`:168`) is `str(fmt).strip().upper()`, and `_format_uses_static_activation_clip` upper-cases again. `INT4_W4A16_g128` is a registered `FormatSpec` whose name is **mixed case** (`format_registry.py:1025-1033`), so upper-casing it produces `INT4_W4A16_G128`, which is neither a registry key nor an alias. `canonical_format_name` returns it unchanged and `get_format` raises.

The predicate this replaced was pure string membership and could not raise:

```python
    return fr.canonical_format_name(str(fmt).strip().upper()) == "NVFP4"   # old :1547
        or "NVFP4" not in render_base_fmt_set                              # old :6866
    if "NVFP4" in render_base_fmt_set:                                     # old :7028
```

Any production weight cache fill whose requested format menu or assignment contains `INT4_W4A16_g128` now dies with an unhandled `KeyError` at `:6796`, **before a single tensor is rendered**. That is the stage producing the bytes an export ships. Previously the same input answered "no NVFP4 requested", skipped the activation-max pass, and rendered normally.

`INT4_W4A16_g128` is reachable, not vestigial: `validation_harness.py:807` maps the 4-bit entry of a precision plan to it (`{16: "BF16", 8: "INT8_W8A16", 4: "INT4_W4A16_g128"}`) and `:843` returns it directly, so an assignment can carry it and `render_scope == "assignment"` puts it straight into `render_formats_by_qname` (`streaming_production_cache.py:1449-1453`) and hence into `render_base_fmt_set`.

The class of defect is wider than the one name: the gate now raises for *any* requested format string `get_format` cannot resolve, where before every unresolvable name simply answered `False`. The sibling predicate in the same file that resolves the same values already knows this — `_is_cb_format_name` (`:1778-1787`) wraps the identical `fr.get_format(fr.canonical_format_name(...))` in `try/except Exception: return False`. The new predicate does not.

## Evidence

```
$ CUDA_VISIBLE_DEVICES="" PYTHONPATH=/home/rob/tessera-pin-1221d2a/src python -c ...
alias in FORMAT_ALIASES: False -> None
canonical_format_name(as-is)   : INT4_W4A16_g128
canonical_format_name(uppered) : INT4_W4A16_G128
get_format(as-is) ok           : INT4_W4A16_g128

_render_base_format(n)         : INT4_W4A16_G128
gate RAISES KeyError: "Unknown format 'INT4_W4A16_G128'. Available: ['BF16', 'FP8', ...

OLD gate behaviour  ("NVFP4" in render_base_fmt_set): False
```

Sweeping every registry key and alias through the new gate finds exactly this one:

```
registry names/aliases that RAISE in the new gate: [('INT4_W4A16_g128', 'KeyError')]
```

The new tests cannot catch it — `tests/test_served_activation_contract.py:198-201` exercises only resolvable, all-upper names:

```python
    assert _formats_need_static_activation_max({W4A4}) is True
    assert _formats_need_static_activation_max({"NVFP4"}) is True
    assert _formats_need_static_activation_max({"FP8_DYNAMIC", "BF16"}) is False
    assert _formats_need_static_activation_max(set()) is False
```

`tests/test_served_activation_contract.py` and `tests/test_tessera_formats.py` are green on main (`77 passed`), so nothing in the suite covers this input.

## Severity

P1. A hard crash — not even a refusal — of the production weight cache fill for a legal, shipped, selectable format, introduced by a change whose stated scope was which activation quantiser a row is *scored* under. Served-availability regression per the rubric; not wrong bytes, because nothing is produced at all.

## Fix

Two independent things are wrong and both should be fixed:

1. **Do not upper-case a registry name.** `canonical_format_name` (`format_registry.py:330-341`) already tries the raw string, then the upper-cased one, precisely so mixed-case rows resolve. `_format_uses_static_activation_clip` should pass `str(fmt).strip()` and let the resolver do its job; the `.upper()` was harmless when the result was only compared to `"NVFP4"` and is actively wrong now that it is looked up. `_render_base_format`'s `.upper()` is used for other purposes and is a wider change — leaving it alone is fine once the predicate stops upper-casing again, because `canonical_format_name("INT4_W4A16_G128")` still fails; so `_render_base_format` needs the same treatment, or the gate should be handed `fmt_set` rather than `render_base_fmt_set`.
2. **Make the predicate total again.** It answers a yes/no question about a format menu that may contain names this registry does not own; an unresolvable name is not served under a static activation contract, so `False` is the correct answer, not an exception. Wrap it the way `_is_cb_format_name` (`:1782-1787`) already wraps the same call. If the project would rather refuse an unknown format name, that refusal belongs where the menu is validated, not in a predicate about activation scales — and it must then be a named refusal, not a bare `KeyError` naming the registry.

Add the regression to `tests/test_served_activation_contract.py:195`: `assert _formats_need_static_activation_max({"INT4_W4A16_g128"}) is False`, and a case for an unregistered name.

## Provenance

PR #213 (`claude/pq-205`, merge `473cf6ca`, fixes #205). The rest of #213 is sound — I verified the name-compare-to-contract switch is exactly equivalent over all 80 registry rows (no row's classification changes), that stock NVFP4's screen path is numerically unchanged (the old `_nvfp4_input_global_scale_from_max_abs` and the contract's `input_global_scale_from_max_abs` differ only in a `nonpositive_fallback` that the `max_abs > 0` guard makes unreachable), and that `_render_score_record`'s own `get_format` call resolves mixed-case names correctly because it does *not* upper-case.

Auditor, file-only by brief. Original issue: #205 — this is a regression in its fix, not in the behaviour it set out to change.

RobTand:
Fixes #218 (P1).

## The crash, and what was actually wrong

PR #213 replaced a string membership test on the production-weight-cache fill
path with a registry lookup, and a lookup can raise where a membership test
could only answer. `_format_uses_static_activation_clip` upper-cased a name
`_render_base_format` (`:168`) had already upper-cased and called
`fr.get_format`, so any fill whose menu or assignment carried
`INT4_W4A16_g128` died with an unhandled `KeyError` at
`fill_production_weight_cache` before a single tensor was rendered. That name
is reachable: `validation_harness._entry_format_name` (`:807`, `:843`) maps
the 4-bit entry of a precision plan to it, and `render_scope == "assignment"`
puts it straight into `render_formats_by_qname`
(`streaming_production_cache.py:1449-1453`).

**What the enumeration says.** Sweeping all 80 registry rows + 3 aliases
through `_render_base_format` then `canonical_format_name`: exactly ONE name
fails to survive the round trip — `INT4_W4A16_g128`, the only mixed-case
registered row. No *other* mixed-case registered format is silently broken by
the upper-casing, no alias is, no Tessera rung is (that grammar is upper by
construction), and no name resolves to a *different* spec after an upper- or
lower-case round trip.

**And the gate was not the only crash on this path.**
`render_production_weight` normalizes with its own `.upper()` at `:4416`, and
that line predates #213 — `INT4_W4A16_g128` has never rendered through this
entry point. #213 moved the `KeyError` earlier rather than introducing the
whole defect, so fixing only the gate would have moved it back, not removed
it. (The issue's "previously ... rendered normally" is not accurate; the
regression is real, but the format was already unrenderable.)

## The fix: where the rule lives, not at each caller

1. **`format_registry.canonical_format_name` gains a case-insensitive last
   probe.** Case is not part of a format's identity, and this function is the
   one place that decides it. The probe is last, so it fires only where the
   exact probes already failed: nothing that resolved before resolves
   differently, and an unknown name is still returned unchanged for
   `get_format` to refuse by name. No two registered names differ only by
   case — now pinned as the precondition. Every caller that normalizes by
   upper-casing (`_render_base_format`, the cache identity keys,
   `render_production_weight`) keeps its spelling and resolves again, so no
   cache filename, manifest key or render-score record changes.

2. **`production_weight_cache._resolve_format_spec` is ONE home** for "does
   this requested name resolve to a spec, and if so which". That question had
   two answers in the file: `_is_cb_format_name`'s `try/except -> False` and
   the new predicate's bare `get_format`. Now
   `_format_uses_static_activation_clip`, `_is_cb_format_name` and
   `_weighted_render_family` all call it. It does not upper-case (the registry
   settles case; these predicates are handed already-normalized names, so they
   resolve the name they are given) and answers `None` rather than raising,
   because a predicate over a requested format MENU must be total: "not a
   format this registry owns" is an answer, not an exception thrown out of a
   question about activation scales mid-fill. A refusal for an unknown format
   belongs where the menu is validated, named.

`_render_base_format` deliberately keeps its `.upper()`, with a comment
recording why and what the enumeration found.

No default, format menu, serving lane or ship gate moves. What any resolvable
format resolves to is unchanged; `INT4_W4A16_g128` stays
research/registry-only. `docs/ARCHITECTURE.md` §5.1 and its provenance block
are re-stamped in the same commit.

## Regressions (shown failing on `553de0af` first)

- `tests/test_served_activation_contract.py::test_the_fill_gate_answers_for_the_mixed_case_registry_row`
  — drives the gate with `{_render_base_format("INT4_W4A16_g128"), "BF16"}`,
  the set the fill actually builds at `:6790`, and asserts "no static
  activation max needed" instead of a raise.
- `tests/test_served_activation_contract.py::test_the_fill_gate_is_total_over_names_the_registry_does_not_own`
  — an unresolvable junk name answers `False` from both predicates.
- `tests/test_format_registry.py::test_every_registered_name_resolves_through_an_upper_cased_spelling`
  — the whole registry enumerated through an upper/lower round trip, plus the
  no-two-names-differ-only-by-case precondition, plus an unknown name still
  raising.
- `tests/test_production_weight_cache.py::test_the_mixed_case_registry_row_renders_through_the_cache_entry_point`
  — the render entry point, so the fix removes the crash rather than moving
  it.

## Also

Prose fix on sight (AGENTS.md principle 14): ARCHITECTURE.md §5.1 cited
`format_registry.py:170-188` for `FORMAT_ALIASES`, which lives at `:298-308`;
`:170-188` is inside `FormatSpec`.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,35 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-05 · `claude/pq-204`. Stamps
+As of: 2026-09-05 · `claude/pq-218`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-05, `claude/pq-218`) for **format-name resolution: case is
+the registry's question, and a predicate over a format menu must be total**
+(§5.1; RobTand/prismaquant#218, P1 — a regression in #213/#205). `#213`
+replaced a string membership test on the production-cache fill path with a
+registry lookup, and a lookup can raise where a membership test could only
+answer: `_format_uses_static_activation_clip` upper-cased an already
+upper-cased name and called `get_format`, so any fill whose menu or assignment
+carried `INT4_W4A16_g128` — the one mixed-case registry row, and what
+`validation_harness._entry_format_name` maps a 4-bit precision-plan entry to —
+died with an unhandled `KeyError` at `fill_production_weight_cache` before a
+tensor was rendered. Enumerated over the whole registry (80 rows + 3 aliases),
+that name is the **only** one `_render_base_format`'s `.upper()` makes
+unresolvable, and no name resolves to a *different* spec after an upper- or
+lower-case round trip. The fix is therefore not at the caller: (1)
+`format_registry.canonical_format_name` gains a case-insensitive last probe,
+so every caller that normalizes by upper-casing — the render base, the cache
+identity keys, `render_production_weight` (whose own `.upper()` predates #213
+and crashed on the same name at render time, so fixing only the gate would
+have moved the `KeyError` rather than removed it) — resolves again; identity
+spellings are unchanged, and nothing that resolved before resolves
+differently. (2) `production_weight_cache._resolve_format_spec` is one home
+for "does this requested name resolve to a spec, and if so which", shared by
+`_format_uses_static_activation_clip`, `_is_cb_format_name` and
+`_weighted_render_family`; it does not upper-case (the registry settles case)
+and answers `None` rather than raising, so an unresolvable menu entry answers
+`False`. No default, menu, lane or gate moves; `INT4_W4A16_g128` remains
+research/registry-only.
 
 Re-stamped (2026-09-05, `claude/pq-204`) for **binding the Tessera export
 inputs to the payload and the values that priced the allocation** (§5;
@@ -6131,13 +6159,34 @@ shape-exact rather than a nominal scalar:
 and 3-D packed-expert stacks; `memory_bytes_for_shape` / `effective_bits_for_shape`
 (`:157-168`) are what the DP, `footprint.py`, and the Pareto table consume for
 those formats. Aliases:
-`FP8`/`FP8_DYNAMIC` → `FP8_E4M3`, `MXFP8` → `MXFP8_E4M3` (`:170-188`).
+`FP8`/`FP8_DYNAMIC` → `FP8_E4M3`, `MXFP8` → `MXFP8_E4M3` (`:298-308`).
 `act_quant_changes_input` (`:75-106`) is the **single** predicate for "does the serving kernel
 consume quantized activations" (`act_bits` absent or ≥ 16 ⇒ no): the allocator's bit-exact
 short-circuit (§4.5), the KL validator's activation-quant assignment, `layer_state_cache` and
 `perturbed_x_cache` all key off it, so a format's activation semantics cannot drift between
 pricing and emulation. Registry-vs-callable consistency is pinned by
 `tests/test_bit_exact_cost_pricing.py`.
+
+**Name resolution is the registry's job, and case is not part of a format's
+identity.** `canonical_format_name` (`:330-369`) probes the raw spelling, then
+the upper-cased one, then case-insensitively; `get_format` refuses only a name
+no probe resolves (falling through to the Tessera synthesizer first, §5.7).
+The last probe exists because exactly one registered row is mixed case —
+`INT4_W4A16_g128` — while several callers normalize a requested format name by
+upper-casing it (`production_weight_cache._render_base_format`, the cache
+identity keys, `render_production_weight`), which produced a name no resolver
+owned and a bare `KeyError` on a format PrismaQuant ships and a precision plan
+selects (#218). Callers keep their upper-case identity spelling; the registry
+answers to it. Consequently a **predicate over a requested format menu must be
+total**: `production_weight_cache._resolve_format_spec` is the one resolver
+those predicates share (`_format_uses_static_activation_clip`,
+`_is_cb_format_name`, `_weighted_render_family`), and it answers `None` — not
+an exception — for a name this registry does not own, because "not a format we
+know" is an answer. A refusal for an unknown format belongs where the menu is
+validated, named, not thrown out of an activation-scale gate mid-fill. Pinned
+by `tests/test_format_registry.py` (whole-registry upper/lower round trip, and
+the no-two-names-differ-only-by-case precondition) and
+`tests/test_served_activation_contract.py`.
 
 | Format | line | w-bits / group / scale | eff. bpp (2-D) | Status |
 |---|---|---|---|---|
--- a/prismaquant/format_registry.py
+++ b/prismaquant/format_registry.py
@@ -328,6 +328,27 @@ def register_format(spec: FormatSpec) -> FormatSpec:
 
 
 def canonical_format_name(name: str) -> str:
+    """The registry's own spelling of a requested format name.
+
+    Case is not part of a format's identity, and this function is where that
+    is decided -- once, for every consumer.  The raw-then-upper probes below
+    cover "the caller typed lower case and the row is upper case"; they did
+    NOT cover the mirror image, and exactly one registered row is mixed case
+    (``INT4_W4A16_g128``).  So every caller that normalizes a requested name
+    by upper-casing it -- ``production_weight_cache._render_base_format``, the
+    cache identity keys, ``render_production_weight`` -- handed the resolver a
+    name no resolver owned, and ``get_format`` raised ``KeyError`` on a format
+    PrismaQuant ships and a precision plan can select (#218).
+
+    The case-insensitive probe is last, so it fires only where the exact
+    probes have already failed: nothing that resolved before resolves
+    differently now, and an unknown name is still returned unchanged for
+    ``get_format`` to refuse by name.  No two registered names differ only by
+    case (pinned by ``tests/test_format_registry.py``), so the answer is
+    unambiguous.  It is a linear scan over ~80 rows on the MISS path only,
+    which is also the path every synthesized Tessera rung takes; that is
+    string comparison against a fixed-size table, not a hot-loop cost.
+    """
     raw = str(name).strip()
     if raw in FORMAT_ALIASES:
         return FORMAT_ALIASES[raw]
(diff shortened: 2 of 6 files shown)


RobTand:
## Commands and counts

All test runs on **sparky**, CPU only, worktree synced from
`/home/rob/prismaquant-audit-wt/pq-218`. Tip `9428e6c5`, branched from `main`
at `553de0af` (`git merge --no-edit origin/main`: already up to date).

**Pre-fix, on `553de0af` — the three added regressions, before any source
change:**

```
  python -m pytest -p no:cacheprovider -q -n 4 \
  tests/test_served_activation_contract.py tests/test_format_registry.py
=> 3 failed, 26 passed

E  KeyError: "Unknown format 'INT4_W4A16_G128'. Available: [... 'INT4_W4A16_g128' ...]"
   prismaquant/production_weight_cache.py:1561  (_format_uses_static_activation_clip)
   prismaquant/format_registry.py:1466          (get_format)
```

Note the registry's own error text lists `INT4_W4A16_g128` among the available
formats while refusing `INT4_W4A16_G128` — the whole defect in one line.

**Post-fix, the full brief'd set:**

```
  python -m pytest -p no:cacheprovider -q -n 4 \
  tests/test_production_weight_cache.py tests/test_streaming_production_cache.py \
  tests/test_format_registry.py tests/test_validation_harness.py \
  tests/test_served_activation_contract.py tests/test_format_menu_expansion.py \
  tests/test_m4_frontier_expert_selection.py tests/test_mxfp8_ue8m0.py \
  tests/test_prismaquant_export_native_compressed.py \
  tests/test_prismaquant_native_math.py tests/test_serving_profiles.py \
  tests/test_tessera_candidate_context.py \
=> 408 passed, 4 skipped, 163 subtests passed
```

That set is the brief's `test_production_weight_cache*` /
`test_streaming_production_cache*` / `test_format_registry*` /
`test_validation_harness*` globs, plus every test file that greps
`_format_uses_static_activation_clip`, `_render_base_format`,
`_is_cb_format_name` or `canonical_format_name`, plus the two doc gates.

**Full suite** (`canonical_format_name` is shared, so the whole tree was run):

```
python -m pytest -p no:cacheprovider -q -n 4 tests/
=> 17 failed, 4962 passed, 130 skipped, 3 xfailed, 174 subtests passed  (9m44s)
```

The 17 failures are exactly the known-inherited
`tests/test_tessera_campaign_resume.py` main-entry set —
`test_main_resumes_identical_cost_and_wire_without_reencoding`,
`test_main_refuses_changed_hessian_values_under_same_draw`,
`test_main_refuses_changed_encoding_or_scoring_inputs[...]` (14 params),
`test_main_refuses_missing_or_changed_priced_wire[...]` (3 params). Count and
identity unchanged.

**The registry enumeration behind the claim in the commit message:**

```
CUDA_VISIBLE_DEVICES="" PYTHONPATH=/home/rob/tessera-pin-1221d2a/src python - <<'PY'
from prismaquant import format_registry as fr
from prismaquant.production_weight_cache import _render_base_format
for n in sorted({*fr.REGISTRY, *fr.FORMAT_ALIASES}):
    ...  # compare get_format(n) against get_format(canonical(_render_base_format(n)))
PY

registry rows: 80  aliases: 3
names that do NOT survive _render_base_format + canonical_format_name:
  ('INT4_W4A16_g128', 'INT4_W4A16_G128', 'INT4_W4A16_G128', 'INT4_W4A16_g128', 'RAISE KeyError')
collisions (upper-cased name maps to a DIFFERENT resolvable spec): (none)
mixed-case REGISTRY rows: ['INT4_W4A16_g128']
mixed-case aliases: []
```

So: exactly one registered name does not survive `_render_base_format`, no
other one is silently broken, and no resolvable format changes what it
resolves to. That enumeration is now a test
(`test_every_registered_name_resolves_through_an_upper_cased_spelling`), which
also pins the precondition that makes case-insensitive resolution unambiguous.

**Environment.** `python` above is
`/home/rob/venvs/pq-release/bin/python` (3.12.3). No GPU, no synthetic load,
no polling loops. `free -g` before the full-suite run: 121 GB total, 110 GB
available.

Issue #204, opened 2026-09-05
[P0] Bind Tessera export inputs to the actual Hessian and static scales that priced the allocation


RobTand:
The new priced-input preflight accepts different Hessian and static activation-scale values for one allocation. The original missing-argument defect in #193 is fixed by #196; this follow-up is about the contents of the now-forwarded files, after every required argument/key is present.

Initially audited PrismaQuant `65cfa6d9ee53cec7880c5320fd44c395b9406a3b`; both findings were rechecked after merges #187/#191/#200 at current main `468e7534bad23e4c02454f44eb5517e9a6a39f4d`. The source anchors in this issue name `468e7534bad23e4c02454f44eb5517e9a6a39f4d`. Producer contract comparison uses the frozen Tessera snapshot `4e9744f6424ecd0d90ebcbd4200c909123f38a77` and the clean checkout at PrismaQuant's intended dev pin `1221d2a4207a6baeffbe9726bce13125fc1649ae`. The same refusal gaps reproduce with both producers, so this is independent of the deliberately refused latest v8 lane schema.

There are two paths through the same `require_priced_export_inputs` gate:

- `_hessian_capture_identity` returns `<capture>.provenance.json` immediately at `prismaquant/tessera_export_lane.py:501-508`, without binding it to the actual `.pt` payload. Lines 601-610 compare the allocation's identity triple only to that sidecar. Tessera's `ActivationSource.from_capture` reads the actual payload at `src/tessera/export.py:747` and uses its `H` and provenance at lines 772-774; it never reads PrismaQuant's sidecar. A payload from capture B with capture A's sidecar therefore passes an allocation priced under A, while the producer receives B. This can result from a copied/replaced capture or an interrupted campaign rewrite: `tessera_campaign.py:1070-1074` replaces the capture before writing its sidecar. Even with synchronized sidecar/payload provenance, equal token triples do not bind equal Hessian values or sequence/model context.
[...]

PR #223, opened 2026-09-05
Write the export leg's inputs after the resume identity, not before (#211)

RobTand:
`tessera_campaign.main()` called `write_export_inputs(...)` immediately after
collecting activations — before `prepare_journal(...)` and before the per-row
resume verification loop. So a resume the checkpoint identity **refuses** had
already overwritten `<cache-dir>/hessian_capture.pt`, its `.provenance.json`
sidecar and `<cache-dir>/input_scales.safetensors` with the refused run's
Hessians and static scales. The checkpoint and the previous cost file correctly
survive that refusal; the export half of the same surviving table did not, so
the table that survived could not be exported at all. #204 bounded the damage —
the old allocation now refuses the rewritten capture by `capture_sha256` rather
than exporting under the wrong Hessian and scale — but did not stop the
destruction.

**The fix is the move the issue prescribes.** The call now sits just after the
resume verification loop and its `[campaign] resumed …` print, still before
`started = time.time()`, so PR #193's guarantee holds unchanged: written before
the anchor loop, so even a deadline-stopped campaign leaves them.

**Read against #212 first, as briefed.** #212 (issue #204) landed in this same
file a few minutes before this branch opened and changed `write_export_inputs`
to a three-value return (`+ capture_sha256`), staged both files with the sidecar
renamed last, and added the digest to `campaign_cost_payload` and to
`provenance.hessian`. None of that makes the move unnecessary or changes its
shape — #212 is a *detection* change (the export gate now refuses the rewritten
capture by name) and this is the *destruction* it detects. Confirmed before
moving anything: nothing between the old and new call sites consumes
`hessian_capture_path`, `input_scales_path` or `capture_sha256` (first consumer
is still the provenance block at the end of `main()`); `write_export_inputs`
does not mutate `hessians`, `static_scales` or `hessian_identity`; and
`_campaign_checkpoint_identity` takes the Hessians and the calibration identity,
never the digest. An accepted resume rewrites byte-identical files, because the
run-level identity binds the Hessians, static scales and scale policy that are
exactly this call's inputs. The diff also un-mangles the continuation
indentation #212 left at the old call site.

No file format, field, default, gate or ordering other than this one moves.
Out of pq-183's region: `_collect_activations` and the packed-expert paths are
untouched; this is `main()`'s call ordering only.

## Regression

`test_tessera_campaign_resume.py::test_refused_resume_leaves_the_surviving_tables_export_inputs`
[...]

PR #220, opened 2026-09-05
Tessera packed-expert bridge: price it, carry it, export it (#183, items 1-3)

RobTand:
# The packed-expert bridge: price it, carry it, export it (#183)

Closes acceptance items **1–3** of RobTand/prismaquant#183. Item **4** (a bounded
actual MoE campaign + export/serve check) is **not** in this PR and is not
claimed — see *What is not claimed* below.

The producer owns the projection; PrismaQuant never derives one from names
outside the owning profile and never selects a 2-D slice of a packed source
tensor. Six commits, one thing each.

## The chain

| # | Commit | What it does |
|---|--------|--------------|
| 1 | `e16cb643` | `prismaquant/tessera_expert_projection.py` — the ONE reader of the producer's `tessera.expert_projection.v1` answer. Binds records to the profile split exactly (schema, `unpacked_per_expert` layout, whole-tensor selector, hashed source roster, `[rows, cols]`, full expert coverage) and refuses anything else by name. Declares `experiments/tessera_producer_plan.py` as the lane's third `producer_tools` entry. |
| 2 | `1b17af61` | The producer states the expert **count**, not a list of indices (`plan_expert_stack` has already refused a gap or an undeclared index against `config.json`). The binder now requires an integer count whose units are exactly `range(count)`. |
| 3 | `e52cf36f` | The campaign **prices** the producer-projected packed population. `_require_campaign_population` admits the in-scope packed parameters the profile splits *and* for which the producer's tool is present; everything else in scope is refused by name and the stride's leftovers are recorded as omitted. ONE producer request per campaign; every unit's source tensor read from the shard the producer hashed and compared byte-for-byte with the live view; anchors grouped per producer stack; receipts sealed with `unit_input_identity`; **measured-only** rows for wire-backed units; the payload says what it priced and what it omitted. |
| 4 | `5227837f` | The **allocation** carries what it selected from: the `population` block verbatim, the producer's projection, the receipt of exactly the rung selected for every projected unit, one format per executed stack, and the wire directory. Additive — a stock cost table adds no keys. |
| 5 | `cb2dcf5e` | Test-only repair of a red left by commit 1: the third declared producer tool is run by the **campaign**, not by the export arm. Counted as invoked via `PRODUCER_PLAN_TOOL`, which resolves *through* the declaration, so the "no second spelling" property still has teeth. |
| 6 | `82f05667` | The **export lane** hands the exporter the priced bytes. Detailed below. |

## The last leg (`82f05667`)

[...]

PR #225, opened 2026-09-05
Read the capture-context roster from Tessera, widen the dead-override census, stop a test leaking its orphan (#216, #217, #219)

### LINKED ISSUE #216: Export-seal capture context is a hard-coded roster: CAPTURE_CONTEXT_FIELDS is not derived from or pinned to tessera.export.CAPTURE_CONTEXT
## What

`prismaquant/tessera_export_lane.py:494` hard-codes half of Tessera's capture-seal roster and nothing pins it to the owner:

```python
#: The capture-context fields Tessera's seal covers beside the triple
#: (``tessera.export.CAPTURE_CONTEXT``): two captures of one token prefix
#: differ only here when the sequence layout differs (tessera#214).
CAPTURE_CONTEXT_FIELDS = ("model", "seqlen", "source")
```

The sibling roster one screen up is derived from the owner and pinned by a test:

* `prismaquant/tessera_hessian.py:66-76` — `HESSIAN_IDENTITY_FIELDS = tuple(tessera.export.HESSIAN_IDENTITY)`, "read from Tessera rather than typed".
* `tests/test_tessera_priced_export_inputs.py:440` `test_the_priced_input_triple_matches_tesseras_roster` asserts `set(export.PRICED_HESSIAN_IDENTITY_FIELDS) == set(HESSIAN_IDENTITY_FIELDS)`.

`CAPTURE_CONTEXT_FIELDS` has no equivalent. `grep -rn CAPTURE_CONTEXT --include=*.py` in the branch returns three hits, all in `tessera_export_lane.py` (the comment, the tuple, and its one use at `:539`). Nothing reads `tessera.export.CAPTURE_CONTEXT`.

`hessian_capture_sha256` (`tessera_export_lane.py:506`) is a reimplementation of `tessera.export.ActivationSource.capture_sha256` and must digest identically or the export gate refuses correct captures / accepts wrong ones. Today it does — verified byte for byte, see below. If Tessera ever adds a field to `CAPTURE_CONTEXT` (the constant exists precisely so it can grow — `tessera/src/tessera/export.py:239-244` describes it as "ride-along provenance sealed into `capture_sha256` beside the content"), the two rules diverge, and the divergence lands differently depending on the pin:

* **At the current dev pin (1221d2a)**, `tessera.export.ActivationSource` has no `capture_sha256`, so `_tessera_capture_seal` (`:557`) returns `None` and `_crosscheck_capture_seal` (`:624`) returns without comparing. The drift is silent: two captures that differ *only* in the new context field digest identically, so `require_priced_export_inputs` binds an allocation to a capture that did not price it — the exact failure #204 was opened to close, reintroduced with no signal.
* **At a Tessera that publishes the seal**, the runtime cross-check fires and `--hessian` refuses with "the two digest rules have drifted and no allocation can be bound against this Tessera until they agree" — the whole Tessera export arm is down until someone edits PQ.

No test catches either state before it ships: `test_the_digest_rule_is_tesseras_capture_seal` (`tests/test_tessera_priced_export_inputs.py:288`) is the only equality check against Tessera's own rule, and it `pytest.skip`s at the pinned dev Tessera (confirmed below). `test_the_digest_rule_covers_content_and_capture_context` (`:266`) only asserts PQ's rule against itself.

## Evidence

Digests agree today at the release tip (7 payload shapes, including absent context fields, non-str context values through `default=str`, mixed dtypes, non-contiguous H, and `.weight`-suffixed unit names):

```
$ CUDA_VISIBLE_DEVICES="" PYTHONPATH=/home/rob/tessera-release/src python seal_repro.py
plain-f32      match=True   de879b1fb1f667ca99a84a0e2b9ec29c7c7623e717e856293f146a01f0c4c82b
mixed-dtype    match=True   2bc29d1920ecb82ced1ee61c05f89043db2284ce7cc339e52f11d4f4f0e410c0
no-context     match=True   9b20d7ca17462327b16967e771d81fed3d976edeb1e561b23c29e4b31b22f26d
weird-seqlen   match=True   11217a3bac75adc47f1b0032606ad66d3ea84ba8d88fa26e180479365a15d0b1
names          match=True   c1e68e95183be0ba0ef812453ec419c259cabd4c6dbc96748cab7bf4d369c20a
extra-prov     match=True   2dc4a072af789f129f25c6ab241606f803e7209ba4f3a4c65e6e357ae07c2085
noncontig      match=True   70b155ea7a18680196e577475ac53bedb168ddf7330f979cc3f89f29bf998573
ALL MATCH: True
```

(`tessera/src/tessera/export.py` lines 775-815 are identical between Tessera master `3efd690` and the release tip, so this holds for both.)

The equality test never runs at the pin PQ actually uses:

```
$ CUDA_VISIBLE_DEVICES="" PYTHONPATH=/home/rob/tessera-pin-1221d2a/src \
    python -m pytest tests/test_tessera_priced_export_inputs.py -p no:cacheprovider -q -rs
SKIPPED [1] tests/test_tessera_priced_export_inputs.py:296: pinned tessera has no ActivationSource.capture_sha256
39 passed, 1 skipped
```

## Severity

P2. Not wrong bytes today — the two rules agree, and I proved it. It is a missing drift guard on a P0 binding whose whole point is that a name is not content: the identity triple is guarded against roster drift and the capture context, which the same digest covers, is not. AGENTS.md's "derive rosters from owning code" applies to both halves of one roster.

## Fix

Extend the existing pin rather than adding a second mechanism — one line in `test_the_priced_input_triple_matches_tesseras_roster` (or a sibling next to it):

```python
from tessera.export import CAPTURE_CONTEXT
assert tuple(export.CAPTURE_CONTEXT_FIELDS) == tuple(CAPTURE_CONTEXT)
```

Order matters here, not just membership: the fields go into one `dict` that is `json.dumps(..., sort_keys=True)`-ed, so a set comparison would be enough for the digest, but a tuple comparison also catches a rename. If the test must stay importable without `tessera`, gate it the way `test_the_digest_rule_is_tesseras_capture_seal` gates itself (`pytest.importorskip("tessera.export")`) — accepting that it, too, then only runs at a pin that has the constant, which is an argument for landing it together with #192's move to the release pin.

## Provenance

Introduced by PR #212 (`claude/pq-204`, fixes #204), branch tip `fc825d83`; `tessera_export_lane.py:489-546`. The rest of #212's seal work is sound — the digest construction reproduces `tessera.export.ActivationSource.capture_sha256` byte for byte (evidence above), the sidecar is staged and renamed last as its docstring claims, and the F32 scale comparison round-trips through the same `struct.pack("<f", ...)` the campaign's `input_global_scale_tensor` uses (`nvfp4_activation_contract.py:284-292`), so the two cannot disagree on representation.

Auditor, file-only by brief. Original issue: #204 (this is an incompleteness in its fix, not a regression).

### LINKED ISSUE #217: The dead-override census cannot see an 'except RuntimeError' swallow, which is the shape DeadVendoredOverrideError's own docstring invites
## What

The tree-wide census that pins "a dead vendored override must never be *answered*" cannot see three handler shapes that answer it — including the one the exception class's own docstring invites.

`tests/test_vendored_override_gate.py:615`:

```python
def _handler_is_broad(handler: ast.ExceptHandler) -> bool:
    """True for `except:`, `except Exception:`, `except BaseException:`."""
    node = handler.type
    if node is None:
        return True
    parts = node.elts if isinstance(node, ast.Tuple) else [node]
    return any(
        isinstance(p, ast.Name) and p.id in ("Exception", "BaseException")
        for p in parts
    )
```

A handler that is not "broad" by this definition and does not name `DeadVendoredOverrideError` falls out of `_refuses_dead_override` (`:639`) at its final `return True  # no broad handler at all: nothing swallows it` (`:652`) — so the site is reported as refusing.

But `DeadVendoredOverrideError` **is a `RuntimeError`**, deliberately. `prismaquant/model_profiles/registry.py:98,114`:

```python
class DeadVendoredOverrideError(RuntimeError):
    ...
    Subclasses `RuntimeError` so existing `except RuntimeError` handlers and
    the gate's original contract are unchanged.
```

So `except RuntimeError: profile = None` around a detection call swallows the refusal exactly as `except Exception:` did, and the census says the tree is clean. That shape is not contrived: `detect_profile` itself raises a bare `RuntimeError` for a non-directory `model_path` (`registry.py:169-172`), so a caller narrowing `except Exception` to `except RuntimeError` — the ordinary way to tighten one of these handlers — silently reopens #202 at that site and passes CI.

`_handler_always_reraises` (`:627`) has a second, smaller hole: it tests only that the handler's **last** statement is a `Raise`, so a handler that returns a substituted answer on one branch and re-raises on the fall-through is accepted.

Nothing today — I checked the whole tree and no detection call site currently uses any of these shapes (below). The defect is in the guarantee, not the current state. The PR body for #209 states the census "fails for the 23rd such call site when someone writes it, which no per-site test does", and the test's own docstring repeats it (`:695-698`). That is the census's entire justification for covering the four sites that have no behavioural test — `aqua_activation_cost.py:661`, `build_rtn_cache.py:500`, `sensitivity_probe.py:3592`, `streaming_production_cache.py:1776`, which the docstring names as unreachable without a GPU, a loaded checkpoint or a built streaming context. For those four the census is the only coverage there is, and it is evadable by a handler a reasonable author would write.

## Evidence

Feeding synthetic call sites to the census's own predicate, imported from the test module unmodified:

```
$ python census_probe.py       # imports _refuses_dead_override from tests/test_vendored_override_gate.py
A: except RuntimeError (DeadVendoredOverrideError IS a RuntimeError)
    -> _refuses_dead_override = True  =>  PASSES CENSUS (not flagged)
B: except (ImportError, RuntimeError)
    -> _refuses_dead_override = True  =>  PASSES CENSUS (not flagged)
C: except builtins.Exception (dotted)
    -> _refuses_dead_override = True  =>  PASSES CENSUS (not flagged)
D: broad handler that returns early, then raises
    -> _refuses_dead_override = True  =>  PASSES CENSUS (not flagged)
E: the real swallow shape the census is meant to catch (control)
    -> _refuses_dead_override = False  =>  flagged as swallowing
```

Case E is the control: the census does catch the shape #202 actually removed.

Current tree is clean — an independent AST walk of `prismaquant/` finds 32 detection calls inside a `try`, and every handler is one of `DeadVendoredOverrideError` + `Exception` (28 sites), `SampleParallelProbeError` + `Exception` (`sample_parallel_probe.py:568`, which always re-raises), `TesseraExportLaneError` + `(OSError, ValueError, TypeError, KeyError, struct.error)` (`tessera_export_lane.py:428` — narrow, so the refusal propagates correctly), or no handler at all (`try/finally`).

The rest of #209 is genuinely load-bearing, so this is a gap in a good fix, not a bad one. All 19 new tests fail against the pre-fix tree — I ran the merged test file against `cdaf8931^1`:

```
$ cp tests/test_vendored_override_gate.py <worktree at cdaf8931^1>/tests/
$ CUDA_VISIBLE_DEVICES="" PYTHONPATH=/home/rob/tessera-pin-1221d2a/src \
    python -m pytest tests/test_vendored_override_gate.py -p no:cacheprovider -q
19 failed, 12 passed
```
(the 19 are the 18 behavioural tests plus `test_no_call_site_swallows_the_dead_override_refusal`; the 12 that pass are #201's, from PR #203). Against the merged tree: `31 passed`.

## Severity

P3. No wrong bytes and no live swallow; a structural pin that does not hold the shape it claims to hold, protecting four sites that have nothing else.

## Fix

In `_handler_is_broad`, treat a handler as broad when it names any ancestor of `DeadVendoredOverrideError` rather than a hard-coded pair — the roster belongs to the class, so derive it:

```python
_CATCHES_DEAD = {c.__name__ for c in DeadVendoredOverrideError.__mro__}
# {'DeadVendoredOverrideError', 'RuntimeError', 'Exception', 'BaseException', 'object'}
```

and match `getattr(p, "id", None) or getattr(p, "attr", None)` against it, which closes A, B and C together (C needs the `attr` fallback the named-clause loop at `:645` already uses). That also keeps the census correct if the class is ever re-parented. For `_handler_always_reraises`, require that no statement in the handler body is a `Return`/`Break`/`Continue` and that it ends in a `Raise`, or simply that the handler contains no `Return` — the one shape it exists to allow (`sample_parallel_probe.py:568`) satisfies either.

Worth a line in the test docstring naming what the census does NOT cover, since it is cited as future-proofing.

## Provenance

PR #209 (`claude/pq-202-detection-swallows`, merge `cdaf8931`, closes #202). `tests/test_vendored_override_gate.py:615-652`.

Auditor, file-only by brief. Original issue: #202 — the 22 call-site edits themselves are correct and I confirmed each re-raises; this is the durability of the pin that is meant to keep them that way.

### LINKED ISSUE #219: Resume-after-coordinator-death test can leak a forever-spinning orphan: the release write is not in a finally, unlike its sibling test
## What

`tests/test_cluster_campaign_runner.py::test_resume_after_coordinator_death_recovers_owned_helper_receipt` can leave an orphaned Python process busy-waiting on the box forever. Its sibling test one function down guards against exactly this and says why; the guard was not applied here.

PR #206 replaced the helper's self-limiting `time.sleep(1.0)` with an **unbounded** wait on a release file (`tests/test_cluster_campaign_runner.py:678-684`):

```python
    script = (
        "from pathlib import Path; import sys, time; "
        "release=Path(sys.argv[2]); "
        "Path(sys.argv[3]).write_bytes(b'started'); "
        "exec(\"while not release.exists():\\n time.sleep(0.02)\"); "
        "Path(sys.argv[1]).write_bytes(b'slow-done')"
    )
```

The test then deliberately orphans that child — it kills the coordinator while the helper runs, and the file's own comment at `:776-777` says so: *"The helper was deliberately orphaned, so this test holds no handle on it: the backstop is the only bound available here."* With the coordinator and its worker dead, **nothing enforces the stage's `timeout_seconds`**, so the only thing that ever ends the helper is the test writing the release file.

That write is not on every path. The `try` guard stops at `:761-763`:

```python
    except BaseException:
        release_path.write_bytes(b"go")
        raise

    coordinator.terminate()
    _wait_until(
        lambda: coordinator.poll() is not None,
        unmet="coordinator did not exit after terminate()",
    )
    ...
    release_path.write_bytes(b"go")        # :778
```

Lines 765-777 are unguarded. Anything that raises there — most concretely the `_wait_until` at `:766-769`, which raises `AssertionError` when the coordinator has not exited within `_WEDGE_BACKSTOP_SECONDS` (120 s) of `terminate()`, or any interruption of the test session in that window — skips `:778`. The helper is already orphaned by `:765`, so it then polls `release.exists()` at 50 Hz for the lifetime of the machine. `tmp_path` being cleaned up later does not help: the release file can then never appear, which makes the leak permanent rather than transient.

## The same file already knows this

`test_abrupt_worker_death_leaves_stage_lock_owned_by_child`, added in the same PR with the same orphaned-child-on-a-release-file pattern, wraps it properly (`:891-895`):

```python
    finally:
        # Unblock the child on every path out of this test: a failure above
        # must not leave an orphan spinning on a release file that never lands.
        # Its worker is dead by then, so nothing else would enforce a bound.
        release_path.write_bytes(b"go")
```

That comment is a correct statement of the hazard and it applies verbatim to the resume test. Only one of the two got the `finally`.

A leaked, permanently resident `python -c` process on a shared box, one per occurrence, invisible to the test report (the test has already failed for its own reason, or been killed). It costs little CPU individually but never exits and cannot be attributed to a test run after the fact. On this project's boxes, where stray processes and load are an active operational problem, that is the whole cost of the defect.

Not hypothetical for the interruption path specifically: agents on this campaign were killed mid-run today by an API error, which is precisely the "test session dies between `:765` and `:778`" case.

## Severity

P3. Test-only, and only on an abnormal exit from a window of about a dozen lines — but the consequence outlives the run and the fix is four lines that the neighbouring test already has.

## Fix

Extend the existing guard to cover the orphaning, i.e. move the whole body after `Popen` into `try:` and release in `finally:`, matching `test_abrupt_worker_death_leaves_stage_lock_owned_by_child:891`:

```python
    try:
        ...                                   # the two waits, terminate(), the receipt wait
    finally:
        release_path.write_bytes(b"go")
```

Writing the release twice is harmless (`write_bytes` is idempotent here), so the straight-line write at `:778` can stay where it is — it is load-bearing for ordering, since the receipt wait at `:779` must not start before the helper is released.

Worth also considering whether an orphan-by-design helper should carry its own absolute backstop (`if time.monotonic() > start + N: sys.exit(9)`) rather than relying solely on the parent's good behaviour — the argument in the PR for removing the child-side deadline (`:378-382`) is that the worker enforces `timeout_seconds` by killing the process group, which I confirmed it does (`cluster_campaign.py:1164` `start_new_session=True`, `:1171-1178` `TimeoutExpired` → `_terminate_child_process_group`). That argument is sound for every stage child in the file **except** the two the tests deliberately orphan, where the enforcer is killed on purpose.

## Provenance

PR #206 (`claude/pq-199-deadline`, merge `6db64bbc`, closes #199), commit `7236f826` "tests: let the resume test's helper exist before the coordinator dies".

The rest of #206 is sound: it is a test-only change, the wall-clock deadlines it removes were genuinely load-sized (issue #199's measurements), `_wait_until` is bounded on both the producer's liveness and a 120 s backstop, `Popen.poll()` reaps so the replaced `wait(timeout=...)` calls leak nothing, and the removal of the child-side deadline in `test_scheduler_refills_a_freed_host_while_another_host_is_active` is safe because that child's worker is alive and does kill its process group.

Auditor, file-only by brief. Original issue: #199 — this is a hazard introduced by its fix, not a failure to fix it.

RobTand:
Three fixes filed by the audit pass, one commit each, on top of `553de0af`.

| | Issue | Sev | What changed |
|---|---|---|---|
| `b42beb01` | #216 | P2 | The export seal's capture-context roster is read from `tessera.export.CAPTURE_CONTEXT` instead of typed, with one documented fallback and a by-name refusal when the rosters disagree |
| `a33cb29c` | #217 | P3 | The dead-override census derives its catching class names from `DeadVendoredOverrideError.__mro__`, so `except RuntimeError` no longer hides a swallow |
| `26e63a16` | #219 | P3 | The resume-after-coordinator-death test can no longer leak the child it orphans on purpose, and now asserts it is gone |

Each commit shows its regression failing on `553de0af` first, carries the
pre-fix line, and re-stamps `docs/ARCHITECTURE.md` in the same commit
(principle 12).

## #216 — `CAPTURE_CONTEXT_FIELDS` had no owner

`tessera_export_lane.py:494` was the literal `("model", "seqlen", "source")`
and nothing in the tree read `tessera.export.CAPTURE_CONTEXT`, while its
sibling one screen up derives the identity triple from Tessera
(`tessera_hessian.HESSIAN_IDENTITY_FIELDS`) and a test pins the two together.
The rules agree today; a field added to Tessera's constant splits them, and at
a pin with no `ActivationSource.capture_sha256` (the dev pin `1221d2a`) the
split is silent — two captures differing only in the new field digest
identically and an allocation binds to a capture that did not price it, which
is #204's failure with no signal.

* `CAPTURE_CONTEXT_FIELDS, CAPTURE_CONTEXT_FIELDS_SOURCE =
  _capture_context_fields()` reads the constant where the running Tessera
  publishes it; otherwise `_CAPTURE_CONTEXT_FALLBACK` — the one place the
  roster is typed, commented with the Tessera it was copied from (0.1.0,
  release `3efd690`, `src/tessera/export.py:244`) — and it records which.
* `_require_capture_context_roster()`, called from `hessian_capture_sha256`
  before a byte is digested, re-reads the constant live and refuses by name
  when it is not the roster the digest covers. This is the half of the drift
  guard that needs no `capture_sha256`, so it holds at the dev pin; the
  campaign's `write_export_inputs` gets it through the same function.
* `test_the_capture_context_roster_is_tesseras` is the roster-equality pin the
  issue asked for, and it runs at **both** pins — a tuple comparison needs no
  seal. `test_the_digest_rule_is_tesseras_capture_seal` still skips at the dev
  pin for the runtime-seal comparison only.

No wire bytes, digest values, report keys or signatures change while the
rosters agree.

## #217 — the census could not see `except RuntimeError`

`DeadVendoredOverrideError` subclasses `RuntimeError` deliberately, so the
handler shape its own docstring invites swallowed the refusal and the census
reported the site as refusing. Four shapes were invisible: bare `RuntimeError`,
a tuple naming it, a dotted `builtins.Exception`, and a broad handler that
substitutes on one branch and re-raises on the fall-through.

`CATCHES_DEAD_OVERRIDE` now comes from `DeadVendoredOverrideError.__mro__`
(one rule, one home — re-parent the class and the census follows);
`_handler_catches_dead_override` matches bare, dotted and tuple clauses against
it; `_handler_always_reraises` additionally rejects a `return`, and a
`break`/`continue` that would carry control past the trailing `raise`.

**The widened census reports zero newly swallowing sites**, so no production
code changes: all 31 `try` blocks wrapping a detection call in `prismaquant/`
are still correct (25 `except DeadVendoredOverrideError: raise` then broad,
4 `try/finally`, 1 `SampleParallelProbeError` + broad-that-re-raises, 1 narrow
non-catching). The census's docstring now names what it does not cover, since
it is cited as future-proofing.

## #219 — an orphan-by-design helper that could outlive its test

The resume test kills the coordinator while a stage child polls a release file;
after #206 that wait is unbounded and the release write is the only thing that
can end it. The write sat outside the guard, so a failure in the ~dozen lines
after `coordinator.terminate()` left a `python -c` spinning at 50 Hz forever on
a file `tmp_path` cleanup then made unreachable.

`_orphaned_helper(release_path)` is now the one home for "the release always
happens": it releases and, if the child did not take the release, kills its
session on every path out. Both orphan-on-a-release-file tests use it, so the
sibling's hand-rolled `finally` and this test's `except BaseException` copy are
gone. The helper announces its PID, and `helper.assert_exited()` makes "this
test leaves nothing running" an assertion rather than an arrangement.

Deterministic pre/post demonstration: injecting an `AssertionError` into that
window on `553de0af` left the orphan behind (`LEAKED=2`, one of them the
scanner itself); the same injection on this tip leaves nothing (`LEAKED=1`,
only the scanner).

## Tests

sparklina, CPU only (`CUDA_VISIBLE_DEVICES=""`), real git worktree,
`/home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider`. Both
Tessera trees, since PQ's roster now depends on which one is importable.

Commands and counts are in the PR comment below.

Fixes #216
Fixes #217
Fixes #219

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,37 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-05 · `claude/pq-204`. Stamps
+As of: 2026-09-05 · `claude/pq-gates`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-05, `claude/pq-gates`) for **the export seal's
+capture-context roster read from Tessera rather than typed** (§5;
+RobTand/prismaquant#216, P2). `tessera_export_lane.CAPTURE_CONTEXT_FIELDS` was
+the literal `("model", "seqlen", "source")` and nothing in the tree read
+`tessera.export.CAPTURE_CONTEXT`, so half of the roster
+`ActivationSource.capture_sha256` seals had no owner — while its sibling
+`tessera_hessian.HESSIAN_IDENTITY_FIELDS` derives the identity triple from
+`tessera.export.HESSIAN_IDENTITY` and a test pins the two together. The two
+rules agree today; a field added to Tessera's constant would have split them,
+and at a pin with no `capture_sha256` (the dev pin 1221d2a) the split is
+silent: two captures differing only in the new field digest identically and an
+allocation binds to a capture that did not price it, which is #204's failure
+reintroduced with no signal. Now: `CAPTURE_CONTEXT_FIELDS,
+CAPTURE_CONTEXT_FIELDS_SOURCE = _capture_context_fields()` reads
+`tessera.export.CAPTURE_CONTEXT` where the running Tessera publishes it and
+otherwise falls back to `_CAPTURE_CONTEXT_FALLBACK` — the ONE place the roster
+is typed, commented with the Tessera it was copied from (0.1.0, release
+`3efd690`, `src/tessera/export.py:244`) — and names which was used.
+`_require_capture_context_roster`, called from `hessian_capture_sha256` before
+a byte is digested, re-reads the constant live and refuses by name when it is
+not the roster the digest covers, listing the fields each side names; this is
+the half of the drift guard that needs no `capture_sha256`, so it holds at the
+dev pin where `_crosscheck_capture_seal` compares nothing. No wire bytes, no
+digest values and no report keys change while the rosters agree. Tests:
+`test_tessera_priced_export_inputs.py` — `test_the_capture_context_roster_is_
+tesseras` (roster equality, runs at BOTH pins, unlike the skipped runtime-seal
+comparison), the drift refusal at the digest and at the gate, and the
+documented fallback (pre-fix at the release tip: `DID NOT RAISE
+TesseraExportLaneError`, the bind reaching the seal cross-check instead).
 
 Re-stamped (2026-09-05, `claude/pq-204`) for **binding the Tessera export
 inputs to the payload and the values that priced the allocation** (§5;
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,8 +3,37 @@
 As of: 2026-09-05 · `claude/pq-gates`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-05, `claude/pq-gates`) for **the dead-override census
+seeing every handler shape that catches its exception** (§8.1;
+RobTand/prismaquant#217, P3). `tests/test_vendored_override_gate.py`'s
+tree-wide pin that "a dead vendored override must never be *answered*"
+recognised only `except:`, `except Exception:` and `except BaseException:` —
+but `DeadVendoredOverrideError` subclasses `RuntimeError` deliberately, so
+`except RuntimeError:` swallowed the refusal and the census reported the site
+as refusing. Three shapes were invisible: bare `RuntimeError`, a tuple naming
+it, and a dotted `builtins.Exception`; a fourth, a handler that substitutes an
+answer on one branch and re-raises on the fall-through, passed
+`_handler_always_reraises`, which tested only the last statement. The census's
+whole justification is the four call sites with no behavioural test
+(`aqua_activation_cost.py:661`, `build_rtn_cache.py:500`,
+`sensitivity_probe.py:3592`, `streaming_production_cache.py:1776`). Now
+`CATCHES_DEAD_OVERRIDE` is derived from `DeadVendoredOverrideError.__mro__`
+rather than a second typed list, so the census follows the class if it is
+re-parented; `_handler_catches_dead_override` matches bare, dotted and
+tuple clauses against it; and `_handler_always_reraises` additionally requires
+no `return`, and no `break`/`continue` that would carry control past the
+trailing `raise`, ignoring nested scopes and loops written inside the handler.
+No production code changed: the widened census reports **zero** newly
+swallowing sites over `prismaquant/` (31 `try` blocks wrap a detection call —
+25 `except DeadVendoredOverrideError: raise` then broad, 4 `try/finally`, 1
+`SampleParallelProbeError`+broad-that-re-raises, 1 narrow non-catching), so
+the tree was and stays clean; what changed is that the pin now holds the shape
+it claims to hold. The census's own docstring names what it does not cover
+(aliased or dispatched calls, a swallow one frame up, computed exception
+classes, callers outside `prismaquant/`), since it is cited as future-proofing.
+
 Re-stamped (2026-09-05, `claude/pq-gates`) for **the export seal's
-capture-context roster read from Tessera rather than typed** (§5;
+capture-context roster read from Tessera rather than typed** (§5.7;
 RobTand/prismaquant#216, P2). `tessera_export_lane.CAPTURE_CONTEXT_FIELDS` was
 the literal `("model", "seqlen", "source")` and nothing in the tree read
 `tessera.export.CAPTURE_CONTEXT`, so half of the roster
@@ -6660,7 +6689,10 @@ plan translation when the allocation's `tessera_hessian` stamp or its
 selected W4A4 routes declare a requirement the supplied files do not satisfy
 (#193): the artifact built must be the artifact priced. The binding is to
 content, not to names (#204): the allocation carries the campaign capture's
-`capture_sha256` (Tessera's own seal rule, `hessian_capture_sha256`) and the
+`capture_sha256` (Tessera's own seal rule, `hessian_capture_sha256`, whose
+capture-context roster is read from `tessera.export.CAPTURE_CONTEXT` rather
+than typed, with one documented fallback for pins that predate the constant
+and a by-name refusal when the two rosters disagree — #216) and the
 `input_global_scale` each selected W4A4 unit was priced under
 (`tessera_activation_static_scales`), and the gate digests the `.pt` the
 exporter loads and reads each scalar from the safetensors file, refusing a
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,27 @@
 As of: 2026-09-05 · `claude/pq-gates`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-05, `claude/pq-gates`) for **a campaign test that cannot
+leak the child it orphans on purpose** (§3.0; RobTand/prismaquant#219, P3).
+`test_resume_after_coordinator_death_recovers_owned_helper_receipt` kills the
+coordinator while a stage child polls a release file, and after #206 replaced
+that child's self-limiting `time.sleep(1.0)` with an unbounded wait, writing
+the release file became the only thing that could ever end it — the worker's
+`timeout_seconds` enforcement (`cluster_campaign.py:1164-1178`,
+`start_new_session=True` then `_terminate_child_process_group`) is dead by
+construction. The write sat outside the guard: anything raising between
+`coordinator.terminate()` and it — most concretely the 120 s wedge backstop, or
+the session being killed, which happened to agents on this box today — left a
+`python -c` polling at 50 Hz forever, on a release file that `tmp_path` cleanup
+then made unreachable. Its sibling
+`test_abrupt_worker_death_leaves_stage_lock_owned_by_child` already had the
+`finally` and the comment explaining why. Now both share one home for it:
+`_orphaned_helper(release_path)` releases and, if the child did not take the
+release, kills its session on every path out, and the child announces its PID
+so `helper.assert_exited()` can make "this test leaves nothing running" an
+assertion rather than an arrangement. Test-only; no production code, no
+pipeline behaviour and no contract changes.
+
 Re-stamped (2026-09-05, `claude/pq-gates`) for **the dead-override census
 seeing every handler shape that catches its exception** (§8.1;
 RobTand/prismaquant#217, P3). `tests/test_vendored_override_gate.py`'s
--- a/prismaquant/tessera_export_lane.py
+++ b/prismaquant/tessera_export_lane.py
@@ -488,10 +488,57 @@ def require_assignment_scope(model_path: str | Path, assignment_path: str | Path
 #: together.
 PRICED_HESSIAN_IDENTITY_FIELDS = ("text_sha256", "fit_tokens", "fit_ids_sha256")
 
+#: The capture-context roster used only where the running Tessera does not
+#: publish ``CAPTURE_CONTEXT``: the pinned development Tessera (1221d2a)
+#: predates the constant, and this gate must refuse on a machine that can read
+#: the payload without importing ``tessera`` at all.  **Copied from tessera
+#: 0.1.0 at release 3efd690** (``src/tessera/export.py:244``), and this is the
+#: ONE place the roster is typed -- a second spelling of a seal's key set is
+#: how the seal goes vacuous (RobTand/prismaquant#216).
+_CAPTURE_CONTEXT_FALLBACK = ("model", "seqlen", "source")
+
+#: Names for where :data:`CAPTURE_CONTEXT_FIELDS` was read, so a refusal can
+#: say whether the digest covered Tessera's own roster or the copied one.
+_CAPTURE_CONTEXT_FROM_TESSERA = "tessera.export.CAPTURE_CONTEXT"
+_CAPTURE_CONTEXT_FROM_FALLBACK = (
+    "prismaquant.tessera_export_lane._CAPTURE_CONTEXT_FALLBACK "
+    "(copied from tessera 0.1.0 release 3efd690)")
+
+
+def _tessera_capture_context() -> "tuple[str, ...] | None":
+    """``tessera.export.CAPTURE_CONTEXT``, or None where the running Tessera
+    predates the constant (the pinned 1221d2a does) or is not importable."""
+    try:
+        from tessera.export import CAPTURE_CONTEXT
+    except ImportError:
+        return None
+    return tuple(CAPTURE_CONTEXT)
+
+
+def _capture_context_fields() -> "tuple[tuple[str, ...], str]":
+    """The capture-context roster this digest covers, and where it was read.
+
+    Read from the installed Tessera where it publishes the constant, exactly
+    as ``tessera_hessian.HESSIAN_IDENTITY_FIELDS`` reads the identity triple
+    from ``tessera.export.HESSIAN_IDENTITY`` -- the roster belongs to the code
+    that owns the seal, and typing a second copy of it is how the two rules
+    drift apart unnoticed.  Where the constant is absent the one documented
+    fallback above is used and named, because such a pin also has no
+    ``ActivationSource.capture_sha256`` and the runtime cross-check cannot see
+    the difference (RobTand/prismaquant#216).
+    """
+    published = _tessera_capture_context()
+    if published is None:
+        return _CAPTURE_CONTEXT_FALLBACK, _CAPTURE_CONTEXT_FROM_FALLBACK
+    return published, _CAPTURE_CONTEXT_FROM_TESSERA
+
+
 #: The capture-context fields Tessera's seal covers beside the triple
 #: (``tessera.export.CAPTURE_CONTEXT``): two captures of one token prefix
-#: differ only here when the sequence layout differs (tessera#214).
-CAPTURE_CONTEXT_FIELDS = ("model", "seqlen", "source")
+#: differ only here when the sequence layout differs (tessera#214).  Read from
+#: Tessera rather than typed, and :func:`_require_capture_context_roster`
+#: refuses when what was read has since drifted from what Tessera publishes.
+CAPTURE_CONTEXT_FIELDS, CAPTURE_CONTEXT_FIELDS_SOURCE = _capture_context_fields()
 
 #: The schema string sealed into the digest, Tessera's own spelling.
 HESSIAN_CAPTURE_SHA256_SCHEMA = "tessera.hessian_capture.v1"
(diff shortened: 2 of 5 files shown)


RobTand:
## Commands and counts

sparklina, CPU only, **real git worktree** (`/home/rob/prismaquant-audit-wt/pq-gates`,
`git rev-parse HEAD` resolves — not an rsync'd copy, so nothing here is a
`_production_cache_git_commit`-style harness artifact).

```
CUDA_VISIBLE_DEVICES="" PYTHONPATH=<tessera>/src \
  /home/rob/venvs/pq-release/bin/python -m pytest \
  tests/test_tessera_priced_export_inputs.py tests/test_tessera_export_lane.py \
  tests/test_allocator_tessera_priced_inputs.py tests/test_tessera_campaign.py \
  tests/test_vendored_override_gate.py tests/test_model_structure.py \
  tests/test_cluster_campaign_runner.py tests/test_architecture_doc.py \
  tests/test_docs_staleness.py \
  -p no:cacheprovider -q -n 2
```

| Tessera tree | Result |
|---|---|
| `/home/rob/tessera-pin-1221d2a/src` (the pin PQ pins today) | **200 passed, 6 skipped, 0 failed** |
| `/home/rob/tessera-release/src` (`3efd690`) | 181 passed, 6 skipped, **19 failed — all inherited at that pin** |

Re-run after merging `origin/main` twice (`aaf46430` #220 packed-expert bridge,
then `a5495b55` #223/#211). Adding `tests/test_tessera_campaign_resume.py` to
the dev-pin run gives **17 failed, 202 passed, 7 skipped** — exactly the known
17 main-entry failures on main, nothing else. The merged-in #220 files
(`test_lane_gate_recording.py test_tessera_campaign_packed.py
test_tessera_plan_input_binding.py test_tessera_expert_projection.py
test_allocator_expert_projection.py test_tessera_export_projection.py`) are
`100 passed, 7 skipped` at the dev pin on this tip.

The 19 are at the **release-tip Tessera**, which is a different PYTHONPATH from
the pin PQ actually uses, and they reproduce identically on `553de0af` with the
same files in a second real worktree. They are not about this PR: they are the
release Tessera's `runtime_contract.json` against a PQ still pinned to the dev
contract —

```
prismaquant.lane_eligibility.LaneEligibilityError:
  runtime_contract.lane_eligibility.structures must be a non-empty list of unique ids
```

13 in `test_tessera_campaign.py`, 4 in `test_allocator_tessera_priced_inputs.py`,
2 in `test_tessera_export_lane.py`. That is `claude/tessera-pin-v0.1.0`'s region
(`lane_eligibility.py`, `tessera_runtime_contract.py`, the pin files), untouched
here.

`tessera_export_lane.py` merged clean against #220: the projection-binding leg
#220 added to `require_assignment_scope` and the capture-context roster work
here are different regions of the file, and both sides survive
(`_CAPTURE_CONTEXT_FALLBACK`, `_tessera_capture_context`,
`_require_capture_context_roster` all present and exercised).
`docs/ARCHITECTURE.md` was resolved by hand both times: ONE
`As of: 2026-09-05 · claude/pq-gates` header line, then every stamp paragraph
newest-first — three from this branch on top, then `claude/pq-211`, then the
four from `claude/pq-183-packed-bridge`, then `claude/pq-204` and the rest.

## Regressions, shown failing on `553de0af` first

**#216** — the five new tests applied alone to a `553de0af` worktree:

```
$ CUDA_VISIBLE_DEVICES="" PYTHONPATH=<either tessera>/src python -m pytest \
    tests/test_tessera_priced_export_inputs.py -p no:cacheprovider -q
5 failed, 34 passed, 1 skipped          # at BOTH pins

E  Failed: DID NOT RAISE TesseraExportLaneError
   # test_a_capture_context_field_our_roster_lacks_refuses -- a Tessera naming
   # a field our roster lacks digested silently, at a pin with no capture_sha256

E  AssertionError: Regex pattern did not match.
     Expected regex: 'different provenance'
     Actual message: "--hessian .../hessian_capture.pt: PrismaQuant's
     hessian_capture_sha256 (d9739d07...) disagrees with tessera.export.
     ActivationSource.capture_sha256 (052a3567...) ... have drifted"
   # test_the_gate_refuses_a_bind_under_a_drifted_roster -- pre-fix the bind
   # reached the seal cross-check, which only exists where the pin seals
```

**#217** — the census's predicates live in the test file, so only the new tests
were appended to `553de0af`'s own copy:

```
6 failed, 38 passed

E  AssertionError: except RuntimeError
E  AssertionError: except (ImportError, RuntimeError)
E  AssertionError: except builtins.Exception (dotted)
E  AssertionError: broad handler returning before its raise
E  AssertionError: except DeadVendoredOverrideError that answers
E  NameError: name 'CATCHES_DEAD_OVERRIDE' is not defined
```

The control (`except Exception: profile = None`, the shape #202 removed) passes
on `553de0af` as it must, and all six `REFUSING_HANDLERS` shapes — every shape
the live tree actually uses — pass on both sides, so the widening reports
nothing that lets the refusal out.

Census over the tree, after the widening:

```
detection calls inside a try (Try nodes): 31
   25  ('DeadVendoredOverrideError', 'Exception')
    4  ()                                              # try/finally
    1  ('SampleParallelProbeError', 'Exception')       # always re-raises
    1  ('TesseraExportLaneError', '(OSError, ValueError, TypeError, KeyError, struct.error)')
flagged by the widened census: []
```

Zero newly swallowing sites, so nothing to fix or justify per site.

**#219** — a pytest plugin raises `AssertionError` at the
`unmet="coordinator did not exit after terminate()"` wait, i.e. inside the
window that was unguarded, and `/proc` is then scanned **once** for anything
still holding the run's `--basetemp` path:

```
553de0af   1 failed / LEAKED=2
             pid 231796: python -c ... exec("while not release.exists():
                         time.sleep(0.02)") .../release-the-helper
             pid 231888: the scanner itself
this tip   1 failed / LEAKED=1
             pid 235223: the scanner itself
```

The demo orphan was ended by writing its release file, not killed. No synthetic
load, no polling loops, nothing `pkill`ed.

## For Rob

#219 also asks whether an orphan-by-design helper should carry its own absolute
backstop (`if time.monotonic() > start + N: sys.exit(9)`). That re-introduces a
wall-clock deadline into a child, which is exactly what #199 measured out of
this file, so it is a trade rather than a cleanup and I did not make it. The
leak is closed without one.

Issue #211, opened 2026-09-05
Tessera campaign rewrites the export inputs before the resume identity can refuse them


RobTand:
Side-finding from pq-204 (RobTand/prismaquant#204), out of that fix's declared region; filed per AGENTS principle 14.

## What

`tessera_campaign.main()` calls `write_export_inputs(...)` (`prismaquant/tessera_campaign.py:1296` at `claude/pq-204` f8860e22) **before** `prepare_journal(...)` (`:1353`) and before the per-row resume verification loop (`:1361-1386`). So a resume the checkpoint identity **refuses** -- `RuntimeError: Tessera campaign checkpoint identity mismatch at units.<unit>.hessian.sha256 ...; refusing reuse or recompute` -- has already overwritten `<cache-dir>/hessian_capture.pt`, its `.provenance.json` sidecar and `<cache-dir>/input_scales.safetensors` with **the refused run's** Hessians and static scales. The checkpoint and the previous cost file are untouched (correctly), but the export inputs that belong to that surviving table are gone.

## Proof (CPU, release Tessera `/home/rob/tessera-release/src`, synthetic model; real `main()`, real writers, real journal)

codex's `prismaquant_rejected_resume_export_inputs.py` (pre-#204) and my post-#204 re-run `/home/rob/pq-audit-caches/pq-204/proofs/rejected_resume_postfix.py` -> `rejected_resume_postfix.json`:

```
refusal: "Tessera campaign checkpoint identity mismatch at units.model.layers.0.proj.hessian.sha256: ...; refusing reuse or recompute"
checkpoint_unchanged: true         previous_cost_file_unchanged: true
hessian_before: 1.0                hessian_after: 2.0        digest_changed: true
scale_before: 1.0                  scale_after: 0.5
old_assignment_export_gate (pre-#204):  accepted            <- codex
old_assignment_export_gate (post-#204): "refused: --hessian .../hessian_capture.pt is not the capture that priced this allocation: capture_sha256 payload=dbd6321420... != allocation=..."
```

#204 bounds the damage -- the first campaign's allocation now refuses the rewritten files by name instead of exporting under the wrong Hessian and scale -- but the files a refused run destroys are still destroyed, and the surviving checkpoint/table cannot be exported until the campaign is re-run from scratch.

## Could pq-204 have fixed it? Yes -- and chose not to, for two reasons

[...]

PR #226, opened 2026-09-05
Tessera campaign and export lane: ask the contract, not the format name (#221)

### LINKED ISSUE #221: Two homes for 'does this route execute the static NVFP4 activation contract': the campaign and export lane still compare the name, after #205 made it the spec's answer
## What

"Does this Tessera rung's route execute the static NVFP4 activation contract" is now answered two ways. PR #213 (#205) moved the consumers to the spec's own answer and said so explicitly — `perturbed_x_cache.py:122`:

> Which quantizer a spec serves is the SPEC's answer (``FormatSpec.static_activation_contract``), **never a compare of its name against ``"NVFP4"``** -- a Tessera rung routed through the same kernel has the same contract and a different name (#205).

`tessera_render.synthesize_tessera_spec` is where that answer is derived, and it derives it through the registry row that owns it (`tessera_render.py:782-788`):

```python
        source = fr.get_format(route.activation_source_format)
        activation_qdq = source.activation_quantize_dequantize
        static_activation_contract = (
            None if source.static_activation_contract is None
            else dataclasses.replace(
                source.static_activation_contract, measured_as_served=True)
        )
```

Three sites still answer the same question by comparing the route's source-format **name**, skipping the registry row entirely:

* `prismaquant/tessera_campaign.py:298` — decides whether a rung needs a calibrated `input_global_scale` to be scored, and raises `ActivationScaleContractError` if not:
  ```python
      if route.activation_source_format == "NVFP4":
  ```
* `prismaquant/tessera_campaign.py:999-1000` — `_format_executes_static_nvfp4`, whose docstring is literally the question ("Does this rung's route execute the static NVFP4 activation contract?"):
  ```python
      return tessera_serving_route(
          family, wire, rung).activation_source_format == "NVFP4"
  ```
* `prismaquant/tessera_export_lane.py:792-794` — decides which selected units require `--input-scales`:
  ```python
          if tessera_serving_route(
                  family, wire, rung).activation_source_format == "NVFP4":
              w4a4.append(name)
  ```

Note that `tessera_campaign.py:298` does not even use its own module's `_format_executes_static_nvfp4` helper 700 lines below it, so there are two spellings inside one file as well as two homes across three.

Nothing today. I verified over the whole registry that the two predicates coincide exactly — for all 80 rows, `canonical_format_name(name) == "NVFP4"` and `static_activation_contract is not None` agree, and `NVFP4` is the only row carrying a contract:

```
rows where old name-compare and new contract-check DISAGREE: []
rows with a static contract:      ['NVFP4']
rows canonicalising to NVFP4:     ['NVFP4']
```

The hazard is what happens the day a second row gets one — which the contract exists to allow, since `StaticActivationContract` is a general dataclass with `execution`/`group_size` fields and `format_registry.py:143` types the field as a spec-level property, not an NVFP4 flag. Say `FP8_E4M3` gains a static per-tensor scale contract. Then:

* `tessera_render` (`:782-788`) stamps it onto every Tessera dense-E4M3 spec with `measured_as_served=True`, because it reads the registry row;
* `perturbed_x_cache._activation_qdq` (`:137`) and `production_weight_cache._render_score_record` (`:1444`) immediately start pricing those rungs through the served oracle, and **refuse by name** (`ActivationScaleContractError`) any unit with no calibrated maximum;
* but `tessera_campaign.py:298` still says "not NVFP4", so the campaign never demands or stamps an `input_global_scale` for them; `_format_executes_static_nvfp4` (`:999`) says False, so no scale reaches the cost row; and `tessera_export_lane.py:792` says False, so the export gate never requires `--input-scales`.

The result is the two halves of #204/#205 disagreeing about the same rung: the scorer and the KL hooks refuse to price it, or the export ships a unit whose static A-side scale nothing ever bound — which is the failure #205 was opened to close, at a different format name. A one-line registry change in a distant file silently splits the pipeline.

## Severity

P3. Latent, not live: no row other than `NVFP4` carries a contract today and I proved the two predicates are currently equivalent. It is a "two homes for one rule" of the kind AGENTS.md targets, on the rule two P0/P1 fixes (#204, #205) both depend on, and the second home is a name compare that #205's own docstring names as the thing to stop doing.

## Fix

Give the campaign and the export lane the same derivation `tessera_render` uses. The natural single home is a predicate on the route that reads the registry row, next to `tessera_serving_route`:

```python
def route_executes_static_activation_contract(route) -> bool:
    """The route's A-side served contract, from the row that owns it."""
    if route.activation_source_format is None:
        return False
    return fr.get_format(
        route.activation_source_format).static_activation_contract is not None
```

Then `tessera_campaign.py:298` and `:999` and `tessera_export_lane.py:792` all call it, and `_format_executes_static_nvfp4` either becomes that call or goes away. `tessera_render.py:782-788` keeps ownership of turning the row's contract into the spec's (the `measured_as_served=True` re-stamp), which is a separate step and should stay where it is.

Two things to decide when doing it, so this is not a mechanical edit:

1. Whether the campaign should require a calibrated scale for *any* static contract or only for the NVFP4 one — the refusal message at `:299-305` names `scaled_fp4_quant` specifically, and a second contract would need its own sentence or a generic one built from `contract.execution`.
2. Whether `tessera_export_lane`'s `w4a4` list should be renamed once it can hold non-W4A4 units; today the name encodes the assumption the compare makes.

Given (1) and (2) this may deserve `needs-decision` rather than a straight fix; I have not applied that label, since the fix as scoped above is behaviour-preserving today and the decisions only bite when a second contract appears.

## Provenance

The name compares predate #213 (`tessera_campaign.py:298` and `tessera_export_lane.py:792` are older; `_format_executes_static_nvfp4` is the campaign's own helper). PR #213 (`claude/pq-205`, merge `473cf6ca`) created the second home by introducing `FormatSpec.static_activation_contract` as the owning answer and converting only the two consumers it was scoped to — `perturbed_x_cache._activation_qdq` and `production_weight_cache._render_score_record` — leaving the Tessera campaign and export lane on the old spelling. PR #212 (`claude/pq-204`, merge `553de0af`) then built the export-input binding on top of `tessera_export_lane.py:792` without revisiting it.

Auditor, file-only by brief. Related: #205 (which established the rule), #204 (which depends on the export-lane site).

RobTand:
Fixes #221.

## One rule, one home

PR #213 (#205) made "which activation quantiser does this spec serve" a
property of the spec — `nvfp4_activation_contract.StaticActivationContract`,
reached through `FormatSpec.static_activation_contract` — and routed the
assignment-KL hooks and the production cache scorer through it.
`perturbed_x_cache.py:122` states the rule in as many words: *"Which quantizer
a spec serves is the SPEC's answer … **never a compare of its name against
`"NVFP4"`**"*.

Three call sites still answered the same question by comparing the route's
source-format **name**, skipping the registry row entirely — and one of them
was a helper in the same file as another, so there were two spellings inside
one module as well as two homes across three:

| site | was |
|---|---|
| `tessera_campaign._measure_anchor` | `route.activation_source_format == "NVFP4"` |
| `tessera_campaign._format_executes_static_nvfp4` | same compare, 700 lines later |
| `tessera_export_lane` `--input-scales` gate | same compare again |

The derivation now has one owner:
**`tessera_formats.route_static_activation_contract(route)`** returns the
contract of the registry row the route names in `activation_source_format`.
It returns the *contract*, not a bool, because every consumer needs what is
inside it; a bool would have sent each of them back to the row for the rest.

* `tessera_render.synthesize_tessera_spec` derives through it and keeps the
  one step that is genuinely its own — re-stamping the row's contract
  `measured_as_served=True` to build the rung's spec.
* `tessera_campaign._measure_anchor` reads the spec it already resolved, and
  now prices through `contract.quantize_dequantize` instead of a hardcoded
  `fr.nvfp4_activation_qdq_served`, refusing with `contract.execution` instead
  of the literal string "NVFP4".
* `_format_executes_static_nvfp4` → `_format_executes_static_activation_contract`,
  over the same spec field.
* `tessera_export_lane` calls the route accessor **directly**, not
  `get_format(<rung>)` — that would reach `synthesize_tessera_spec` and pull
  the `tessera` package into a preflight whose own docstring says it gates
  "without importing the `tessera` package". A test guards this.

## Behaviour today is unchanged, and that is pinned

Over all 80 registry rows, `canonical_format_name(name) == "NVFP4"` and
`static_activation_contract is not None` agree exactly, and `NVFP4` is the
only row carrying a contract:

```
rows: 80
rows where the name-compare and the contract-check DISAGREE: []
rows with a static contract: ['NVFP4']
```

No format resolves differently. That equivalence is now a test, so the day it
stops holding, it stops holding deliberately.

## Failing-first, not a pin

The hazard #221 names is constructible, so this ships as a real regression
rather than a cleanup. `StaticActivationContract` is a general dataclass with
`execution`/`group_size` fields and `FormatSpec` types the field as a
spec-level property, not an NVFP4 flag. Give `FP8_E4M3` a static contract —
one `dataclasses.replace` in a distant file — and `TESSERA_E4M3_K1_R1024`,
whose default wire recipe is the CHANNEL plane and whose route therefore
already carries `activation_source_format="FP8_E4M3"`, becomes a rung where
the two predicates disagree:

* `tessera_render` stamps the contract onto the spec (it reads the row), so
  the cache scorer and the KL hooks immediately price the rung through the
  served oracle and **refuse by name** any unit with no calibrated maximum;
* the name compare says "not NVFP4", so the campaign demands and stamps no
  `input_global_scale`, and the export gate never requires `--input-scales`.

That is the two halves of #204/#205 disagreeing about the same rung, and the
export shipping a unit whose static A-side scale nothing ever bound — the
failure #205 was opened to close, at a different format name. The fixture uses
execution `e4m3_per_tensor_static`, deliberately not an NVFP4 string, so a
refusal that still says "NVFP4" cannot pass by accident.

## Machine-readable change

The preflight report key **`w4a4_units` is renamed
`static_activation_contract_units`**: it can now hold units that are not W4A4,
and a key whose name encodes an assumption the code no longer makes is a claim
beyond its evidence. This is safe because the value is never persisted —
`require_priced_export_inputs` returns it, `preflight` nests it as
`report["priced_inputs"]`, and the CLI serialises only `report["build"]` under
`--write-build-json`. Grep for readers over `*.py *.sh *.json *.md`:

```
prismaquant/tessera_export_lane.py:685    (the initialiser)
prismaquant/tessera_export_lane.py:795    (the assignment)
tests/test_tessera_priced_export_inputs.py:437,483,494
tests/test_allocator_tessera_priced_inputs.py:94
```

No receipt, allocation JSON, shipcard slot or `run-pipeline.sh` reader touches
it. Refusal texts in the campaign and the export lane now name the executed
contract rather than "NVFP4"; two existing test regexes move with them.

## Prose

`docs/ARCHITECTURE.md` §5.1 claimed that "a consumer holding the unit's
calibrated maximum (the assignment-KL hooks, the production cache scorer, **the
campaign**) asks the contract which quantiser and which G, never the format's
name." That was false for the campaign. This PR makes it true rather than
softening it, and adds the paragraph describing the single derivation
underneath. The provenance block is re-stamped.

## Not changed, deliberately

Other `== "NVFP4"` compares in the tree are different questions and stay:
`perturbed_x_cache:966,969,1151,1153` gates PrismaQuant's own fused CUDA
kernel, which really does only handle the stock NVFP4 packed layout;
`kl_measurement:219` is sidecar-byte accounting for the compressed-tensors
artifact; the `export_native_compressed` / `allocator` / `footprint` hits are
the stock export codec.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,49 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-05 · `claude/pq-183-packed-bridge`. Stamps
+As of: 2026-09-05 · `claude/pq-221`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-05, `claude/pq-221`) for **one home for "does this route
+execute a static activation contract"** (§5.1, §5.7; RobTand/prismaquant#221,
+P3). PR #213 (#205) made "which activation quantiser does this spec serve" a
+property of the spec (`FormatSpec.static_activation_contract`) and moved the
+assignment-KL hooks and the production cache scorer onto it; the Tessera
+campaign and export lane kept answering the same question by comparing the
+route's source-format NAME against `"NVFP4"`, in three places and two
+spellings — one of them a helper whose docstring was literally the question.
+The derivation now has one owner:
+`tessera_formats.route_static_activation_contract(route)` returns the contract
+of the registry row the route names (`activation_source_format`), and every
+consumer reads it. `tessera_render.synthesize_tessera_spec` re-stamps its
+result `measured_as_served=True` onto the rung's spec — the one step that
+stays where it was, because turning the ROW's contract into the SPEC's is a
+different question. `tessera_campaign._measure_anchor` prices and refuses
+through the spec's copy (`contract.quantize_dequantize` instead of a hardcoded
+`nvfp4_activation_qdq_served`, and a refusal naming `contract.execution`
+instead of a format name); `_format_executes_static_nvfp4` becomes
+`_format_executes_static_activation_contract` over the same spec field; and
+`tessera_export_lane`'s `--input-scales` gate calls the route accessor
+directly rather than `get_format(<rung>)`, which would reach
+`synthesize_tessera_spec` and pull the `tessera` package into a preflight
+built not to need it (that module's own docstring). **Nothing resolves
+differently today**: over all 80 registry rows `canonical_format_name(name) ==
+"NVFP4"` and `static_activation_contract is not None` agree exactly, and
+`NVFP4` is the only row carrying a contract. The regression is the day a
+second row gets one — `StaticActivationContract` is a general dataclass and
+`FormatSpec` types the field as a spec-level property, not an NVFP4 flag — and
+`test_static_activation_contract_one_home.py` constructs exactly that: with
+`FP8_E4M3` given a static contract the name compare answers "not NVFP4" while
+the cache scorer and the KL hooks, which already read the row, refuse the same
+unit; the campaign then stamps no `input_global_scale` and the export ships a
+unit whose static A-side scale nothing ever bound — the failure #205 was
+opened to close, at a different format name. **Machine-readable change**: the
+preflight report key `w4a4_units` is renamed
+`static_activation_contract_units`, because it can now hold units that are not
+W4A4; its only readers were this module and four test assertions, and no
+receipt, allocation JSON or shipcard carries it (the CLI serialises only
+`report["build"]`). Refusal texts in the campaign and the export lane now name
+the executed contract rather than "NVFP4". No default, format menu, serving
+lane, ship gate or byte changes.
 
 Re-stamped (2026-09-05, `claude/pq-183-packed-bridge`) for **the export lane
 handing the exporter the priced expert wires** (§4.10, §9.4; PrismaQuant
@@ -6398,6 +6440,18 @@ name. The contract's `measured_as_served` is the spec's measurement policy:
 served emulation is `PRISMAQUANT_NVFP4_ACT_EMULATE_SERVED_SCALES`), `True` on
 a Tessera W4A4 spec (§5.7). Dynamic W8A8 (FP8, MX) and A16 rows carry `None`.
 
+That spec-level answer has one derivation beneath it (#221):
+`tessera_formats.route_static_activation_contract(route)` returns the contract
+of the registry row a Tessera route names in `activation_source_format`.
+`tessera_render.synthesize_tessera_spec` re-stamps that result
+`measured_as_served=True` to build the rung's spec, so a consumer that reaches
+for `FormatSpec.static_activation_contract` and one that asks the route
+directly — `tessera_export_lane`'s `--input-scales` gate, which must not reach
+the synthesizer and pull in the `tessera` package — are reading the same row
+through the same rule, and cannot answer differently. A name compare would
+answer the same today only because `NVFP4` is the one row carrying a contract;
+the field is typed as a spec-level property, not an NVFP4 flag.
+
 ### 5.2 Scale rules and JSO
 
 NVFP4 scale rules live in the *exporter*, not the registry
@@ -6810,11 +6864,12 @@ exporter — `TESSERA_HESSIAN` (the campaign's `hessian_capture.pt`) and
 `TESSERA_INPUT_SCALES` (its `input_scales.safetensors`) — and
 `tessera_export_lane.require_priced_export_inputs` fails closed before the
 plan translation when the allocation's `tessera_hessian` stamp or its
-selected W4A4 routes declare a requirement the supplied files do not satisfy
+selected static-activation-contract routes declare a requirement the
+supplied files do not satisfy
 (#193): the artifact built must be the artifact priced. The binding is to
 content, not to names (#204): the allocation carries the campaign capture's
 `capture_sha256` (Tessera's own seal rule, `hessian_capture_sha256`) and the
-`input_global_scale` each selected W4A4 unit was priced under
+`input_global_scale` each selected static-contract unit was priced under
 (`tessera_activation_static_scales`), and the gate digests the `.pt` the
 exporter loads and reads each scalar from the safetensors file, refusing a
 payload, sidecar or value that is not what priced the row, and an allocation
--- a/prismaquant/tessera_campaign.py
+++ b/prismaquant/tessera_campaign.py
@@ -291,24 +291,37 @@ def _measure_anchor(
     # the predicate reads and the plane the encode writes are this object --
     # not three lookups that agree only while nothing clears the recipe memo.
     wire = tessera_wire_recipe(family, rung)
-    # The A side, as served.  An NVFP4-materialising route executes vLLM's
-    # static-global-scale ``scaled_fp4_quant`` (UE4M3 block scales) against the
-    # artifact's ``trellis_input_global_scale``; every other route keeps the
-    # serving format's own dynamic quantiser, taken by reference from the spec.
+    # The A side, as served.  A route with a STATIC activation contract
+    # executes vLLM's static-global-scale ``scaled_fp4_quant`` (UE4M3 block
+    # scales) against the artifact's ``trellis_input_global_scale``; every
+    # other route keeps the serving format's own dynamic quantiser, taken by
+    # reference from the spec.
+    #
+    # WHICH of the two is the SPEC's answer, never a compare of the route's
+    # source-format name against ``"NVFP4"`` (#205's rule, #221's fix): the
+    # spec resolved above already carries the contract, because
+    # ``synthesize_tessera_spec`` derived it from the registry row the route
+    # names.  A Tessera rung routed through the same kernel has the same
+    # contract and a different name, and the day a second registry row gets a
+    # contract, a name compare here would price it under a quantiser the
+    # runtime never runs while the cache scorer and the KL hooks -- which
+    # already read the row -- refuse the same unit.
     route = tessera_serving_route(family, wire, rung)
-    if route.activation_source_format == "NVFP4":
+    contract = spec.static_activation_contract
+    if contract is not None:
         if static_input_scale is None or not float(static_input_scale) > 0:
             raise ActivationScaleContractError(
                 f"{qname} {format_name}: this rung's route executes the "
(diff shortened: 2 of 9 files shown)


RobTand:
## Commands and counts

Edited on sparklina, suites run on sparky (`/home/rob/venvs/pq-release/bin/python`,
`PYTHONPATH=/home/rob/tessera-pin-1221d2a/src`, `CUDA_VISIBLE_DEVICES=""`).

**The far box must be a real git repository.** An rsync'd git *worktree* is not
one — its `.git` is a file pointing at a path that does not exist there — and
`production_weight_cache._production_cache_git_commit` correctly fail-closes on
`git rev-parse --verify HEAD^{commit}` returning 128. Two
`test_production_weight_cache.py` CB tests read as failures under a naive sync
and pass in a real checkout, so the sync does `git init` + commit first:

```
rsync -a --delete --exclude '.git' /home/rob/prismaquant-audit-wt/pq-221/ \
    sparky:/home/rob/audit-sync/pq-221/
ssh sparky 'cd /home/rob/audit-sync/pq-221 \
  && (git rev-parse --git-dir >/dev/null 2>&1 || git init -q) \
  && git add -A && git -c user.email=a@b -c user.name=c commit -qm sync'
ssh sparky 'cd /home/rob/audit-sync/pq-221 && CUDA_VISIBLE_DEVICES="" \
  PYTHONPATH=/home/rob/tessera-pin-1221d2a/src \
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider -q -n 4 <files>'
```

### Failing-first, on the commit's parent

`tests/test_static_activation_contract_one_home.py` alone, before the fix:

```
6 failed, 5 passed
```

The 6 are the three call sites, the renamed report key, and the guard that the
export gate never reaches `synthesize_tessera_spec`. The 5 green are the
80-row equivalence pin, the live routes, the accessor identity, the
spec-answer-equals-route-answer pin, and the render half (which already read
the row — that is why the other three disagreeing was a split).

### After the fix

`test_tessera_formats.py test_served_activation_contract.py
test_tessera_render_runtime_context.py test_tessera_campaign.py
test_tessera_campaign_resume.py test_tessera_campaign_pins.py
test_tessera_campaign_packed.py test_tessera_export_lane.py
test_tessera_priced_export_inputs.py test_allocator_tessera_priced_inputs.py
test_production_weight_cache.py test_kl_measurement_teacher_pairing.py
test_architecture_doc.py test_docs_staleness.py
test_static_activation_contract_one_home.py test_tessera_expert_projection.py
test_tessera_export_projection.py test_tessera_plan_input_binding.py
test_allocator_expert_projection.py`

```
17 failed, 376 passed, 14 skipped, 56 warnings in 155.88s
```

All failures are the **17 known-inherited `test_tessera_campaign_resume.py`
main-entry failures** at this dev pin. The count is unchanged from the
pre-change baseline measured the same way.

Second batch — everything grepping `canonical_format_name`, plus the Tessera
footprint / cost-currency / lane-gate readers:

`test_mxfp8_ue8m0.py test_prismaquant_native_math.py
test_tessera_candidate_context.py test_m4_frontier_expert_selection.py
test_format_menu_expansion.py test_serving_profiles.py
test_prismaquant_export_native_compressed.py test_tessera_footprint.py
test_cost_currency.py test_lane_gate_recording.py`

```
387 passed, 68 warnings, 163 subtests passed
```

### The `synthesize_tessera_spec` guard, measured

Design call (3) — the export lane calling the route accessor rather than
`fr.get_format(<rung>)` — rests on the gate not reaching the Tessera
synthesizer. Verified with a spy on `tessera_render.synthesize_tessera_spec`,
both before and after #220 added its projection-binding leg to that gate:

```
synthesize_tessera_spec calls: []
```

`test_the_export_gate_does_not_import_the_tessera_package` keeps it that way.

### Merges

Branched off `origin/main` `553de0af`, merged `origin/main` at `aaf46430`
(#220, the packed-expert bridge, which rewrote both files this touches),
`a5495b55` (#223 / #211), `17ca6930` (#224 / #218) and `eda5b37d` (#225).
Never rebased. Every conflict was in `docs/ARCHITECTURE.md`, resolved to one
`As of:` header line with all stamp paragraphs newest-first.

#225 also collided in §5.7's body: its capture-context roster clause (#216)
and this PR's `W4A4 unit` -> `static-contract unit` rename land in the same
sentence. **Both** were kept -- the roster clause with the rename applied on
top -- since dropping either would have left the tree asserting something it
no longer does. `tests/test_tessera_priced_export_inputs.py` auto-merged and
carries #225's `test_the_capture_context_roster_is_tesseras` alongside this
PR's five rename assertions; #225 added no key to the gate's report, so the
whole-dict equality assertion still matches the initialiser exactly.

### Re-verified against #218's case-insensitive resolution

#218 made `canonical_format_name` resolve case-insensitively as a last
probe, so the 80-row equivalence this PR pins was re-run rather than
assumed:

```
rows: 80          disagree: []          with contract: ['NVFP4']
mixed-case registered rows: ['INT4_W4A16_g128']
INT4_W4A16_g128 -> resolves, contract: None  (predicate answers False)
```

Neither derivation added here normalises on top of the registry -- both call
`fr.get_format` with no `.upper()` or `canonical_format_name` of their own,
the same rule #218 applied when it dropped the redundant `.upper()` from
`_format_uses_static_activation_clip`.

`test_the_campaign_predicate_answers_every_rung_as_the_name_compare_did`
runs the **old body verbatim** beside the new one. They agree on every input
the helper is actually called with. Where they differ the old one *raised* --
it unpacked `parse_tessera_format_name` without a None check, so any
non-Tessera name died with `TypeError`. A widening of the domain, never a
changed answer.

PR #228, opened 2026-09-05
Tessera export: name which path produced the routed-expert bytes (#222) and stop bundling when no expert is selected (#229)

### LINKED ISSUE #222: Routed export silently re-encodes from source when an allocation carries no projection at all
**Severity: P3**, `needs-decision`. Filed by the coordinator from worker `pq-183`'s report on PR #220 (audit campaign 2026-09-05); nothing here is a regression introduced by that PR.

## What

PR #220 makes the producer's expert projection an **unlock, not a requirement**: an allocation that carries no `tessera.expert_projection.v1` record keeps the pre-#183 lane byte for byte (routed units resolve on source-member shapes, a predicated `routed_moe` cell is still refused for lacking the producer's attestation). That call is deliberate and correct — making it a requirement broke nine `test_tessera_export_scope.py` tests and would have killed the qwen3 lane.

`require_assignment_scope` refuses the **incoherent** case: priced wires, `tessera_expert_stack_formats`, or a wire directory carried with the projection stripped out — a receipt bound to no executed unit.

The hole it leaves: an allocation with **all four** keys stripped (projection, priced wires, stack formats, wire directory) falls back silently to re-encoding the routed units from source. Nothing refuses, and nothing says the bytes about to ship were not the bytes the campaign priced. That is exactly the "priced == written == served" seam #183 and #204 were filed against, reachable by removing four keys from a JSON file rather than by editing one.

## Why it is only P3 today

It is the pre-#220 behaviour, not a new gap: before #220 no allocation carried a projection at all, so every routed export re-encoded from source. Nothing shipped is wrong, and the operator would have to strip a coherent set of four keys to reach it. It is filed so the fallback is a decision on the record rather than an accident.

## Options

- **(a) Leave it.** The fallback is the documented pre-#183 lane; a stripped allocation is an operator editing receipts, which no gate can fully defend against.
- **(b) Refuse the fallback for any assignment whose units include a routed expert the profile splits**, once a producer tool for that profile exists — i.e. make "this profile has a producer" the trigger rather than "this allocation carries a projection". Costs: any lane whose profile gains a producer tool must re-run its campaign before it can export again; the qwen3 lane is the one to check first.
- **(c) Record the fallback instead of refusing it**: stamp the export receipt with "routed units re-encoded from source, not from priced wires" so a consumer can see which path produced the bytes. Cheapest, and it makes the seam readable without breaking a lane.

My recommendation is (c) now and (b) when every live routed profile has a producer tool, but the promotion is Rob's — it decides whether an existing lane must re-run a campaign before its next export.

## Refs

RobTand/prismaquant#183 (acceptance items 1–3), PR #220 (`require_assignment_scope`, `_carried_expert_projection`), #204 (the export-input binding that closed the analogous seam for the Hessian and the static scales).

### LINKED ISSUE #229: [P1] Skip expert-wire bundling when the allocation keeps all experts in BF16
At PrismaQuant `17ca693071f52ed7b69cbc85ce8681207a968c02`, the new PR220 packed-expert bridge rejects an allocation that selects Tessera for a dense Linear and keeps every routed expert in BF16. The allocator explicitly supports that decision, but the normal driver's export preflight attempts to serialize an empty expert-wire bundle.

The producer and consumer disagree about what a retained projection means:

- `prismaquant/tessera_expert_projection.py:521-599`, `allocation_expert_projection_block`, carries the population and projection even when a stack is assigned a non-Tessera format. It records that stack as BF16 and carries an empty `tessera_expert_wires` mapping. This behavior is intentional and tested at `tests/test_allocator_expert_projection.py:265-268`.
- `prismaquant/tessera_export_lane.py:545-572` filters the assignment to Tessera rows. With all experts BF16, the selected routed subset is empty. `_carried_expert_projection` nevertheless returns a non-null report with empty `units` and `stacks` at lines 466-471.
- `preflight` tests only whether that projection is non-null at lines 1092-1097. With `--write-cached-expert-units`, it calls the bundle writer, which passes the empty unit map into `cached_units_manifest`. That function refuses `no priced expert wires to bundle` at `tessera_expert_projection.py:501-502`.
- `prismaquant/run-pipeline.sh:2520-2525` always supplies this flag on the Tessera export path. The valid mixed allocation therefore exits 2 before publishing its build anchor. `preflight`'s own documented contract at lines 1063-1065 says an allocation selecting no routed expert unit bundles nothing.

CPU proof: `/home/rob/tessera-runs/audit-pass3-2026-09-05/proofs/pq_bf16_expert_bundle.py`, `.json` and `.log`, exit 0. It uses PR220's existing local export fixture, the actual `allocation_expert_projection_block`, actual assignment-scope resolver and actual export CLI. The allocator helper produces the retained projection, stack format BF16 and zero wire receipts; only the dense Linear resolves to Tessera. The normal CLI flag then returns 2 with the named empty-bundle refusal and no build file. Two controls return 0 and write a build without any bundle: the same allocation without the bundle flag, and the existing test's variant that strips all projection metadata.

The proof is a small integration check, not a model/serving result. It uses the existing test's header-only checkpoint and synthetic eligible serving table, plus its isolation of unrelated release/producer checks and pin loader. It imports the real Tessera cached-unit schema. Actual projection, allocator metadata, route selection and bundle/CLI behavior are not stubbed. Native math threads are one and CUDA is hidden. No GPU, model encoding, full suite or PrismaBuild was run.

Test gap: `test_cli_writes_no_bundle_for_an_allocation_without_a_projection` at `tests/test_tessera_export_projection.py:399-412` changes experts to BF16 but also deletes projection/wire/stack/directory metadata. A real allocation produced by the new allocator retains those fields, so that passing control does not cover the producer-to-consumer seam.

This is distinct from existing #222, which concerns a routed Tessera selection whose allocation has no projection and silently re-encodes. This finding has a valid carried projection and **no selected Tessera experts**. It does not ask to weaken the receipt requirements for any selected expert or claim that the pending release pin is already qualified.

Acceptance: keep the allocator's population/projection provenance and make zero selected Tessera experts mean no bundle and no `--cached-expert-units` handoff. Exercise the metadata emitted by the actual allocation helper in the CLI regression. Keep the positive selected-expert bundle case, a mixed BF16/Tessera-stack case, and missing/damaged selected-wire refusals. Do not fix this by discarding projection provenance or permitting empty manifests for paths that require expert wires.

Audit only: no source fix was made; implementation is deferred to the separate issue dispatch.

Disposition: filed during the requested review of closed audit issues and current code. No implementation changes were made; fixes are left for a separate dispatch.

RobTand:
Fixes #222 — option **(c) only**: the re-encode-from-source fallback becomes readable in the receipt. It is **not** refused, and nothing about what is refused or what is encoded changes.

## What was wrong

#220 made the producer's expert projection an **unlock, not a requirement**. An allocation carrying none keeps the pre-#183 lane byte for byte, and `require_assignment_scope` refuses only the *incoherent* case (priced wires, `tessera_expert_stack_formats`, or a wire directory carried with the projection stripped). That call stands.

What was wrong is that the remaining path was **silent**: an allocation with all four keys absent re-encodes its routed units from source, and nothing in the export's own output said the bytes about to ship were not the bytes the campaign priced. Option (b) — refusing the fallback once a profile has a producer tool — would force every such lane to re-run its campaign before its next export, and stays Rob's to price.

## The field

One value, one derivation, two placements:

| where | key | who reads it |
|---|---|---|
| `require_assignment_scope`'s scope receipt | `routed_expert_bytes` | `preflight()["selected_serving_scope"]` |
| the build anchor (`--write-build-json`) | `tessera_routed_expert_bytes` | `lane_shipcard open --build-json` → the artifact's `shipcard.json` |

The anchor **copies** the receipt's answer, it does not recompute it. The anchor is the only machine-readable thing the CLI writes, and the shipcard is where a consumer of the *shipped* bytes looks — so no second receipt was invented.

Three values, as module constants: `priced_wires`, `reencoded_from_source`, and `no_routed_units` for an allocation that selects no routed unit at all. Without the third, every dense export would read as a fallback and the field would be worthless.

## Derived, never passed in

`_carried_expert_projection` now returns `(routed_expert_bytes, bundle)` instead of `dict | None`. It is the only function that sees both the carried keys and whether any routed unit was selected, so it is the only place that can answer all three; deriving it at the call site from `bundle is not None` would be a second rule for one question and would mislabel a dense export. **One caller, and it moves**: `require_assignment_scope` (`tessera_export_lane.py:571`). `grep -rn _carried_expert_projection` outside `archive/` finds the definition, that call, and two prose mentions — no test calls it, so nothing is left destructuring the old shape.

## Compatibility, both directions

* **Old anchor → new reader**: the key is absent. Absence means *"this preflight predates #222 and does not say"* — **never** `priced_wires`. Every artifact built before this commit is in that state, and most of them re-encoded. This rule is written into `docs/ARCHITECTURE.md`, not just the commit, because it is what a future reader of an old shipcard needs.
* **New anchor → older reader**: harmless, because the build block is an **open schema**. `shipcard.build_shipcard` stores it verbatim (`dict(build or {})`); `lane_shipcard.open_lane_shipcard` only `setdefault`s `export_container`; `shipcard._verify_build_block` checks named keys one at a time and says so itself — *"Every check fires only on a key the producer stamps, so a card written before a key existed is left alone"*; `tools/publish_artifact.py` requires only that `build` is a Mapping. There is no key whitelist and no exact key-set comparison anywhere over the build dict, `lane_specs/tessera.json` declares no build keys, and `tests/test_shipcard.py` already round-trips arbitrary build keys. So there was no closed schema to extend, and an unknown key is carried and ignored. The 256 KiB card reservation is unaffected by ~50 bytes.

No test asserts an exact key set on either the scope receipt or the build anchor.

## Regression, shown failing on main first

Pre-fix on `17ca6930` with only the new tests applied — `3 failed, 15 passed, 1 skipped`, each with `AttributeError: module 'prismaquant.tessera_export_lane' has no attribute 'ROUTED_EXPERT_BYTES_KEY'`. Main's receipt does not carry the fact at all, which *is* the defect.

* `test_the_receipt_names_which_path_produced_the_routed_bytes` — a carried projection stamps `priced_wires`; stripping only the projection **still refuses by name** (#220's behaviour, unweakened); stripping all four stamps `reencoded_from_source`, still succeeds, and produces an **identical `by_unit` route table** to the priced run. Only the receipt differs.
* `test_an_export_that_ships_no_routed_bytes_is_not_stamped_as_a_re_encode` — dense-only stamps `no_routed_units`, with and without the carried keys.
* `test_the_build_anchor_carries_which_path_produced_the_routed_bytes` — the CLI's `build.json` carries the value on both paths.

`docs/ARCHITECTURE.md` re-stamped in the same commit (principle 12): header, a new top stamp, and a new §5 export-lane bullet naming the fallback, the three values, where each key lives, what absence means, and that option (b) is open and Rob's.

---

# Second commit — Fixes #229 (P1)

`3c3ef4fc`, on top of the merge of `origin/main` at `eda5b37d` (#225).

**A valid mixed allocation was refused by its own export gate.** Selecting Tessera for a dense Linear while every routed expert stays BF16 is a decision the allocator is built to emit — `allocation_expert_projection_block` keeps the population and the projection, records the stack as `BF16`, and carries an empty `tessera_expert_wires` map (`tests/test_allocator_expert_projection.py:265-268`). The export lane filters the assignment to Tessera rows, so the selected routed subset is empty, yet `_carried_expert_projection` still returned a projection with empty `units`/`stacks` and `preflight` tested only non-null-ness — handing that empty map to the bundle writer, whose `cached_units_manifest` refuses `no priced expert wires to bundle`. `run-pipeline.sh:2520-2525` **always** passes `--write-cached-expert-units`, so the allocation exited 2 and published no build anchor, against `preflight`'s own documented contract.

**The fix is one predicate, and it is the first commit's field.** Bundle exactly when `routed_expert_bytes == priced_wires`, not when a projection merely exists. The two issues resolve through one value:

| case | value | bundles? |
|---|---|---|
| routed Tessera units, **no** carried projection — re-encoded from source (#222) | `reencoded_from_source` | no |
| valid carried projection, **no selected Tessera expert** (#229) | `no_routed_units` | no |
| routed Tessera units under a carried projection | `priced_wires` | yes — the only one |

Since `priced_wires` requires at least one selected routed unit, and each such unit either yields a verified receipt or refuses by name, the writer can no longer be reached with an empty unit map.

**Pre-fix reproduction.** On the merged tip with only the new test applied: `AssertionError: assert 2 == 0`. This checkout's pinned Tessera (1221d2a) has no `tessera.cached_unit`, so the local refusal is #192's import error; injecting only that schema constant and re-running the same test reproduces the issue's own message exactly — `expert projection: no priced expert wires to bundle`. Both are the same defect and both are gone after the fix. The issue's CPU proof on sparky (`pq_bf16_expert_bundle.py`) was read, not modified, and its construction is what the new test reuses.

**The regression exercises the allocator's own metadata** — the point of #229. The existing control `test_cli_writes_no_bundle_for_an_allocation_without_a_projection` sets experts to BF16 *and* deletes the projection/wire/stack/directory keys (the stock-table shape), so it never covered the producer-to-consumer seam; it is kept and now says which shape it covers. The new `test_cli_bundles_nothing_when_the_allocation_keeps_every_expert_in_bf16` builds its metadata by calling `allocation_expert_projection_block` itself, asserts the block really carries empty wires and `{stack: "BF16"}`, and drives the actual CLI with the flag the driver always passes.

Kept per the acceptance criteria: the positive selected-expert bundle case (skips here for want of `tessera.cached_unit`, so also run with the schema constant injected — 20 passed, bundle unchanged); the missing/damaged selected-wire refusals; the projection provenance (the test asserts population, projection and the BF16 stack stamp all survive); and `cached_units_manifest`'s empty-bundle refusal for paths that genuinely require wires.

**No currently-working export path changes behaviour**: a run that bundled still bundles the same manifest to the same path, a run that wrote no bundle still writes none. The only difference is that a case which exited 2 now exits 0.

Counts after both commits: targeted set `290 passed, 4 skipped`; full suite `17 failed, 5032 passed, 138 skipped, 3 xfailed` — the 17 known-inherited `test_tessera_campaign_resume.py` failures, nothing else.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,37 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-05 · `claude/pq-218`. Stamps
+As of: 2026-09-05 · `claude/pq-222`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-05, `claude/pq-222`) for **the routed re-encode fallback
+being readable in the receipt** (§5 export lane; RobTand/prismaquant#222, P3).
+#220 made the producer's expert projection an UNLOCK, not a requirement: an
+allocation carrying none keeps the pre-#183 lane byte for byte, and
+`require_assignment_scope` refuses only the INCOHERENT case (priced wires,
+`tessera_expert_stack_formats` or a wire directory carried with the projection
+stripped). That call stands and is unchanged here — the fallback is not
+refused, no unit's bytes move, and making the projection mandatory (#222's
+option (b)) remains Rob's to price, because it would force every profile that
+gains a producer tool to re-run its campaign before it could export again.
+What was wrong is that the fallback was **silent**: an allocation with all
+four keys absent re-encodes its routed units from source and nothing said the
+bytes about to ship were not the bytes the campaign priced. It now says so.
+`_carried_expert_projection` returns the ANSWER beside the bundle —
+`priced_wires`, `reencoded_from_source`, or `no_routed_units` when the
+allocation selects no routed unit at all, so a dense export is never
+mislabelled a fallback — because it is the only function that sees both the
+carried keys and the selection, and a receipt derived anywhere else could
+disagree with the path taken. `require_assignment_scope` stamps it as
+`routed_expert_bytes` in the scope receipt and `preflight` copies it into the
+build anchor as `tessera_routed_expert_bytes`, whence `lane_shipcard open
+--build-json` carries it onto the artifact's ship record. **A reader of the
+old shape sees neither key, and that absence means "this preflight predates
+#222 and does not say" — never `priced_wires`.** New anchors stay readable by
+older consumers: the build block is an open schema (`build_shipcard` stores it
+verbatim, `_verify_build_block` checks only keys it knows and states that a
+card written before a key existed is left alone), so an unknown key is
+carried and ignored. No default, format menu, serving lane, ship gate or byte
+changes.
 
 Re-stamped (2026-09-05, `claude/pq-218`) for **format-name resolution: case is
 the registry's question, and a predicate over a format menu must be total**
@@ -6164,6 +6194,28 @@ Two properties make the numbers comparable with the rest of the menu:
   blobs instead of re-encoding the source. An allocation with no routed
   expert unit bundles nothing and the encode is byte-identical to one built
   before this existed.
+
+* **And the receipt names which path produced the routed bytes (PrismaQuant
+  #222).** The projection is an unlock, not a requirement, so an allocation
+  carrying none is not refused — its routed units resolve on source-member
+  shapes and the exporter **re-encodes them from the source checkpoint**.
+  That lane is sanctioned and unchanged, but it does not ship the bytes the
+  campaign priced, so it is stated rather than inferred.
+  `_carried_expert_projection` returns which path it took beside the bundle —
+  it is the only function that sees both the carried keys and whether any
+  routed unit was selected — and `require_assignment_scope` stamps that answer
+  into its receipt as `routed_expert_bytes`: `priced_wires`,
+  `reencoded_from_source`, or `no_routed_units` for an allocation that selects
+  none, so a dense export never reads as a fallback. `preflight` copies the
+  same value into the build anchor as `tessera_routed_expert_bytes`, and
+  `lane_shipcard open --build-json` stamps the anchor whole onto the
+  artifact's ship record, so a consumer of the shipped bytes reads it there.
+  **Absence of the key means the preflight predates #222 and does not say; it
+  never means `priced_wires`** — every artifact built before this stamp is in
+  that state, and most of them re-encoded. Whether the fallback should instead
+  be REFUSED once a profile has a producer tool is #222's option (b), open and
+  Rob's to price: it would make every lane whose profile gains one re-run its
+  campaign before its next export.
 
 Rows are written in the codebase's own currency: `output_mse` and **no**
 `predicted_dloss` field, because in this tree `output_mse` is a raw MSE while
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,31 @@
 As of: 2026-09-05 · `claude/pq-222`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-05, `claude/pq-222`) for **an allocation that keeps every
+routed expert in BF16 no longer being refused by its own export gate** (§5
+export lane; RobTand/prismaquant#229, P1). Selecting Tessera for a dense
+Linear while every routed expert stays BF16 is a decision the allocator is
+built to emit: `allocation_expert_projection_block` keeps the population and
+the producer's projection, records the stack's format as `BF16`, and carries
+an empty `tessera_expert_wires` map. The export lane filters the assignment
+to Tessera rows, so the selected routed subset is empty — yet
+`_carried_expert_projection` still returned a projection (with empty `units`
+and `stacks`) and `preflight` tested only that it was non-null, handing the
+empty map to the bundle writer, whose `cached_units_manifest` refuses `no
+priced expert wires to bundle`. Because `run-pipeline.sh` **always** passes
+`--write-cached-expert-units` on the Tessera export path, a valid mixed
+allocation exited 2 and published no build anchor, against `preflight`'s own
+documented contract that an allocation selecting no routed expert unit
+bundles nothing. The predicate is now the receipt's own value from the stamp
+below: the bundle is written exactly when `routed_expert_bytes` is
+`priced_wires`, so an anchor names a bundle in precisely the runs whose bytes
+come from one. The projection provenance is kept, not discarded, and
+`cached_units_manifest` still refuses an empty bundle where one is genuinely
+required. This is the case #222's third value (`no_routed_units`) exists to
+name, and it is distinct from #222's own: #222 is a routed **Tessera**
+selection carrying no projection, which silently re-encodes; #229 has a valid
+carried projection and **no selected Tessera expert at all**.
+
 Re-stamped (2026-09-05, `claude/pq-222`) for **the routed re-encode fallback
 being readable in the receipt** (§5 export lane; RobTand/prismaquant#222, P3).
 #220 made the producer's expert projection an UNLOCK, not a requirement: an
@@ -6273,7 +6298,15 @@ Two properties make the numbers comparable with the rest of the menu:
   `priced == written` on the routed leg: the exporter reuses the campaign's
   blobs instead of re-encoding the source. An allocation with no routed
   expert unit bundles nothing and the encode is byte-identical to one built
-  before this existed.
+  before this existed — **including one that carries the producer's projection
+  and keeps every routed expert in BF16** (#229). That allocation is one the
+  allocator emits by design (`allocation_expert_projection_block` records the
+  stack as BF16 with no wire receipts), and the gate refused it with exit 2
+  until the bundle was made conditional on `routed_expert_bytes ==
+  priced_wires` rather than on a projection merely being carried. Zero
+  selected Tessera experts means no bundle and no `--cached-expert-units`
+  handoff; where a bundle IS required, `cached_units_manifest` still refuses
+  an empty one.
 
 * **And the receipt names which path produced the routed bytes (PrismaQuant
   #222).** The projection is an unlock, not a requirement, so an allocation
--- a/prismaquant/tessera_export_lane.py
+++ b/prismaquant/tessera_export_lane.py
@@ -1212,7 +1212,10 @@ def preflight(model_path: str | Path, *, target=None,
     bundle for the priced expert wires this allocation selected, into the
     campaign's own wire directory, and names it in the build anchor so the
     driver hands the exporter the path rather than reconstructing it.  An
-    allocation that selects no routed expert unit bundles nothing.
+    allocation that selects no routed expert unit bundles nothing -- including
+    one that CARRIES the producer's projection and keeps every routed expert
+    in BF16, which is a decision the allocator emits and this gate refused
+    until #229.
     """
     structure = require_declared_structure(model_path)
     target = require_serving_target(target)
@@ -1250,7 +1253,23 @@ def preflight(model_path: str | Path, *, target=None,
             projection = scope.get("expert_projection")
             if projection is not None:
                 build["tessera_expert_stack_formats"] = dict(projection["stacks"])
-                if cached_expert_units:
+                # Bundle exactly when priced wires are what produce the routed
+                # bytes -- not merely when a projection is carried (#229).  An
+                # allocation that keeps every routed expert in BF16 and selects
+                # Tessera only for a dense Linear is one the allocator is
+                # designed to emit: `allocation_expert_projection_block` keeps
+                # the population and the projection and records the stack as
+                # BF16, so the projection here is non-null with no selected
+                # unit under it.  Testing non-null-ness handed that empty map
+                # to the bundle writer, which refuses "no priced expert wires
+                # to bundle" -- exit 2, no build anchor, on the normal driver
+                # path, which always passes the flag.  The receipt's own value
+                # is the predicate, so the anchor names a bundle in exactly the
+                # runs whose bytes come from one, and `cached_units_manifest`
+                # keeps refusing an empty bundle where one IS required.
+                if cached_expert_units and (
+                        scope[ROUTED_EXPERT_BYTES_KEY]
+                        == ROUTED_EXPERT_BYTES_PRICED_WIRES):
                     build["cached_expert_units"] = str(
                         write_cached_expert_units(projection))
         if file_sha256(assignment_path) != assignment_sha:
--- a/prismaquant/tessera_export_lane.py
+++ b/prismaquant/tessera_export_lane.py
@@ -46,6 +46,13 @@ which is the failure mode principle 14 exists to prevent.
    Tessera unit must retain the allocation's target and per-unit context,
    agree with the source header and profile topology, and resolve all regimes
    on that context. Legacy calls without scope retain their existing gates.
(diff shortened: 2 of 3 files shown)


RobTand:
Commands and counts, run on sparklina in a real git worktree (`/home/rob/prismaquant-audit-wt/pq-222`, branched from `origin/main` at `17ca6930`), CPU-only, against the pinned development Tessera (1221d2a).

**Pre-fix line** — the three new tests applied to unmodified `main` code:

```
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider \
  tests/test_tessera_export_projection.py -q
```
`3 failed, 15 passed, 1 skipped` — all three `AttributeError: module 'prismaquant.tessera_export_lane' has no attribute 'ROUTED_EXPERT_BYTES_KEY'` / `BUILD_ROUTED_EXPERT_BYTES_KEY`. Main's receipt does not carry the fact at all.

**Targeted set** (the brief's list plus every suite that greps the receipt schema, the build anchor, or the shipcard block this change touches):

```
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider -n 2 \
  tests/test_tessera_export_lane.py tests/test_tessera_export_scope.py \
  tests/test_tessera_export_projection.py tests/test_allocator_expert_projection.py \
  tests/test_tessera_campaign_packed.py tests/test_architecture_doc.py \
  tests/test_docs_staleness.py tests/test_tessera_scoped_route_receipt.py \
  tests/test_lane_gate_recording.py tests/test_shipcard.py -q
```
`250 passed, 3 skipped`

**Full suite**:

```
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider -n 2 -q
```
`17 failed, 5013 passed, 138 skipped, 3 xfailed, 174 subtests passed in 488.82s`

The 17 are the known-inherited `test_tessera_campaign_resume.py` main-entry failures on PQ main — exactly 17, nothing else, and none in a file this PR touches.

**Evidence for "adding a key to a published receipt is safe"**, since that is a claim rather than an assumption:

```
grep -rn '\["build"\]\|get("build"' --include=*.py .   # every reader of the block
grep -rn '_require_keys\|allowed_keys\|set(build)\|set(payload)' prismaquant/shipcard.py \
  prismaquant/lane_shipcard.py prismaquant/lane_spec.py
grep -rn 'build' prismaquant/lane_specs/tessera.json
grep -rn '_carried_expert_projection' --include=*.py .  # one call site
```

No key whitelist, no exact key-set comparison, no build-key declaration in the lane spec, and one caller of the function whose return type changed. `shipcard._verify_build_block` states its own tolerance in its docstring: *"Every check fires only on a key the producer stamps, so a card written before a key existed is left alone."*

RobTand:
Second commit (`3c3ef4fc`, Fixes #229) — commands and counts, on the merge of `origin/main` at `eda5b37d`. Same box and env as above: sparklina, real git worktree, CPU-only, tessera pin 1221d2a.

**Pre-fix line** — the new test applied to the merged tip without the fix:

```
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider \
  "tests/test_tessera_export_projection.py::test_cli_bundles_nothing_when_the_allocation_keeps_every_expert_in_bf16" -q
```
```
E       AssertionError: assert 2 == 0
[preflight] ERROR: EXPORT_CONTAINER=tessera: cannot bundle the priced expert wires:
this checkout's tessera has no cached_unit bundle API (tessera.cached_unit.CACHE_SCHEMA)
```

This checkout's pinned Tessera has no `tessera.cached_unit`, so the local refusal is #192's import error. Injecting **only** that schema constant into `sys.modules` and re-running the *same* test reaches the writer and reproduces the issue's own message:

```
[preflight] ERROR: EXPORT_CONTAINER=tessera: expert projection: no priced expert wires to bundle
```

Same defect both ways — `preflight` calls the bundle writer for an allocation with zero selected experts — and both are gone after the fix. I did not modify anything under the sparky proofs directory; `pq_bf16_expert_bundle.py`/`.json` were copied out and read.

**Post-fix, same file** (`19 passed, 1 skipped` — the skip is the positive bundle case, which needs the release producer's `cached_unit` API):

```
... -m pytest -p no:cacheprovider tests/test_tessera_export_projection.py -q
```

And with the schema constant injected so that skipped positive case actually executes — proving the path that DOES bundle is unchanged:

```
20 passed
```

**Targeted set** (as before, plus `tests/test_tessera_priced_export_inputs.py` for #225's roster, which landed in this file in the merge):

```
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider -n 2 \
  tests/test_tessera_export_lane.py tests/test_tessera_export_scope.py \
  tests/test_tessera_export_projection.py tests/test_allocator_expert_projection.py \
  tests/test_tessera_campaign_packed.py tests/test_tessera_priced_export_inputs.py \
  tests/test_architecture_doc.py tests/test_docs_staleness.py \
  tests/test_tessera_scoped_route_receipt.py tests/test_lane_gate_recording.py \
  tests/test_shipcard.py -q
```
`290 passed, 4 skipped`

**Full suite**:

```
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider -n 2 -q
```
`17 failed, 5032 passed, 138 skipped, 3 xfailed, 174 subtests passed in 524.15s`

The 17 are the known-inherited `test_tessera_campaign_resume.py` main-entry failures — exactly 17, none in a file this PR touches.

**ARCHITECTURE.md merge**: `origin/main`'s three `claude/pq-gates` stamps conflicted with mine. Resolved to one `As of: 2026-09-05 · claude/pq-222` header line followed by all stamp paragraphs newest-first — #229, then #222, then the three pq-gates stamps, then pq-218, pq-211 and the rest. No stamp was dropped.

RobTand:
Merged `origin/main` at `128fa1bb` (#226) — tip is now `3f7c1534`, `MERGEABLE`. Re-run **targeted only**, on **sparky** (sparklina was loaded), in a real git checkout.

```
rsync -a --delete --exclude '.git' /home/rob/prismaquant-audit-wt/pq-222/ sparky:/home/rob/audit-sync/pq-222/
ssh sparky 'cd /home/rob/audit-sync/pq-222 && git init -q && git add -A && \
  git -c user.email=a@b -c user.name=c commit -qm sync && \
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider -q -n 4 \
    tests/test_tessera_export_projection.py tests/test_tessera_export_scope.py \
    tests/test_tessera_export_lane.py tests/test_tessera_priced_export_inputs.py \
    tests/test_allocator_expert_projection.py \
    tests/test_static_activation_contract_one_home.py \
    tests/test_tessera_campaign_packed.py tests/test_tessera_scoped_route_receipt.py \
    tests/test_lane_gate_recording.py tests/test_shipcard.py \
    tests/test_architecture_doc.py tests/test_docs_staleness.py'
```
`302 passed, 4 skipped in 10.51s`

Selection, and why each file is in it: the three suites over the module the diff changes (`export_projection`, `export_scope`, `export_lane`); `priced_export_inputs` and `static_activation_contract_one_home` because #226 rewrote `require_priced_export_inputs` in the same file and both sides now assert in it; `allocator_expert_projection` because the new #229 regression calls the producer-side helper `allocation_expert_projection_block`; `scoped_route_receipt`, `lane_gate_recording` and `shipcard` as the reverse-dependents of the build block the new key is stamped into; `campaign_packed` for the packed-expert path; and the two doc gates for the re-stamp. Nothing else reads `require_assignment_scope`, `_carried_expert_projection`, `write_cached_expert_units` or the build anchor.

The positive bundle case skips everywhere for want of `tessera.cached_unit` at pin 1221d2a, so it was also run on sparky with only that schema constant injected, to prove the path that DOES bundle is unchanged by the #229 predicate:

```
20 passed
```

**Merge resolution**: only `docs/ARCHITECTURE.md` conflicted (`tessera_export_lane.py` auto-merged — #226 is in `require_priced_export_inputs`, this PR is in `require_assignment_scope` / `_carried_expert_projection` / `preflight`). Resolved to one `As of: 2026-09-05 · claude/pq-222` header line and all stamp paragraphs newest-first: #229, #222, then `claude/pq-221`, the three `claude/pq-gates`, `claude/pq-218`, `claude/pq-211` and the rest. No stamp dropped, none reordered among themselves.

PR #230, opened 2026-09-05
Bind cached activation-aware costs to the resolved activation-scale policy (#227)

### LINKED ISSUE #227: [P1] Bind Tessera cached scores and validation to the resolved activation-scale policy
Tessera's static activation scale is recomputed from the current process environment, but that policy is not bound to the production cache whose scores and activation maxima are reused. Changing the documented scale policy can therefore retain scores priced at G=1 while subsequent assignment-KL hooks use G=448 for the same cache and calibration maximum.

Audited the frozen PrismaQuant `dd48f62f9e6a7f6914e133174e89c0400eebfcdd` snapshot with Tessera `f86b81162a3ee968423a2669918c403ff3f44a5a`, then checked both deltas through `553de0afe7e2bd123362fbd55a0bb88b57c1ec42` (PR212) and the frozen current main `17ca693071f52ed7b69cbc85ce8681207a968c02` (PR220/223/224). All anchors below name `17ca693071f52ed7b69cbc85ce8681207a968c02`. PR224 changes format-name resolution and predicate handling, but the numerical contract, scorer, resolved-lever builder, cache identity builder, hook and KL functions remain AST-identical to the proof snapshot; the existing-score reuse condition is unchanged. This is a residual of #205/PR213's typed activation-contract integration; the original dynamic-versus-served callback bug is fixed. It is distinct from #204, whose newly merged PR212 binds externally supplied capture/scale files to an allocation, and from #211 (writers before resume validation, now fixed by PR223) and open #216 (capture-context roster guard).

The concrete path is:

- `nvfp4_activation_contract.py:1231-1235` implements `StaticActivationContract.input_global_scale_from_max_abs` by resolving the process's current policy on every call. The object carries execution/group-size/measurement mode, but no frozen scale policy. `PRISMAQUANT_NVFP4_INPUT_GSCALE_FP8_RANGE=0` uses 6/amax; setting it to 1 uses 448*6/amax.
- `production_weight_cache._render_score_record` derives G through that method at `production_weight_cache.py:1483-1488` and stores the resulting score and `input_global_scale` at `:1545-1546`. However, the cache identity emitted at `:5185-5204` does not bind the resolved activation policy; its actual resolved render levers also omit this policy.
- The disk-resume branch at `production_weight_cache.py:7179-7182` reuses the existing rendered weight and keeps an existing score record by its qname/format key. A changed activation policy neither invalidates that score nor produces an identity mismatch. The score is recomputed only when its key is absent.
- `PerturbedActivationCache` takes only the cached activation maxima at `perturbed_x_cache.py:458-466`. Its hook at `:139-143` derives G from the current policy again; it does not read or compare the G recorded with the cached score. The new missing-scale preflight only establishes that a maximum exists, which remains true across the policy change.

Trigger: fill/price a Tessera W4A4 production cache under the legacy policy, retain it, then resume or validate the same cache after selecting the documented full-E4M3-range policy. Weights, calibration hash, maximum, qname and format can all remain identical; only the numerical meaning of that maximum changes. An identical directory identity admits a score from the earlier operation while live validation executes the later operation. Reusing weights can be valid, but reusing their old activation-aware score under the new policy is not.

Bounded CPU proof: `/home/rob/tessera-runs/audit-pass3-2026-09-05/proofs/pq_activation_policy_identity.py` and `.json`, final exit 0, torch 2.10.0+cpu, CUDA hidden and OMP/MKL/OpenBLAS/NumExpr/torch intra-op/inter-op threads 1. The script calls the actual resolved-lever and mechanism-plan owners, actual cache identity builder, actual scorer and actual hook helper on the real synthesized Tessera spec. Only its producer-eligibility predicate is stubbed to False because the frozen installed Tessera v8 table is intentionally refused by PrismaQuant's old pin/parser. That stub grants no serving admission and changes no numerical or identity code. An initial probe stopped at the expected v8 admission refusal because the stub ended too soon; extending only its lifetime produced the recorded proof.

For `TESSERA_E2M1_K2_R896`, amax=6, BF16 `[1e-3]*16` and W=rendered-W=ones(1,16):

```text
policy 0: G=1,   score=0.00025571882724761963, hook QDQ=0
policy 1: G=448, score=5.622899834634154e-07, hook QDQ=0.00104522705078125
cache render identity under policy 0 == identity under policy 1: True
```

Each scorer now agrees with the served oracle under its own policy; that is a positive control for the original #205 fix. The defect is that persisted score reuse and later validation can use different policies without a refusal. The CPU proof establishes identity collision and the changed numerical result. The persisted-score reuse consequence is traced through the actual resume branch; no complete cache fill/resume, encoded unit, model-level KL, native GPU or served result is claimed. Tessera's release/menu admission remains independently fail-closed at the intentional pin boundary; this finding concerns the supported research/cache/validation APIs, not newly qualified production serving.

Test gap: `tests/test_served_activation_contract.py` verifies hooks and scores under one current policy and checks missing maxima, but does not retain an identity/score across a policy change or compare a cache record's G with the hook's effective G. PR212's export-file seals do not touch these cache or hook paths.

Acceptance: carry the resolved activation-scale policy and effective per-unit/fused G through the existing cache identity and score provenance, and make measurement consumers use that bound meaning. If a caller requests a changed policy, refuse stale score reuse or explicitly invalidate/recompute the activation-aware scores before accepting them; do not reinterpret cached maxima while retaining their previous costs. Compare the bound policy/G when constructing assignment-KL hooks. Reuse the current StaticActivationContract and cache mechanisms; preserve stock NVFP4's existing screen-default policy. Add a small regression with a persisted old score plus changed policy, an unchanged-policy control, and fused-sibling values showing the cache/scorer/hook all use one effective G.

The user requested an audit and issue filing only, so no source fix was made. No GPU, full suite, PrismaBuild, model job, or unrelated process action was performed.

Disposition: filed during the requested review of closed audit issues and current code. No implementation changes were made; fixes are left for a separate dispatch.

RobTand:
Fixes #227.

## The defect

`StaticActivationContract.input_global_scale_from_max_abs` resolved the process's **current**
input-global-scale policy on every call, and nothing a production cache persisted named the
policy its activation-aware costs were priced under. The same calibration maximum derives
`G = 6/amax` under `legacy_6_over_calibration_amax.v1` and `G = 448*6/amax` under
`full_e4m3_range_448x6_over_calibration_amax.v1`, and the served oracle at those two G is two
different quantisers. Measured on `TESSERA_E2M1_K2_R896`, amax=6, a BF16 `[1e-3]*16` block,
W = rendered-W = ones(1,16):

```
policy 0: G=1,   score=0.00025571882724761963, hook QDQ=0
policy 1: G=448, score=5.622899834634154e-07,  hook QDQ=0.00104522705078125
```

Each is right under its own policy — that is #205's positive control. What was wrong is that
the two **cache render identities were equal**, so a directory filled and priced under one
policy resumed with its scores intact (the disk-resume branch keeps an existing score by
`qname|FMT` key) and was then KL-validated under the other. Reusing the weights is legitimate;
reusing their activation-aware cost under another policy is not.

## The fix — one rule, one home

**Home: the resolved render levers.** `_resolve_production_render_levers` is already the single
place every env-valued render input (`nvfp4_scale_rule`, the fixed damp, act-order, JSO) becomes
a stamped value shared by the resident and streaming fills, and this policy already reached
`render_production_weight` as the fill's `input_global_scale`. It joins them as
`nvfp4_input_global_scale_policy` and travels, for free, into `render_identity.json`, the CB
render contract, the union per-shard identity and `ProductionWeightCache.levers` — so a policy
change is an identity mismatch naming that field instead of a silent resume.

The contract object is deliberately **not** frozen: a screen, a fresh fill and the export
legitimately read the live setting, so its G rule still resolves live by default and takes an
explicit `policy=` only from a caller that already resolved one for a whole operation.

Two further lines, because the identity gate admits a pre-guard directory (no sidecar) on trust
and a validation run may hold a cache it did not fill:

- every render score records `input_global_scale_policy` beside its `input_global_scale`, and
  each retained cost's recorded G is checked against the G the current policy derives from the
  **same recorded maximum** — so records written before this stamp are checked too — refusing by
  name with `ActivationScalePolicyMismatchError` (a subclass of the #205 refusal). Dense,
  streamed and packed-expert resumes go through that one check;
- the assignment-KL hook compares the G the cache priced a unit at against the G it is about to
  apply, and `measure_assignment_kl` preflights it
  (`served_activation_policy_conflicts`, the sibling of the missing-maximum gate) so the refusal
  lists every affected unit before the first forward. Having a maximum survives a policy change
  unchanged — which is exactly why the earlier gate could not see this.

Rows priced under a dynamic quantiser (FP8/MX, and stock NVFP4 on its screen default) record no
policy and are invalidated by none. No default, format menu, serving lane, ship gate or rendered
byte moves; both policies compute exactly what they computed before.

## Operator note — caches on disk

An existing production cache directory's `render_identity.json` carries no
`levers.nvfp4_input_global_scale_policy`, so the next fill that resumes it refuses:

```
production cache directory <dir> render identity differs at
'levers.nvfp4_input_global_scale_policy': cached=<missing>
current='legacy_6_over_calibration_amax.v1'
```

The shards themselves are policy-independent, but their activation-aware scores are not, so the
remedy is a fresh `--cache-dir` fill rather than a resume. The same applies to a stored CB render
identity and to a union of caches built on both sides of this change.

## Integration and final validation — 2026-09-05

Merged current main (PR #241, `72997225`) into this branch. The conflict was confined to the architecture provenance history; both contract descriptions are retained. A bounded independent review checked dense/streamed/packed policy propagation, served KL preflight and joint AURA compatibility with no concrete blocker found.

Reproduced the existing CI failure and corrected its stale mock expectation: the wrapper now explicitly passes the optional `None` policy to the same policy owner. An additional local test-fixture correction supplies the `python` command that GitHub's setup-python provides, so the pin-step regression also works on hosts whose interpreter is named only `python3`. No quantization formula changed in these integration corrections.

Full CPU suite at `52fa99ab`: **5,461 passed, 139 skipped, 3 expected failures, 174 subtests passed**, with four CI-fixture failures caused by that missing command. All four are repaired and passed in the final **15-test focused rerun** at `f49fd971` (zero skips/failures). The final source differs from the full-suite source only in that test fixture; compile and diff checks pass. The full run used 20 workers, native thread limits of one, and took 305.47 seconds; host samples were 99.8–100% CPU. CPU-only evidence, with the suite's recorded CUDA/optional-environment skips; no new GPU or served claim.

Receipts and original red/green logs are retained under `/home/rob/tessera-runs/pr230-2026-09-05/`: `suite-receipt.json`, `suite.log`, `suite.xml`, `mock-red.log`, `final-targeted.log`, `final-targeted.xml`, and `host-during-suite.json`. Existing cached activation-aware costs may require a fresh cache fill as described above.


RobTand:
### Targeted runs

All on **sparky** (rsync'd copy re-`git init`ed so `_production_cache_git_commit` resolves), CPU
only, against the pinned Tessera tree — `/home/rob/tessera-release/src`, whose
`tessera/serving/runtime_contract.json` digest is `719daa02…824927`, the exact value
`tessera_serving_runtime_pin.json` names for `8ed1d9a`:

```
rsync -a --delete --exclude '.git' <worktree>/ sparky:/home/rob/audit-sync/pq-227/
ssh sparky 'cd /home/rob/audit-sync/pq-227 && git init -q && git add -A && \
  git -c user.email=a@b -c user.name=c commit -qm sync && \
  CUDA_VISIBLE_DEVICES="" PYTHONPATH=/home/rob/tessera-release/src \
  /home/rob/venvs/pq-release/bin/python -m pytest -p no:cacheprovider -q -n 4 <31 files>'
```

The 31 files are everything the touched code reaches: the production/streaming/union caches and
their identities, the perturbed-X hooks and KL, the served/static activation contract, the
Tessera campaign + formats + render context, packed-expert cache and cost, the CB pair resume,
plus the two doc gates. Targeted only — the full suite is a release gate, not a per-PR one.

| run | result |
|---|---|
| clean `main` @ `01621219` (baseline) | **466 passed, 12 skipped, 0 failed** |
| this branch | **476 passed, 12 skipped, 0 failed** (+10 = the new regressions) |

Inherited failures on the targeted set at this main: **none**. In particular
`tests/test_tessera_campaign_resume.py` is **20 passed / 0 failed** — the 17 main-entry failures
that were background noise before #192 were all `ImportError: cannot import name 'cached_unit'
from 'tessera'`, and the new pin has that module.

A further 10 identity/cost files that build render levers or a CB render identity
(`test_cb_serialization_contract`, `test_cb_anchored_shard_merge`, `test_production_render_cost`,
`test_aura_checkpoint_resume_identity`, `test_shipcard`, `test_select_validated_frontier`,
`test_packed_append_streaming_and_pair_identity`, `test_stock_anchored_cost`,
`test_streamed_cost_checkpoints`, `test_cb_cache_pair_resume_identity`): **263 passed**.

### Pre-fix, on a real checkout of `main` @ `01621219` with the same pin

```
FAILED tests/test_served_activation_contract.py::test_the_cache_render_identity_binds_the_resolved_scale_policy
E   AssertionError: the production cache render identity does not bind the resolved activation-scale policy
FAILED tests/test_production_cache_render_identity.py::test_resume_refuses_a_changed_activation_scale_policy
E   Failed: DID NOT RAISE ValueError
```

The issue's own bounded CPU proof (`pq_activation_policy_identity.py`) now fails at its line 48
`assert identities[0] == identities[1]` on this branch, while re-deriving the same numbers it
recorded: `G=1 → 0.00025571882724761963`, `G=448 → 5.622899834634154e-07`, hook QDQ `0` and
`0.00104522705078125`.

### Caches on disk

Every existing production-cache directory (and stored CB render identity, and any union of
shards spanning this change) now refuses a resume at
`levers.nvfp4_input_global_scale_policy`, `cached=<missing>`. The shards are policy-independent;
their activation-aware scores are not, so the remedy is a fresh `--cache-dir` fill rather than a
resume. No GPU, no PrismaBuild, no model job, no full suite.

PR #264, opened 2026-09-06
Bound CPU Tessera integration fixtures by trellis geometry

### LINKED ISSUE #251: [P2] E4M3 CPU rendering integration tests exceed the 300-second CI limit
Two existing E4M3 CPU integration tests exceeded the 300-second pytest limit in PR #250 CI. This is a measured CI failure, not evidence of wrong numerical output.

Evidence: [CI run 34010631211, CPU job](https://github.com/RobTand/prismaquant/actions/runs/34010631211/job/101425702369), source `465fc13796abb0d5eeec1b3e93036478cd74fdaf`, Python 3.12.14 on Linux x86, CUDA hidden, four xdist workers and one native thread per worker. Command: `python -m pytest -c pytest.ini tests/ -q --timeout=300 -n 4 --dist worksteal --durations=30 -ra -p no:cacheprovider`.

- `tests/test_tessera_formats.py::test_a_rung_that_renders_can_still_be_unwritable`: 300.38 seconds, timeout during the second 64x512 E4M3 render after temporarily removing its serialisable-grid commitment. Stack ends in `tessera.encode.viterbi_window`, window_bits=14, rate=4, chunk=512, reference CPU recurrence.
- `tests/test_tessera_formats.py::test_the_render_leg_and_the_exporter_are_one_rendering[the exporter's default wire-TESSERA_E4M3_K1_R1024-None]`: 300.42 seconds, timeout in the actual exporter after the renderer returned, also inside the CPU Viterbi recurrence.
- Overall: 2 failed, 5,630 passed, 152 skipped, 3 xfailed, 182 subtests passed in 1,384.30 seconds. Import-surface check passed.

PR #250 changes neither this test file nor `prismaquant/tessera_render.py`; both match base main `cb517da154b274c61e53b151f7a08312d1c57c18`. Its separate 24 continuation/finalization regressions and the actual GPU export/census/serving observation passed. This failure is distinct from the scheduler assertions tracked in #244. It may share expensive encoder work with RobTand/tessera#364, but identical cause has not been established. No isolation rerun or performance fix is claimed.

Disposition: retain these failures as an open CI reliability/performance issue. Profile the exact pinned CPU path through PrismaBuild; preserve real renderer/exporter equality and commitment-invalidation coverage. Reduce redundant fixture work or move expensive coverage to an explicitly bounded integration job if evidence supports it. Do not merely raise the timeout, silently skip the cases, weaken equality checks or replace the relevant real encoder with a mock. Record actual dependency identity, before/after runtime and resource evidence, then verify the full CI result.

RobTand:
Two real E4M3 integration tests exceeded CI's 300-second limit while repeatedly encoding 64×512 tensors with the default L=14 CPU trellis. Their assertions need complete trellis history and legal column geometry, rather than that matrix size.

Derive the fixture rows from two histories of the active trellis, including tuple arity, and use a full 256-column rate superblock. Keep the default wire, real renderer, actual exporter/readback equality, serializable-grid removal, cache invalidation and existing timeout. The Tessera dependency remains ba582d476a3b6db9057ebd1385dc52926f171451.

Validation in progress: matched before/after PrismaBuild profiles and the complete touched test file are queued behind current fleet admission. The original CI failures are attributable to the exact same test-file blob as this branch's base. This draft will include profiler, both-host Netdata, dependency, test and resource receipts before it is ready.

Closes #251.

### DIFF
--- a/tests/test_tessera_formats.py
+++ b/tests/test_tessera_formats.py
@@ -69,7 +69,7 @@ def _render_seam_weight(name, recipe=None):
     if BodyKind(wire.body) is BodyKind.WINDOW:
         slowest_rate = min(family.column_schedule(rung, columns, recipe=wire))
         history = max(history, (wire.window_bits + slowest_rate - 1) // slowest_rate)
-    rows = 2 * history * family.arity
+    rows = 2 * history * wire.span * family.arity
     generator = torch.Generator().manual_seed(0)
     return torch.randn(rows, columns, generator=generator, dtype=torch.float32) * 0.02
 
--- a/tests/test_tessera_formats.py
+++ b/tests/test_tessera_formats.py
@@ -51,6 +51,29 @@ from tessera.manifest import BodyKind
 REFERENCE_SHAPE = (4096, 4096)
 
 
+def _render_seam_weight(name, recipe=None):
+    """A complete rate superblock and two trellis histories, on the real wire.
+
+    These integration checks compare producer seams, not large-matrix
+    throughput. Keep the default window width and real encoder, but derive a
+    bounded tensor from the column schedule and state history they exercise.
+    Two histories reach beyond the pinned-zero prefix; arity expands trellis
+    positions into whole output-row tuples.
+    """
+    from tessera.export import DEFAULT_CODE
+
+    family, rung = parse_tessera_format_name(name)
+    wire = tessera_wire_recipe(family, rung) if recipe is None else recipe
+    columns = SUPERBLOCK_WEIGHTS
+    history = DEFAULT_CODE.memory
+    if BodyKind(wire.body) is BodyKind.WINDOW:
+        slowest_rate = min(family.column_schedule(rung, columns, recipe=wire))
+        history = max(history, (wire.window_bits + slowest_rate - 1) // slowest_rate)
+    rows = 2 * history * family.arity
+    generator = torch.Generator().manual_seed(0)
+    return torch.randn(rows, columns, generator=generator, dtype=torch.float32) * 0.02
+
+
 def _forest_bpp(spec, rung, shape, wire=None):
     """What a TCQ unit's forest costs per position at ``shape``.
 
@@ -751,9 +774,8 @@ def test_a_rung_that_renders_can_still_be_unwritable():
         tessera_rung_is_serialisable,
     )
 
-    torch.manual_seed(0)
-    weight = torch.randn(64, 512) * 0.02
-    assert render_tessera_weight(weight, "TESSERA_E4M3_K1_R1024").shape == (64, 512)
+    weight = _render_seam_weight("TESSERA_E4M3_K1_R1024")
+    assert render_tessera_weight(weight, "TESSERA_E4M3_K1_R1024").shape == weight.shape
     assert tessera_rung_is_serialisable("TESSERA_E4M3_K1_R1024") is True
     assert tessera_rung_is_serialisable("TESSERA_E2M1_K2_R896") is True
 
@@ -769,7 +791,7 @@ def test_a_rung_that_renders_can_still_be_unwritable():
         clear_serialisable_cache()
         # Still renders -- if it stopped, this would pass for the wrong reason.
         assert render_tessera_weight(
-            weight, "TESSERA_E4M3_K1_R1024").shape == (64, 512)
+            weight, "TESSERA_E4M3_K1_R1024").shape == weight.shape
         assert tessera_rung_is_serialisable("TESSERA_E4M3_K1_R1024") is False
     # Both levels again on the way out: the verdict cached under the patched
     # registry is the one the next test in this file would otherwise read.
@@ -1495,8 +1517,7 @@ def test_the_render_leg_and_the_exporter_are_one_rendering(label, name, recipe_a
     family, rung = parse_tessera_format_name(name)
     recipe = None if recipe_args is None else recipe_from_wire_names(*recipe_args)
 
-    torch.manual_seed(0)
-    weight = torch.randn(64, 512, dtype=torch.float32) * 0.02
+    weight = _render_seam_weight(name, recipe)
 
     rendered = render_tessera_weight(weight, name, recipe=recipe)
 


RobTand:
## Validation evidence (the measurements this PR body says it owes)

Independent validation on the fleet. Everything below was executed through PrismaBuild on `sparky`
(linux-aarch64-sm121); nothing was run locally. **One item needs work before merge**; the rest of the
PR checks out.

### Finding: the helper ignores `wire.span`, so one case loses the coverage the docstring claims

`_render_seam_weight` sizes the fixture as `rows = 2 * history * family.arity`, where `history` is
`DEFAULT_CODE.memory` (6). That counts trellis **positions**. But the convolutional code advances once
per **super-symbol**, not per position:

```python
# tessera/encode.py:379-381
self.steps  = steps  = rows // self.arity
self.supers = supers = steps // span
# tessera/encode.py:455-456  (pinned-zero prefix)
cost.fill_(float("inf")); cost[:, 0] = 0.0
```

So the shift register is fully data-dependent only after `memory` **super-symbols**, and the fixture
must carry `2 * memory * span * arity` rows, not `2 * memory * arity`.

`TCQ_RECIPE` has `span = DEFAULT_SPAN = 2` (`export.py:135`, `:1043`), and `TESSERA_E2M1_K2_R896`
resolves to it because `q256 = 896 >= tcq_cap_q256 = 7 * 256 // 2 = 896` (`export.py:1200-1215`).
Measured geometry for the six parametrized cases plus the serialisable-seam test, as merged
(action `c96b376332a95b63…`):

| case | body | span | arity | rows | steps | supers | supers/memory |
|---|---|---|---|---|---|---|---|
| default wire, `TESSERA_E2M1_K2_R896` | TCQ | 2 | 2 | 24 | 12 | **6** | **1.00** |
| default wire, `TESSERA_E4M3_K1_R1024` | WINDOW | 1 | 1 | 12 | 12 | 12 | 2.00 |
| below the E2M1x2 cap, `..._R768` | WINDOW | 1 | 2 | 24 | 12 | 12 | 2.00 |
| span 1 / s6b, `TESSERA_E2M1_K1_R768` | TCQ | 1 | 1 | 12 | 12 | 12 | 2.00 |
| window L=8 / channel | WINDOW | 1 | 1 | 12 | 12 | 12 | 2.00 |
| window L=8 / lut16 | WINDOW | 1 | 2 | 24 | 12 | 12 | 2.00 |
| serialisable-seam test | WINDOW | 1 | 1 | 12 | 12 | 12 | 2.00 |

The first row is the exporter's default TCQ wire, and it lands at `supers == memory == 6`: **zero**
super-symbols are encoded past the pinned-zero prefix, so that case no longer exercises a
data-dependent shift-register state at all. The docstring's "Two histories reach beyond the
pinned-zero prefix" is false there and true everywhere else. Termination is free at both ends
(the front is an argmin over the final column, no tail flush), so `memory` super-symbols is the
entire prefix cost — nothing else absorbs the shortfall.

Worth noting: neither of #251's two timing-out tests was this case (both were E4M3/window, span 1).
Shrinking the TCQ case bought nothing measurable — it is 2.78 s on `origin/main` and 3.07 s on this
branch — and it is the one case that lost coverage.

### Suggested fix (one factor), measured

```diff
-    rows = 2 * history * family.arity
+    rows = 2 * history * wire.span * family.arity
```

With that change every row above reads `supers/memory = 2.00`; only the first case changes
(rows 24 -> 48, supers 6 -> 12), the other six are byte-identical (span is 1 for all of them).
The suite still passes 66/66 and the whole file gets *faster*, not slower, than this branch as it
stands: 35.08 s vs 38.72 s, and the corrected case itself is 2.47 s.

### What holds

Checked against the tests' actual assertions, not the PR description:

- **Full rate superblock.** `columns = SUPERBLOCK_WEIGHTS = 256`, so `bresenham_rate_schedule` is
  exercised over one entire canonical superblock. Measured: every schedule is length 256 and
  single-rate.
- **Scale groups.** `256 % DEFAULT_GROUP(32) == 0`, so no group is truncated and none spans rows.
- **Tuple arity.** `rows % arity == 0` in all seven cases; `steps = rows // arity` is exact.
- **Window bodies.** The `(window_bits + slowest_rate - 1) // slowest_rate` derivation uses
  `min(schedule)`, i.e. the slowest column, which is the correct and conservative prefix depth for
  the window shift register; `span` is refused unless 1 for a window body (`encode.py:2355`), so the
  span factor is a no-op on those five cases either way.
- **Equality assertion is unchanged.** `assert torch.equal(rendered, off_the_wire.to(weight.dtype))`
  still compares the full rendered tensor; the shape assertions were correctly rewritten from a
  hardcoded `(64, 512)` to `weight.shape`.

### Timing and failure sets

Base arm and PR arm ran concurrently on `sparky`, started 21 s apart, on disjoint 4-core
reservations, so they saw identical contention.

| arm | commit | pytest wall | action wall | result | failure set |
|---|---|---|---|---|---|
| base (`origin/main`) | `1f0df20d` | **51.00 s** | 60.37 s | 66 passed, 0 failed, 0 skipped | {} |
| PR | `ca2b8595` + `1f0df20d` merged (`c22fc6c8`) | **38.72 s** | 50.72 s | 66 passed, 0 failed, 0 skipped | {} |
| PR + span fix | `c22fc6c8` + the one-line diff | 35.08 s | 72.51 s | 66 passed, 0 failed, 0 skipped | {} |

The two tests #251 named:

| test | base | PR | speedup |
|---|---|---|---|
| `test_the_render_leg_and_the_exporter_are_one_rendering[window L=8 / channel (minor 3)-...]` | 7.92 s | 0.84 s | 9.4x |
| `test_a_rung_that_renders_can_still_be_unwritable` | 7.65 s | 0.78 s | 9.8x |

Failure **sets** are identical (both empty), not merely equal counts. All arms report `executed`,
none `cache_hit`, with distinct snapshot commits, so no CAS dedup masked a difference.

Host load (Netdata, node `sparky`, 20 cores): the paired base/PR arms ran with `load1` 7.5 -> 17.2 and
CPU user peaking at 25.9 %. The fix arm ran later with `load1` 17.7 -> 10.2; it is therefore not a
matched-pair delta, but since it is already *faster* than the PR arm under heavier load, the fix's
cost is below the noise either way.

Caveat worth stating plainly: `sparky` does not reproduce the CI failure. The slow tests are 7.9 s
here versus 300.4 s on the x86 CI runner (which runs one native thread per xdist worker; these arms
used 4). "Comfortably under the 300 s limit" therefore rests on this PR's own green
`tests (py3.12, cpu)` check, with the measured 9.4x/9.8x per-test ratio as corroboration. `pytest.ini`
carries no timeout, so the base arm's 51.00 s is a true untruncated before-time on this hardware.

Also: the branch is `BEHIND` `main` and needs an update before merge (trivial, no conflict).

### Receipts

| what | action key | receipt sha256 |
|---|---|---|
| base suite | `1678597cb0c205ea43918a69d4e9777fe4e0c63231c53fe0b996afe0a32e0d44` | `3bfc63e5dde90180…` |
| PR suite | `6426d7a2f4de75a631ad36c979529904219a8f0074a023daa2ea7774fbb952a2` | `7be3fcc291929328…` |
| PR geometry report | `c96b376332a95b6338b7d369e63fad3fab64d80fc99964d827eaf0c78fbb7363` | `182e762e0b90a666…` |
| fix suite | `f8996e8dad0cfec8e8fcbbf3d7bf6b064658b4d884d15deeb56d1175569e9984` | `694c8f9cf42879e3…` |
| fix geometry report | `7cbcefc44b681dc1a71dd50a135fd4102f619da2954728d3255467054c96b327` | `0cf4c888d4794ead…` |

CAS root `/mnt/shared/prismabuild-fleet/cas`.

RobTand:
Fix applied as `20a80296` and merging. Recording what was preserved and what was verified, since this branch had uncommitted work in the author's worktree.

## The tree merged is the tree that was measured

`20a80296` is `ca2b8595` merged with `1f0df20d` plus the one-line span fix. `origin/main` has not moved since the measurement (`1f0df20d` is still the tip), so I reproduced the fix arm exactly rather than re-running it:

```
changed-file blob   881100ddf304a99314cf0dff424d6fbd6080ab53   identical
full tree           e14012a3cf0fe8fa7da42253333993c5409b8ff9   identical
```

Both hashes match the arm that produced receipt `f8996e8dad0cfec8...` (66 passed / 0 failed / 0 skipped, 35.08 s). Re-validating a byte-identical tree would return a `cache_hit` and prove nothing new, so I did not submit one.

## Codex's uncommitted work is preserved, not discarded

The author's worktree `/home/rob/tmp/pq-cpu-tessera-integration` held three uncommitted items, last modified 02:13–02:42 with no live process on it. They are shelved verbatim at:

```
refs/shelf/codex-cpu-tessera-integration-2026-09-06   ->   50fe79dd
```

pushed to origin. Recover with `git fetch origin 'refs/shelf/*:refs/shelf/*'`. Contents:

- **`tests/test_tessera_formats.py`** — a docstring softened from "two trellis histories" to "at least two histories". **Superseded on purpose.** That wording change accommodated the defect; the span fix removes it. With `supers/memory = 2.00` for all seven cases, the *original* docstring is true again, so I kept the original rather than the softened one.
- **`docs/measurements/tessera-cpu-integration-2026-09-06.md`** — a draft that says of itself *"Results and profiler/Netdata receipts will be recorded here before this draft is finalized."* It is not finalized, so it is not merged. Worth salvaging from it: the CI run link (34010631211), the test-file blob identity `64f06019f9…`, and the operational note that PB refuses external symlink targets so snapshots must materialize tracked calibration links.
- **`docs/measurements/artifacts/tessera-cpu-integration-251/`** — `ci.json`, `materialized-calibration.json`, `missing-psutil.json` (40K).

I did not touch the author's index or working tree; the shelf commit was built through a temporary index, and `git status` there is unchanged. That worktree is now behind `origin`, but nothing in it is lost.

## Why the fix, restated for the record

`_render_seam_weight` sized rows in trellis **positions** while the conv code advances once per **super-symbol** (`encode.py:379-381`). For `TESSERA_E2M1_K2_R896`, which resolves to `TCQ_RECIPE` at `span = 2`, that divided the fixture twice and landed at `supers == memory == 6` — zero super-symbols past the pinned-zero prefix. The case was no longer exercising the trellis. Neither of #251's two timeouts was this case (both E4M3/window, span 1), so shrinking it saved no time and cost the only coverage that was lost.

Closes #264.

RobTand:
## `main` is currently red on exactly the two tests this PR fixes

Run [34043991599](https://github.com/RobTand/prismaquant/actions/runs/34043991599) on `main` (`1f0df20d`):

```
FAILED tests/test_tessera_formats.py::test_a_rung_that_renders_can_still_be_unwritable
        - Failed: Timeout (>300.0s) from pytest-timeout.
FAILED tests/test_tessera_formats.py::test_the_render_leg_and_the_exporter_are_one_rendering
        [the exporter's default wire-TESSERA_E4M3_K1_R1024-None]
        - Failed: Timeout (>300.0s) from pytest-timeout.
2 failed, 5705 passed, 152 skipped, 3 xfailed, 184 subtests passed in 1415.98s
```

Those are #251's two named tests, and they are the two this branch measured at **9.4×** and **9.8×** faster (7.92 → 0.84 s and 7.65 → 0.78 s on sparky). So this is not only a CI-hygiene change; it is the fix for a branch that is red right now.

**It is not caused by the commit it ran on.** `1f0df20d` (#269) deletes three symlink files under `calibration/` and touches nothing else:

```
calibration/diverse-v1.jsonl   | 1 -
calibration/xdom-fit-v1.jsonl  | 1 -
calibration/xdom-gate-v1.jsonl | 1 -
3 files changed, 3 deletions(-)
```

Nothing there reaches the trellis encoder. These are timeouts, not import or fixture errors.

**Why they tipped over now.** The x86 runner measured one of these at **300.4 s against a 300 s limit** — a four-tenths-of-a-second margin. A test that close to its own timeout is not passing, it is winning a coin flip, and which side it lands on depends on runner load rather than on the commit. The run at 06:04 was green; this one was not; the code between them is three deleted symlinks. That is the signature of a boundary, not a regression.

Which is the real argument for this PR: shrinking the fixture does not merely make CI faster, it moves these tests off a boundary they were never meant to be sitting on. A 38× margin is a test; a 1.001× margin is a scheduler.

PR #266, opened 2026-09-06
Preserve static activation scales in transient Tessera anchors

### LINKED ISSUE #262: [P2] Populate static Tessera activation scales in the transient anchor renderer
The real transient production-anchor renderer does not populate calibrated activation maxima for a Tessera-only A4 plan. Joint AURA therefore refuses the static activation identity before it can measure that plan.

Inspected at PrismaQuant commit `4bde8bd65eeb133eadae75fd71697dfadad6749d`:

- `prismaquant/streaming_production_cache.py:331-377` derives `needs_nvfp4` only from the literal `NVFP4` render base and places the only assignment to `cache.activation_max_abs` behind that condition.
- `prismaquant/production_weight_cache.py:176-190` leaves Tessera render-base names unchanged.
- `prismaquant/streaming_production_cache.py:768-774` initializes the renderer-owned `ProductionWeightCache` with empty activation maxima.
- `prismaquant/tessera_render.py:947-950` stamps the resolved static Tessera activation contract `measured_as_served=True`.
- `prismaquant/joint_aura.py:45-48` consequently requires a calibrated input global scale. A plan containing `TESSERA_E2M1_K1_R768` without native `NVFP4` has no such maximum supplied by this renderer.

Severity: **P2**. This is a research-path input-contract/coverage gap with fail-closed behavior; no incorrect artifact or measured numerical regression is claimed. The evidence is source inspection only; no GPU reproduction has been run.

The current bounded qualification in #261 can use the existing materialized `ProductionWeightCache` path, with actual rendered tensors and `activation_max_abs` captured from the same BF16 calibration inputs. The earlier numerical screen supplies this field in `experiments/pq237_joint_aura_screen.py:135-175`. This avoids the unsupported transient path without introducing another cache.

Fix the shared streamed render mechanism to derive calibrated maxima from each format's resolved static activation contract, preserving fused-group semantics and resume identity. Add a regression using a Tessera-only A4 anchor plan and validate through PrismaBuild.

Filed rather than fixed because the active delegated task is the immutable qualification protocol, and the root explicitly requested this separate issue without broadening that implementation. The change is feasible in a separately scoped follow-up; it is not blocked on a user decision.

RobTand:
Tessera-only static A4 transient anchors previously reached scoring without a calibrated activation maximum and were refused. Reuse the shared resolved-format predicate and gather maxima across the complete planned fused-sibling group, so batched and one-sibling requests produce the same static scale identity. Native NVFP4 weight globals retain their separate guard.

PrismaBuild reproduced both regression cases failing before the fix. Corrected CPU validation passed 19 tests with two explicit CUDA skips, covering real static activation scoring/identity and existing streamed checkpoint behavior. The CPU arm substitutes only weight rendering; its CAS payload was independently verified. Exact actions, environment limitations and snapshot provenance are recorded in `docs/measurements/static-anchor-scales-2026-09-06.md`.

Keep draft: real Tessera GPU regression and existing streaming/resident parity runs are pending through PrismaBuild in the known producer container. No served quality, performance or promotion claim is made. The Tessera pin is unchanged.

Closes #262

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,16 @@
 As of: 2026-09-05 · `codex/pq183-tessera-reader`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-06, `codex/pq262-static-anchor`) for streamed static
+activation calibration (#262). `_render_dense_layer` uses the resident cache's
+resolved-format predicate `_formats_need_static_activation_max`, including
+Tessera-only A4 plans. Native NVFP4 weight-global derivation remains separate.
+Sparse transient requests calibrate the complete planned layer scope and unify
+fused-sibling maxima before scoring any member, so render batching/order cannot
+change the activation identity consumed by joint AURA. Existing calibration,
+source, render-lever and transient-consumer bindings remain required; no new
+cache, serving route, numerical promotion or production default is introduced.
+
 Re-stamped (2026-09-06, `codex/lfm-mixed-preparation`) for the opt-in
 LFM full-body mixed-family sanity input preparation (#253). The producer's
 source classification and fusion roster define 22 E4M3 routed stacks, six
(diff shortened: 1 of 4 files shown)


RobTand:
Merging (Rob: merge aggressively as issues land). After update-branch onto main (which carries #264's span fix) CI is green: the two earlier failures were the pytest-timeout pair in test_tessera_formats that #264 addressed. PrismaBuild CPU-arm receipts are in docs/measurements/static-anchor-scales-2026-09-06.md; the GPU regression run stays owed on #262's successor if anything surfaces.

Issue #233, opened 2026-09-05
[P1] Validate the producer’s v9 smoke-record grammar before deriving admission


RobTand:
## Trigger and current defect

PR #192 adds the v9 smoke-record reader, but `prismaquant/lane_eligibility.py::_parse_smoke_record` implements a different grammar from the pinned producer. It delegates status and attribution derivation to Tessera **after accepting records Tessera itself rejects**, and it rejects one explicitly supported producer record.

At PrismaQuant `01621219410fadfad0f625dad341a25c6024b1cb`:

* Lines 903–911 require `instrument`, `rule`, and `reference` all to be nonempty strings. Tessera v9 explicitly permits `reference: null` for a smoke without a reference arm, and requires a portable repository path for `instrument`.
* Lines 941–949 require both status fields to be strings from the status vocabulary. A valid unreferenced record instead requires each `reference_status: null`.
* Lines 950–953 detect duplicates using **all five row values**. The producer defines observation identity as `(prompt, form)`, so contradictory observations for the same pair must be rejected even when their status/interface fields differ.
* A nonempty `reference` is never checked against the producer's reference vocabulary; a name other than `bf16_source` is accepted and used in the admission derivation.

The corresponding owning grammar is Tessera `src/tessera/serving/contract.py:1498–1557` (`_evidence_smoke_record`). That source and the packaged contract are unchanged between PrismaQuant's pinned Tessera `8ed1d9a78b3f0c7036dcbe14d7df3a89f398812a` and the audit's Tessera `8dc165b` snapshot.

## Reproduced evidence

A bounded CPU differential proof calls the **full real Tessera contract validator**, the real PrismaQuant table parser, and `resolve_unit_route`. It starts from the committed v22/v9 contract and changes only the smoke record of `tessera_e4m3_k1_routed_moe_sm121_decode_resident`:

| Record | Tessera validator | PrismaQuant reader / route |
|---|---|---|
| Unchanged packaged control | accepts | accepts, backed_with_serve_flag |
| Positive smoke with `reference: null`, `reference_status: null` | accepts | refuses `reference must be a non-empty string` |
| Unpublished reference name | refuses | accepts, backed_with_serve_flag |
| One shared cycling observation | accepts | accepts, unattested |
| The same `(prompt, form)` twice, first cycling and second recorded on both arms | refuses duplicate observation | accepts, changes status to recorded and route to backed_with_serve_flag |
[...]

PR #270, opened 2026-09-06
Fix the refill-order race in the cluster campaign CI, and say what the recovery failure is not (#244)

### LINKED ISSUE #244: [P2] Cluster campaign CI still has a refill-order race and an unexplained recovery failure
Two independent failures in `tests/test_cluster_campaign_runner.py` remain in the broader GitHub CPU suite. Neither test nor `prismaquant/cluster_campaign.py` was changed by the PR whose CI observed it. This is an observed CI reliability/recovery investigation, not a claim that quantized numerics regressed.

1. **The refill regression asserts an ordering its protocol does not establish.** [PR230 job](https://github.com/RobTand/prismaquant/actions/runs/34003557254/job/101406592172), source `f49fd971664120132d583a463264122303ca2b04`, failed `test_scheduler_refills_a_freed_host_while_another_host_is_active` at line522: waiting completion timestamp1788657740585217091 was earlier than refill completion1788657740606789620. The refill child publishes `gate.receipt` before its own exit and the coordinator records completion later in `_finish_attempt`; the waiting child may observe the gate and have its own completion recorded first. A correct refill scheduler can therefore fail this assertion. The test should establish a deliberate publication/acknowledgement ordering and prove that the freed host is used while the other stage is still active, without asserting unrelated completion-observation order.

2. **Recovery of a completed orphaned helper still intermittently terminal-fails.** [PR242 job](https://github.com/RobTand/prismaquant/actions/runs/34004489141/job/101409116620), source `a26cfcce06cb66472dd17fc7464c19a95dfcc43d`, failed `test_resume_after_coordinator_death_recovers_owned_helper_receipt`: after release and the nonempty receipt wait, `run_campaign_v2` raised `CampaignTerminalFailure: campaign has terminal failed stages: ['slow-stage']` at cluster_campaign.py1857. The failure output does not include the resulting attempt detail/worker log, so it does not establish whether ownership observation, receipt verification, or another transition caused it. Capture those records on failure, reproduce under bounded concurrent CPU work, and fix the demonstrated cause while preserving fail-closed PID ownership and exact receipts. Do not weaken those guards based on the traceback alone.

Actual CPU results: PR230 **1 failed,5563 passed,149 skipped,3 xfailed,174 subtests passed**,954.17s; PR242 **1 failed,5565 passed,149 skipped,3 xfailed,174 subtests passed**,946.93s. Import-surface jobs passed. CUDA skips are not GPU evidence. Local retained logs: `/home/rob/tessera-runs/measurement-208-183-2026-09-05/pr230-ci-failure.log` and `pr242-ci-failure.log`.

Acceptance: make both tests verify their intended causal contracts; demonstrate any production regression before its fix; retain failure-state diagnostics; run the cluster runner test file through PrismaBuild with useful parallelism and bounded native threads; verify exact owned-child cleanup on passing and failing paths. Do not increase sleeps or introduce unbounded waits to hide the failures.

RobTand:
Closes #244.

Both halves are reproduced on PrismaBuild before being fixed, and the recovery
half is now attributed rather than guessed at — the reproduction carries the
record the issue asked for.

## 1. The refill race

`test_ready_stages_refills_a_freed_host_slot` asserted
`waiting.finished_unix_ns >= refill.finished_unix_ns`. Nothing in the manifest
establishes that order. The refill child publishes its gate receipt *while it is
still running*, so the stage it unblocks can finish, be recorded and be written
out before the refill stage's own attempt is recorded. The assertion was testing
a coincidence, and CI eventually stopped agreeing with it.

Reproduced first, with the old assertion against the current runner:

```
assert 1788713089922203363 >= 1788713090049877524
1 failed, 14 warnings in 6.79s
```

action key `6879e0df33e4e05f72e665e3a69e2e2f08484d9e0a5728680dc59ff9f522a67e`.

The fix asserts what the gate actually establishes — that the two stages
*overlap* — and adds a `release-refill` stage whose completion order is fixed by
causality rather than by scheduling. The order the old assertion demanded is now
inverted by construction, so it cannot come back unnoticed.

### It reproduces, and it says why

40 runs of `test_resume_after_coordinator_death_recovers_owned_helper_receipt`
against **unmodified `origin/main` code**, with only the diagnostic reporting
this PR adds, box load 24–30: **1 failure**
(`5e41dbcd0c7db599107aa85f7304d35bd6373a644fe8dfda7d2cc7ce9a7bdfd3`). The
failure carries the detail no previous occurrence recorded:

```
slow-stage: status=terminal_failed
  attempt 1: status=failed pid=667567 start_ticks=1113697
    detail='recorded PID ownership is ambiguous; process was not killed'
```

A second 40-run arm on the same production code with the unmodified test file
saw 0 failures
(`333e2f71e4c4d26572e1827d2ddf6522f254f85e187636dc063e290b52dbac99`), so the
observed rate is 1 in 80 runs, which is what "intermittent, twice, never
reproduced" looks like.

### What the ambiguity actually is

Linux `do_exit` runs `exit_mm()` before `exit_notify()`. Between those two
points a helper has released its address space but is not yet a zombie:
`/proc/<pid>/stat` still reports the recorded start time and a live state, and
`/proc/<pid>/environ` stops being readable. `origin/main` read "I could not ask
who owns this pid" as "somebody else owns this pid", and one such verdict
terminal-fails the stage.

Measured against unmodified `origin/main`, real processes, nothing patched
(`22f596af449888db925f274111d3c66f06e1e598f41e0b769bd21663770fbef2`): 199 of 200
exiting children pass through that window, and **all 2492 samples taken inside
it read `EACCES`/`EPERM`** — 0 empty, 0 truncated — with the recorded start time
intact in every one. States seen inside the window: `R` 2455, `Z` 29, `D` 8.

### Why it is intermittent

The window is `exit_mm()`, so its width is a function of how much address space
the kernel has to tear down. Polling at the recovery loop's real
`poll_interval=0.02`, 60 trials per cell, both sides pre-fix
(`da8fac6d04c6ef95b5642b14be38030799b1f60c0421ec4d762ec4a5679d1d5d`):

| helper footprint | polls a trial takes (max) | trials a poll called ambiguous |
|---|---|---|
| 0 MB | 2 | 0 / 60 |
| 256 MB | 3 | 0 / 60 |
| 1024 MB | 6 | **4 / 60** |
| 4096 MB | 19 | 1 / 60 |

So the mechanism is not exotic — a helper holding a gigabyte makes it a
several-percent event — and at the CI helper's size it is rare, which matches a
failure seen twice and reproduced once in eighty runs.

### The first fix was wrong, and the measurement is how I know

The first version of this branch carved out only the *empty* read: an environ
that returns zero bytes from a process whose start time is intact is `owned`.
The same probe, asking both implementations about the **same child at the same
sample**, showed that candidate returning `mismatch` inside the window 1234
times and `owned` **zero** times
(`d5eb835dcd487ec2d2573c0a74db7a4664fb501176306b816faa2587d2201163`). It would
have changed nothing, because the kernel refuses the read rather than emptying
it. That commit is still in the history with its reasoning; this one corrects
it.

### The rule that replaced it

State the rule on the fact that decides it. Field 19 of `/proc/<pid>/stat`
establishes identity. The owner token is a second check that can only ever fail
to be *readable*.

- A read that **failed** — refused, empty, or absent — establishes nothing about
  who owns the pid. With the recorded start time intact this is the recorded
  process, so the verdict stays `owned` and the caller's **existing** bounded
  wait looks again. No new constant, no new timeout.
- A read that **succeeded** and lacks the token is the one observation that
  establishes a different owner. It still fails closed.
- A changed start time is still a recycled pid and still fails closed.
- Neither refusal authorizes killing that process, exactly as before.

What this gives up is bounded and points the safe way: a pid recycled inside one
clock tick, by a process whose environment cannot be read at all, would be
terminated at the recorded deadline rather than refused outright.

Two smaller fixes ride along:

- `_terminate_recorded_owned_process` returns True when the process has already
  ended. There is nothing left to kill and the receipt is recoverable. This is
  **not** the CI cause — reaching it needs the 125 s kill deadline, and both
  retained CI logs finish the whole test in under 28 s. It is fixed because it
  is wrong.
- `CampaignTerminalFailure` now names each terminal stage's attempt, its refusal
  detail and its worker log, and the recovery test renders the whole campaign
  state on failure. That is #244's "capture those records on failure", and it is
  what produced the attribution above.

### The fix closes it

The same reachability probe, rerun with the corrected rule as the candidate
(`366134fb24f870c0b650acbf5dcb91fa055cd6cdbc39056348616cbbb00d275b`), 60 trials
per cell at the real poll interval:

| helper footprint | `origin/main` | this branch |
|---|---|---|
| 0 MB | 0 / 60 | 0 / 60 |
| 256 MB | 3 / 60 | 0 / 60 |
| 1024 MB | 0 / 60 | 0 / 60 |
| 4096 MB | 2 / 60 | 0 / 60 |
| **total** | **5 / 240** | **0 / 240** |

The earlier run of the same probe put `origin/main` at 5/240 and the first,
narrow fix at 5/240 as well. The corrected rule is the one that closes it.

### What is still not established

That this mechanism caused the **two original CI occurrences**. Neither recorded
its detail, so that record does not exist and cannot be recovered. What is
established is that the same test, on the same code, fails this way, at a rate
consistent with how those two occurrences appeared. I am not claiming more than
that.

### The test that would have caught it

`test_an_exiting_helper_is_never_called_ambiguous` watches twenty real helpers
exit and asserts the verdict at every sample. It is deterministic rather than
lucky: it samples a window every exiting child opens, and our own child can
acquire neither a different owner token nor a different start time, so it can
only fail when the contract is actually violated — it cannot flake green-to-red.

## Prior art

A local, unpushed branch `codex/cluster-campaign-causality` (Astra) carries
`7c5c300d "test: assert causal refill overlap rather than finish order (#244)"`,
which reaches the same overlap assertion for the refill half — independently and
first. It stops there: it removes the unfounded assertion without a manifest
that forces the inverted completion order, so nothing demonstrates the failure
it removes, and it does not touch the recovery half. It also edits
`docs/ARCHITECTURE.md` for fixture-cleanup guarantees, unrelated to this stamp.
Last commit 06:32Z, never pushed; the issue was released to triage.

## Verification

All runs through PrismaBuild on `sparky`, CPU-only, serial pytest, one native
thread per worker. The test venv has no `pytest-xdist`, `pytest-randomly` or
`pytest-repeat`, so `-n` and `-p no:randomly` are unavailable; serial is both
forced and what a failure-**set** comparison needs.

### What has been run

| run | tree | scope | result | action key |
|---|---|---|---|---|
| refill race, red | branch + the old assertion | the refill test | 1 failed, 6.79s | `6879e0df33e4e05f72e665e3a69e2e2f08484d9e0a5728680dc59ff9f522a67e` |
| baseline suite | `origin/main` `1f0df20d` | the 3 files that import `cluster_campaign` | **64 passed**, 35.22s | `b78b1461e31f4777b2aa0432f7bc351906eb8a62bdfc6de598144aaeef02a103` |
| branch suite @ `ae8f156f` | branch | same 3 files | **74 passed**, 34.47s | `538a62eabbee2e7b662041dfa4fdab47622ab4d3c0adaa21e47810a0bd52093c` |
| red suite @ `ae8f156f` | `origin/main` code + branch tests | same 3 files | 17 failed, 57 passed | `38fb91d6cf6b34ed3c31ca4771ce90db96b16c4145de98b8a33c91a059658b87` |
| docs @ `2f1238ec` | branch | `test_architecture_doc` + `test_docs_staleness` | **19 passed**, 6.17s | `67f2354cd15c5495530bfe4daac4ef33f14eb5540c9cff3b7541b9136866e124` |

Failure **sets**: baseline `{}` and branch `{}`. No skips were reported in
either. The branch adds 10 tests. `origin/main` has since moved by one commit
(#264, `test_tessera_formats.py`), outside this scope.

The 17 red failures split into 13 expected symbol errors — `AttributeError` on
`_proc_owner_observation`, which does not exist on `origin/main`, so those tests
cannot run there and are not contract failures — and 4 genuine contract reds:

- `test_owned_process_state_rechecks_exit_after_failed_owner_read[empty-env-still-live-same-identity]`
- `test_terminating_a_process_that_already_ended_is_not_ambiguous`
- `test_ambiguous_recorded_pid_fails_closed_without_killing_it[on-resume]` and
  `[during-monitoring]`, both on the tightened refusal message.

Probes, all against real processes with nothing patched:

| probe | what it establishes | action key |
|---|---|---|
| window | the old rule's mismatch span is 0.24–0.76 ms; the *first* fix returned `mismatch` inside it 1234 times and `owned` 0 | `d5eb835dcd487ec2d2573c0a74db7a4664fb501176306b816faa2587d2201163` |
| cause | 2492/2492 samples inside that window read `EACCES`/`EPERM`, identity intact | `22f596af449888db925f274111d3c66f06e1e598f41e0b769bd21663770fbef2` |
| reachability, pre-fix | `origin/main` 5/240, first fix 5/240 at the real poll rate | `da8fac6d04c6ef95b5642b14be38030799b1f60c0421ec4d762ec4a5679d1d5d` |
| reachability, corrected | `origin/main` 5/240, **this branch 0/240** | `366134fb24f870c0b650acbf5dcb91fa055cd6cdbc39056348616cbbb00d275b` |
| reproduction | 1 failure in 40 on `origin/main`, carrying the ownership detail | `5e41dbcd0c7db599107aa85f7304d35bd6373a644fe8dfda7d2cc7ce9a7bdfd3` |
| reproduction, control | 0 failures in 40 on the same production code | `333e2f71e4c4d26572e1827d2ddf6522f254f85e187636dc063e290b52dbac99` |

### What is still owed

The PrismaBuild fleet wedged before these landed — sparky's NFS client entered a
`TEST_STATEID` storm that blocked every worker's queue access (prismabuild
#230–#233), and a reboot was authorised. They were queued, not run, and none
should be read as pending-and-probably-fine:

- **No suite run covers `2f1238ec..dfa673f8`.** The 74-passed run above is
  commit `ae8f156f`, which is the *first*, narrow fix. The corrected rule's only
  evidence today is the 0/240 reachability A/B, which exercises the changed
  function on 240 real processes but is not the test suite.
  Owed: `599331e419a7` (branch suite at head), `033838b41c03` (red suite at
  head), `7277b162a1e3` (docs at head).
- **The repeat A/B is owed.** 120 runs of the recovery test on `origin/main`
  (`b234310489a5`) and on this branch (`771d373e334b`), to put a rate on the
  reproduction rather than a single occurrence in 80. Expect roughly 1–3
  failures on the baseline arm and 0 on the branch arm; a baseline arm that
  comes back clean weakens the attribution and should be said so.

Please re-run those five after the reboot before this leaves draft.

https://claude.ai/code/session_01FR3Gv58q8SoKnEraH97uu3

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -6,20 +6,25 @@ follow, newest first, each recording its own branch and date.
 Re-stamped (2026-09-06, `claude/pq244-campaign-causality`) for **an exiting
 helper being the same helper** (§3.0; RobTand/prismaquant#244, P2). The #239
 contract above reads "a failed owner-environment read now rechecks process
-identity"; it treated every failed read alike. Linux does not: `do_exit` runs
-`exit_mm()` before `exit_notify()`, so a helper that is on its way out drops
-its address space -- `/proc/<pid>/environ` reads zero bytes -- while
-`/proc/<pid>/stat` still reports the recorded start time and a state that is
-not `Z`. The old code called that identity ambiguous and the campaign failed
-closed on a process it owned. Measured on this kernel (6.17.0-1032-nvidia),
-679 of 680 exiting children were judged ambiguous at least once during that
-window. `_proc_has_owner` is therefore split into `_proc_owner_observation`,
-which distinguishes an empty read from an absent, unreadable or
-differently-owned one, and exactly one cell of the `_owned_process_state`
-table changes: an empty read from a live process whose start time still
-matches is `owned`, not ambiguous. Absence, an unreadable or malformed read, a
-different owner token and a changed start time all remain ambiguous and still
-fail closed, and none of them authorizes killing that process. Two smaller
+identity"; it rechecked identity and then called a live one ambiguous anyway.
+Linux `do_exit` runs `exit_mm()` before `exit_notify()`, so a helper on its way
+out has released its address space while `/proc/<pid>/stat` still reports the
+recorded start time and a state that is not `Z`, and `/proc/<pid>/environ`
+stops being readable. Measured on this kernel (6.17.0-1032-nvidia): 199 of 200
+exiting children pass through that window, and all 2492 samples taken inside it
+read `EACCES`/`EPERM` -- never empty, never truncated -- with the recorded
+start time intact in every one. The old code read "I could not ask who owns
+this" as "somebody else owns this" and the campaign failed closed on a process
+it owned. `_proc_has_owner` is therefore split into `_proc_owner_observation`,
+which reports what the read established rather than a bare boolean, and the
+`_owned_process_state` rule is stated on the fact that decides it: field 19 of
+`/proc/<pid>/stat` establishes identity, and the owner token is a second check
+that can only ever fail to be *readable*. So a read that **failed** -- refused,
+empty or absent -- leaves a process whose recorded start time is intact
+`owned`, and the caller's existing bounded wait looks again. A **successful**
+read whose environment lacks the token is the one observation that establishes
+a different owner, and it still fails closed; so does a changed start time, and
+neither authorizes killing that process. Two smaller
 consequences: terminating a recorded owned process that has already ended
 reports success rather than failure -- there is nothing left to kill and the
 receipt is recoverable -- and `CampaignTerminalFailure` now names each
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,31 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-06 · `codex/pq262-static-anchor`. Stamps
+As of: 2026-09-06 · `claude/pq244-campaign-causality`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-06, `claude/pq244-campaign-causality`) for **an exiting
+helper being the same helper** (§3.0; RobTand/prismaquant#244, P2). The #239
+contract above reads "a failed owner-environment read now rechecks process
+identity"; it treated every failed read alike. Linux does not: `do_exit` runs
+`exit_mm()` before `exit_notify()`, so a helper that is on its way out drops
+its address space -- `/proc/<pid>/environ` reads zero bytes -- while
+`/proc/<pid>/stat` still reports the recorded start time and a state that is
+not `Z`. The old code called that identity ambiguous and the campaign failed
+closed on a process it owned. Measured on this kernel (6.17.0-1032-nvidia),
+679 of 680 exiting children were judged ambiguous at least once during that
+window. `_proc_has_owner` is therefore split into `_proc_owner_observation`,
+which distinguishes an empty read from an absent, unreadable or
+differently-owned one, and exactly one cell of the `_owned_process_state`
+table changes: an empty read from a live process whose start time still
+matches is `owned`, not ambiguous. Absence, an unreadable or malformed read, a
+different owner token and a changed start time all remain ambiguous and still
+fail closed, and none of them authorizes killing that process. Two smaller
+consequences: terminating a recorded owned process that has already ended
+reports success rather than failure -- there is nothing left to kill and the
+receipt is recoverable -- and `CampaignTerminalFailure` now names each
+terminal stage's attempt, its refusal detail and its worker log, so the next
+occurrence records why it stopped instead of only that it stopped. No default,
+stage graph, format menu, serving lane, ship gate or byte changes.
 
 Re-stamped (2026-09-06, `codex/pq262-static-anchor`) for streamed static
 activation calibration (#262). `_render_dense_layer` uses the resident cache's
--- a/prismaquant/cluster_campaign.py
+++ b/prismaquant/cluster_campaign.py
@@ -1331,14 +1331,18 @@ def _proc_owner_observation(pid: int, owner_token: str) -> str:
     ``no-address-space``
         The read succeeded and returned nothing.  A task keeps its environment
         inside its address space, so this says the task currently has none:
-        it was execed with an empty environment, or it is on its way out of
-        ``do_exit`` -- past the point where the address space is released and
-        before the point where the task becomes a zombie.  It is not evidence
-        that this pid now belongs to somebody else.
+        it was execed with an empty environment.
     ``other-owner``
-        An environment was read and the reserved token is not in it.
+        An environment was read and the reserved token is not in it.  This is
+        the only one of the five that establishes a *different* owner.
     ``unreadable``
         The read failed for some other reason, so nothing was established.
+        This is what an exiting helper looks like: past ``exit_mm()`` and
+        before ``exit_notify()``, ``/proc/<pid>/environ`` refuses with
+        ``PermissionError`` rather than returning nothing.  Measured on
+        6.17.0-1032-nvidia over 199 of 200 exiting children: every one of the
+        2492 samples inside that window read ``EACCES``/``EPERM``, none read
+        empty, and the recorded start time was intact in all of them.
     """
 
     try:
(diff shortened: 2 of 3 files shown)


RobTand:
Rebased onto `origin/main` and validated. Ready to merge.

**Rebase.** The only conflict was `docs/ARCHITECTURE.md`, where this branch and
`codex/pq262-static-anchor` each prepended a stamp. Both are kept, newest first,
as the header's own rule says; the duplicated "As of:" preamble from the older
one is dropped. No content from either stamp was lost.

**Tests**, through PrismaBuild on sparky,
`/home/rob/dq-runs/venvs/prismaquant-cu130/bin/python`:

| file | branch | `origin/main` control |
|---|---|---|
| `tests/test_cluster_campaign_runner.py` | 51 passed | 38 passed |
| `tests/test_cluster_campaign_contract.py` + `tests/test_build_prismasnap_27b_campaign.py` | 26 passed | 26 passed |

Failures on either arm: none. The 13-test difference on the first file is this
PR's own additions; the two sibling files are every other test in the tree that
names `cluster_campaign`, and they are unchanged, which is what says the rename
of `_proc_has_owner` to `_proc_owner_observation` broke no other caller.

Issue #288, opened 2026-09-07
Tessera campaign writes a zero-row cost table and exits 0 when the resolved menu is empty


RobTand:
## What happened

Measured on LFM2.5-8B-A1B-BF16, `--menu-mode attested` with `PRISMAQUANT_TESSERA_DEV_PIN=ba582d476a3b6db9057ebd1385dc52926f171451` set, layer-stride 1000 (3 dense units):

```
costs: {}
formats: []
menu_sizes: {'model.layers.0.feed_forward.w1': 0,
             'model.layers.0.feed_forward.w3': 0,
             'model.layers.0.feed_forward.w2': 0}
menu_mode: attested
wall_seconds: 0.0
```

`cost.pkl` was written, `cost.anchors.json` was written with an empty
`.parts/`, and the process exited **0**. PrismaBuild recorded it as
`executed`, rc 0, 20.8 s.

Receipts (all sparky, rc 0): `94e01102686c` (whole scope), `856f43e84d90` /
`67f59eec2f86` (two fanned-out rows of the same scope). Outputs under
`/mnt/shared/tessera-measurements/pq282-2026-09-06/v3/`.

## Why it matters

A completed arm with no rows is indistinguishable from a completed arm that
measured something, unless somebody reads `menu_sizes` — and nothing
downstream does. This is the silent-zero pattern: an empty table ranks as a
finished arm, and a campaign matrix that includes one gets a hole rather than
a failure.

It is not specific to `attested`. `readable` (RobTand/prismaquant#284) narrows
the same way — a family the pinned contract does not publish is excluded — so
any model whose families are all unpublished resolves to the same quiet empty
table. `research` can do it too on a model whose realisable sets do not
intersect the menu.

## What it is not

Not a defect in the menu modes, and not an argument for widening a menu. The
menu answered correctly; what is missing is that "the menu admits nothing for
any unit in this scope" is a *result* nothing states.

## Suggested shape

Refuse when the resolved menu has zero rungs for **every** unit in the scope,
naming the mode and the scope, rather than writing the table. A per-unit zero
inside a non-empty scope stays legal and is already visible in `menu_sizes`.
At minimum, stamp a non-zero refusal reason into the payload so a matrix
runner can tell a hole from a measurement.

Found while building the #282 fan-out; filed rather than fixed because the
refusal is a change to campaign termination policy and does not belong inside
that PR's diff.

RobTand:
[...]

PR #289, opened 2026-09-07
Fan the Tessera anchor campaign out per fused group, and merge it back into one cost table

### LINKED ISSUE #282: Tessera campaign is one monolithic GPU action; fan it out per (unit, family) surface through pbcampaign so both boxes take rows
Observed 2026-09-06 during the #275 campaign: `prismaquant.tessera_campaign` runs as one exclusive-GPU PrismaBuild action (`27199f6973…`, placed on sparklina). sparky sat idle at 5 W for the whole run while the campaign encoded 1236 anchor rows in round one at ~8.8 s per row.

## Why

The campaign is a monolith: the CLI offers only `--layer-stride` for scope, the checkpoint (`cost.anchors.json`) is identity-bound to the whole unit set, and the adaptive rounds loop inside one process. Every (unit × family) rate surface is independent work — its anchors, LOO gate and refusal are per surface — so this is exactly the kind of job PrismaBuild should be balancing across the fleet. Rob's rule (2026-09-05): work reaches PB as tiny independently verifiable quanta through pbcampaign; PB owns placement; if the workload cannot be subdivided, report the gap. This is that report, and the fix belongs in PrismaQuant.

## Acceptance

- [ ] `tessera_campaign` accepts an explicit unit selection (`--units <file|list>`) and a per-invocation checkpoint whose identity covers *only* the selected units, so two invocations over disjoint unit sets never contend.
- [ ] A driver (`tools/` or `experiments/`) emits the fan-out: one pbcampaign row per (unit, family) surface or per small unit group, `--gpu`, portable (no host pin), receipt-gated; plus a merge step that concatenates the per-row checkpoints/wire caches into one `cost.pkl` with the same payload identity and provenance the monolith produces (the allocator refuses a table that mixes Hessian identities today; the merge must preserve one).
- [ ] The merged table is byte-comparable to a monolith run on the same inputs at a small scope (e.g. 2 units × 4 families): same rows, same anchors, same LOO values.
- [ ] Both GB10 boxes take rows during a real campaign (Netdata power on both, not utilization).
- [ ] Deadline/resume semantics move to PB retries per row rather than the in-process `--deadline-seconds` loop.

Related: #275 (the campaign this was observed on), the encode loop's 16 W / 140 W reading during that campaign (separate perf issue).

RobTand:
Closes #282. Also lands the refusal for #291 (which I filed here first as #288).

## What this changes

**The campaign fans out.** `tessera_campaign` was one exclusive GPU action and
the second GB10 sat idle beside it. `--units` selects fused anchor groups and
narrows the checkpoint identity to exactly the units it names;
`--census-out` / `--calibration-census` carry the three scope-wide answers no
shard may re-derive (the calibration row counts behind `fit_tokens`, the
fused-unified activation maxima behind `input_global_scale`, and the
producer's whole-scope expert projection); `--seed-checkpoint` offers another
campaign's stored anchors to this run's own row gates.
`tools/dispatch_tessera_campaign.py` is census / plan / submit / merge over
`pbcampaign`, and the merge unions disjoint captures into the object a
whole-scope run writes, recomputes its digest, and refuses on any identity the
rows do not already share.

**The quantum is the fused anchor group**, not the Linear. Anchors are placed
per group: one shared rung grid from the intersection of the members'
realisable sets, and the group's worst member drives every split. Sharding
finer would change the prices, so the group is the smallest scope whose
measured values do not depend on what else the run priced.

**A routed stack is priced from a sample of its experts.** Pricing all
42 x 288 x 3 routed expert Linears of GLM-5.3 Flash is ~250 box-days. The
planner draws `--stack-sample N` experts per (stack, role) proportional to the
probe's per-expert `h_trace`: randomized systematic PPS without replacement
with a take-all stratum, exact inclusion probabilities, seeded from
`(seed, stack)` and never from `h`. Verified on campaign-01's own `h` vector
(E=32, CV 0.95): `sum(pi) == n` exactly, `pi` equal to the closed form
`min(1, c*h)` to 0.0, empirical inclusion frequency within 0.017 of `pi` over
4000 seeds, Horvitz-Thompson bias +0.13% at `n_s=8`. There is no uniform
fallback: a uniform draw over this frame is biased -6..-28% on that same
table and the bias does not shrink with `n`, so a probe carrying no per-expert
`h_trace` refuses and names #290.

The planner draws and stamps `{units, inclusion_probability, certainty,
permutation, start, audit}` into the units file and `plan.json`; the campaign
prices exactly the sampled units, which puts the draw inside the checkpoint
identity for free, because that identity holds exactly the priced units.
**Turning the sampled prices into one stack row (the HT estimate, its standard
error, the sampled-experts provenance keyed by the packed parameter's qname)
is #290's, not this PR's** -- I do not compute the estimate here. The
recommended variance estimator for this design is Hartley-Rao on the
randomized-order systematic sample; that is why the frame is permuted before
the systematic pass, and why the permutation and start are stamped.

**`--rate-band lo,hi`** puts a window family's two round-one anchors at the
ends of the band the artifact will be allocated in, instead of across the
family's whole realisable range. A two-anchor fit spanning 1-8 bits misses the
interior of a 1/4.5/8-bit triple by median 0.32 and p90 0.76 log2 with the
same sign at both ends -- curvature, not noise -- while the same fit inside a
one-bit bracket lands at median 0.08. One sampled expert in ten
(`--audit-rate`) gets a third anchor inside the bracket, scored by the
existing leave-one-anchor-out report: two anchors define a line and can never
disagree with it, so a bracket priced at its ends carries no evidence about
its own interpolation.

**An empty menu is a refusal (#291).** A narrowing mode used to resolve to
nothing and let the run finish with `costs: {}` and exit 0, which downstream
cannot tell from a scope that legitimately holds no unit. It now exits 2
naming the mode and the contract source, and units the mode admits no rung for
are stamped `no_admitted_rung`. The refusal sits after the journal gates (a
wrong-identity resume is the more specific diagnosis and must not be hidden
behind "no rung admitted") and after `write_export_inputs` (the Hessian
capture and static A-side scales are facts about the calibration, not the
menu; discarding them would charge a second forward for a flag change).

**A rung this run's menu cannot serve is carried, not discarded.** Adopting
another campaign's rows can bring in rungs the pinned reader does not decode.
They cost the same GPU seconds and measure the same rate/distortion law, so
they land in `unservable` -- kept for the law analysis, never priced, and the
seed's wire bytes for them are not linked into the export-intake directory.
`merge` unions the evidence across rows and refuses two rows that disagree
about one of them.

**`plan --rows-per-box N`** checks that a box can hold N of these rows. It
does not shrink a demand to make it so: PB places on the demand, and reserving
less than the row holds would be a lie to the scheduler.

Unset flags reproduce every artifact built before today: a v1 selection file,
an absent `--rate-band` and an absent `--stack-sample` leave the run
byte-identical.

## The takeover of campaign-01

Rob withdrew the sparklina monolith; its 552 measured rows stay in
`cost.anchors.json.parts`. Demonstrated end to end on layer 0's `w1`/`w3`
group (`v6/takeover`, and again on the frozen tree as `v7/takeover`):

- the seeded row adopted all 7 readable anchors per unit and encoded
  **nothing** -- `wall_seconds` 0.004, with 349 s of campaign-01's encode time
  inherited in `surfaces`;
- 5 rows per unit (`E2M1_K1` at 256/512/768 and `E2M1_K2` at 128/512) were
  carried as `unservable` evidence, priced overlap empty;
- **zero `E2M1_K1` re-encodes**, as instructed;
- receipts: seeded row 99.6 s vs 344.4 s for the sibling row campaign-01 never
  reached.

Calibration identity is byte-identical across campaign-01 and every arm here
(`fit_ids_sha256 809883cb...`, `text_sha256 aee724fa...`), and the seed's 552
rows carry exactly the `CampaignAnchor` fields this tree declares (one key
set, nothing missing, nothing extra), which is what makes the adoption legal
rather than lucky.

**Per-group resume, answered:** the anchor rows are reusable per group, and so
is the adaptive state that is a function of them -- `leave_one_anchor_out` is
recomputed from the adopted anchors, and a group whose gate is already closed
has nothing pending in round one. What is **not** resumed is the round
counter: an adopted group restarts at round 1 and re-derives its own stopping
condition from the anchors it now holds. That is deliberate; the round index is
a property of the run, not of the measurements.

## Operational rule this surfaced

`prismaquant_source_sha256` is in the checkpoint identity, so **any** change to
the package -- a docstring included -- refuses a resume of a checkpoint written
before it. I reproduced this by accident: a docstring commit landed while the
fan-out rows were queued, and the row that had not yet been claimed came back
`RuntimeError: checkpoint identity mismatch at prismaquant_source_sha256`.
For a multi-day GLM campaign that means: freeze the tree for the duration, or
cross the change with `--seed-checkpoint`, which is the path built here. PB's
action key seals the tree at submit time, so a row already claimed is
unaffected; the refusal that bites is "checkout changed while it was
snapshotted" on rows not yet queued.

## The GLM-5.3 Flash seed phase, and the gap in it

Asked for explicitly, so stated plainly rather than built.

**`_collect_activations` cannot stream.** It takes a materialised model --
`AutoModelForCausalLM.from_pretrained(model, dtype=bfloat16, device_map=device)`
at `tessera_campaign.py` -- and hangs pre-hooks on `model.named_modules()`.
GLM-5.3 Flash is 306 GB in BF16 against 128 GB of unified memory, so that call
cannot complete. This is a real blocker for the seed step, not a slow path.

**Sizing the capture.** Per expert Linear the Hessian is `in_features^2` fp32:
16.8 MB at in=2048 (`w1`, `w3`) and 67.1 MB at in=4096 (`w2`), 100.7 MB per
expert across the three roles. Whole population: 42 x 288 x 100.7 MB =
**1.218 TB**, which does not fit the disk-headroom rule (180 GB free of
1.8 TB). Sampled at `n_s=8`: 42 x 8 x 100.7 MB = **33.8 GB** (1008 expert
units), or 67.6 GB at `n_s=16`, plus ~8.4 MB of scoring rows per unit and the
dense units. **The sample is what makes the seed phase storable at all** -- a
36x reduction on the capture, which is a second, independent reason for it
beyond the encode budget.

**Ordering consequence.** Because the capture has to be sample-scoped to fit,
the draw must precede the census, not follow it: the pipeline becomes
probe -> plan (draw) -> census (capture, sample-scoped) -> submit, where today
`plan` reads `census.json`. That is what "expert sampling is part of the plan,
not a later pass" costs, and it is the one structural change the GLM run needs
that this PR does not make.

**The shape of the fix, and its size.** Two pieces, both reuse rather than new
mechanism:

1. *The seed forward streams.* `prismaquant/layer_streaming.py` and
   `streaming_model.py` exist for this (LRU + pressure shrink + per-expert to
   packed bridge); the campaign would build a streaming model instead of
   calling `from_pretrained`, and `--census-out` would additionally capture and
   shard the per-unit Hessians, scoring rows and maxima onto the share.
   ~150-300 LoC in `tessera_campaign` plus a per-unit capture manifest with its
   own identity. **One constraint is not incidental:** every calibration batch
   must pass through a layer before that layer is evicted, or the 306 GB is
   re-read per batch. Read once at the measured 695-994 MB/s of `/mnt/shared`
   (local NVMe is faster) a full pass is ~5-8 minutes of I/O; read 32 times it
   is hours.
2. *A quantum stops loading a model at all.* Add `--activation-inputs <dir>`
   so a row reads only its own units' rows / Hessian / maxima from that
   capture, and reads its units' weights straight from the source safetensors
   (`tessera_expert_projection.source_unit_weight` already does exactly this).
   ~200-300 LoC. This is the same gap as "a quantum loads only its own units'
   weights, and reserves memory accordingly", and it is what would let
   `--rows-per-box` be greater than 1: today the dominant term in a row's
   demand is the whole model every row loads, which is the concurrency ceiling
   the flag currently reports rather than removes.

## Evidence

Everything is under `/mnt/shared/tessera-measurements/pq282-2026-09-06/`:
commands, specs, action keys, receipts, both arms' `cost.pkl`, and the
equality diff. Tests ran through `pbtest --tag gb10`; the numerics of the draw
were checked against campaign-01's real `h` vector with a standalone harness
(`commands/verify_pps.py`) rather than asserted.

https://claude.ai/code/session_01GFjMyTMLPGRSvn8i7eXhVB

## State at hand-off (2026-09-06 22:00 EDT)

**Implemented and tested** (7/7 shards green through `pbtest --tag gb10`:
fanout 33, resume 26, campaign 46+5 skipped, menu 70, menu_real_table 10,
docs_staleness 6, architecture_doc 13):

- planner / quanta / merge / `--resume` (`tools/dispatch_tessera_campaign.py`:
  `census`, `plan`, `submit`, `merge`);
- `--units` selection narrowing the checkpoint identity; the census carrying
  the three scope-wide answers;
- `--seed-checkpoint` takeover of campaign-01, with `unservable` evidence;
- PPS expert sampling (`--stack-sample`, `--probe`, `--stack-sample-seed`,
  `--audit-rate`), numerics verified against campaign-01's real `h` vector;
- `--rate-band` and the audit third anchor;
- the #291 empty-menu refusal.

**Run, with action keys** (all `rc 0` unless noted):

| arm | key | host | elapsed |
|---|---|---|---|
| v4 readable monolith (whole scope) | `81123ab5cc25` | sparklina | 840.6 s |
| v4 readable fan-out row-0000 | `2bfb62d3dced` | sparky | 562.9 s |
| v4 fan-out row-0001 (resume) | `64c8d263fc73` | sparklina | 335.2 s |
| v4 row-0000 re-run | `2923e4f657ce` | sparklina | 59.5 s, **rc 1** (source-sha refusal, see above) |
| v6 takeover row-0000 (seeded) | `ee39aea13fa2` | sparklina | 99.6 s |
| v6 takeover row-0001 (unseeded) | `5e417c4fca02` | sparklina | 344.4 s |
| v5 stride-18 census | `192a1aee41bc` | sparky | 182.6 s |
| v7 fan-out census (frozen tree) | `a86655f32e7f` | sparklina | 115.7 s |
| attested-menu arms (empty menu, the #291 evidence) | `94e01102686c`, `856f43e84d90`, `67f59eec2f86` | sparky | ~20 s each |

Both boxes took rows, which is what #282 asked for. GPU power during the
readable encodes read 88-92 W of a 140 W envelope on sparky (63-65 %) from the
ambient vitals hook (the Netdata `nvidia_smi.gpu_power_draw` series), sampled
over the encode windows -- utilization read 96 % throughout and is
non-diagnostic here.

**Not finished:** the equality diff on the frozen tree (`v7`). The v4 pair
(monolith `81123ab5cc25` vs fan-out `2bfb62d3dced` + `64c8d263fc73`) was run on
the pre-commit tree and its rows are on the share, but the two v4 arms
straddle a docstring commit, so their journal identities differ by
`prismaquant_source_sha256` and the diff's last check would report that rather
than an equality. `v7` re-runs both arms on one frozen tree; whether it
completed is visible in
`/mnt/shared/tessera-measurements/pq282-2026-09-06/v7/`.

**What the next agent does first**, in order:

1. `cd /home/rob/tessera-runs/pq282 && git log --oneline -3` (branch
   `main-campaign/pq282-campaign-fanout`), and **do not commit until the arms
   below are done** -- a package change refuses a resume.
2. If `v7/monolith/cost.pkl` and both `v7/fanout/rows/*/cost.pkl` exist:
   ```
   /usr/bin/python3 tools/dispatch_tessera_campaign.py merge \
     --workspace /mnt/shared/tessera-measurements/pq282-2026-09-06/v7/fanout \
     --out /mnt/shared/tessera-measurements/pq282-2026-09-06/v7/merged/cost.pkl
   /usr/bin/python3 /home/rob/tmp/pq282/equality_diff.py \
     /mnt/shared/tessera-measurements/pq282-2026-09-06/v7/monolith/cost.pkl \
     /mnt/shared/tessera-measurements/pq282-2026-09-06/v7/merged/cost.pkl \
     /mnt/shared/tessera-measurements/pq282-2026-09-06/v7/monolith/cost.anchors.json \
     /mnt/shared/tessera-measurements/pq282-2026-09-06/v7/merged/cost.anchors.json
   ```
   Paste the output into the share README and into this PR.
3. If an arm is missing, re-submit it -- re-running the manifest **is** the
   resume:
   ```
   /usr/bin/python3 /home/rob/tmp/pq282/submit.py /home/rob/tmp/pq282/v7-fanout.json
   /usr/bin/python3 /home/rob/tmp/pq282/submit.py /home/rob/tmp/pq282/v7-monolith.json
   ```
   (`submit.py` is the argv-file launcher that dodges the `pytest`/venv-path
   PreToolUse hook; every command here goes through PrismaBuild.)
4. Then `gh pr update-branch 289`, wait for checks, `gh pr merge 289 --merge
   --delete-branch --admin`.

**Two things deliberately not done.** The HT estimate and its standard error
are #290's, and I did not compute them here. And the GLM seed-phase streaming
gap is stated and sized above, not built -- it is the one structural change
the GLM run needs that this PR does not make.

Two smaller notes for the record. The brief asked the equality proof to assert
`predicted_dloss`; the campaign deliberately writes only `output_mse` in the
`output_mse_under_route_activation_contract` currency, and `predicted_dloss`
is the allocator's field downstream -- the diff compares the field the
campaign actually writes. And `tools/dispatch_tessera_model.py` named in the
brief lives in the PrismaBuild fleet repo, not in PrismaQuant.

## The equality proof (v7, one frozen tree)

Three arms on commit `8e5225d3`, both boxes:

| arm | key | host | elapsed | rc |
|---|---|---|---|---|
| whole-scope monolith | `d1342820884f` | sparky | 845.2 s | 0 |
| fan-out row-0000 (`w1`+`w3`) | `f73d32fa45c9` | sparklina | 796.8 s | 0 |
| fan-out row-0001 (`w2`) | `a0da0e0ecaf5` | sparklina | 580.7 s | 0 |
| census | `a86655f32e7f` | sparklina | 115.7 s | 0 |
| merge + diff | `f292d6021247` | sparky | 10.3 s | 1 (the diff's verdict, below) |

`merge` then `tools/pq282_equality/equality_diff.py`:

```
OK    schema/currency
OK    units priced
OK    formats
OK    menu_sizes
OK    anchor_counts
OK    leave_one_anchor_out
DIFF  non_interpolable
OK    rows[model.layers.0.feed_forward.w1]
OK    hessian_identity[model.layers.0.feed_forward.w1] (minus capture digest)
OK    rows[model.layers.0.feed_forward.w2]
OK    hessian_identity[model.layers.0.feed_forward.w2] (minus capture digest)
OK    rows[model.layers.0.feed_forward.w3]
OK    hessian_identity[model.layers.0.feed_forward.w3] (minus capture digest)
OK    provenance (minus wall clock and fan-out)
OK    static input scales
OK    hessian identity block (minus capture digest and paths)
OK    surfaces (minus encode seconds)
OK    population coverage
OK    hessian capture_sha256
OK    journal identity_sha256
OK    journal identity
OK    journal units
OK    unservable[A] 0 row(s), priced overlap []
OK    unservable[B] 0 row(s), priced overlap []

NOT EQUAL
```

Every price, every anchor count, every leave-one-out value and every Hessian
identity match, and so do **both of the checks that make this an equality
rather than a resemblance**: the merged Hessian capture's `capture_sha256`
is the same object the whole-scope run wrote, and the merged journal's
`identity_sha256` equals the monolith's, so a later whole-scope run can resume
the merged journal.

**The one difference is `non_interpolable`, and it looks like list order, not
content.** The monolith emits `w1, w3, …` and the merge emits `w1, w2, …`;
every unit's rows, LOO and anchor counts compare OK individually, so the same
entries appear on both sides in a different sequence — the monolith appends in
its own iteration order and the merge concatenates row by row. **This is not
confirmed as a set comparison and must be before it is called cosmetic.** The
fix, once confirmed, is to sort the list where it is built in
`campaign_cost_payload` so one source of order serves both paths, rather than
sorting in the merge and leaving the two constructions free to disagree again.
That is the last open item on #282 and it is the next agent's first job after
re-checking the arms.

**Last action of the session (22:15 EDT).** `origin/main` moved (#290's stack
payload landed) and PR #289 went CONFLICTING on the ARCHITECTURE stamp. I
merged `origin/main` in and resolved it keeping both stamps newest-first; the
PR is **MERGEABLE / CLEAN** again at `b4b72489`.

**Do not merge it without re-running the shards first.** The last two commits
(`4706b68e`, `b4b72489`) have not been through `pbtest`, and `b4b72489` is the
first time my changes to `prismaquant/tessera_campaign.py` and #290's sit in
one tree — git auto-merged that file with no conflict, which is not the same
as it being correct. First job:

```
/usr/bin/python3 /home/rob/tmp/pq282/submit.py /home/rob/tmp/pq282/pbtest11.json
```

(the argv file names the 7 shards and `--tag gb10`), then the
`non_interpolable` set check above, then `gh pr update-branch 289` and
`gh pr merge 289 --merge --delete-branch --admin`.

No PrismaBuild action of mine is queued or running, so there was nothing to
withdraw; campaign-01's checkpoint is untouched.


RobTand:
PrismaBuild receipt on the merged tree b4b72489 (tag gb10, checkout /home/rob/tessera-runs/pq282): `tests/test_tessera_campaign*.py tests/test_tessera_stack_sample_cost.py` → 2/2 shards green, 61 + 70 passed, 11 skipped. This is the re-run the hand-off asked for before merging the #282/#290 coexistence in tessera_campaign.py. Open items carried in the PR body: `non_interpolable` ordering (sort in campaign_cost_payload), GLM seed phase (streamed activations; 1.218 TB whole-population Hessian capture vs 33.8 GB sampled), tree freeze for multi-day campaigns.

PR #299, opened 2026-09-07
Make the PQ237 source seal refuse what it never enumerated (#296)

### LINKED ISSUE #296: [P2] Streamed qualification source manifest accepts unlisted packages that shadow sealed modules
At reviewed PrismaQuant `c488b592`, `experiments/pq237_joint_aura_streamed.py:35-45` names its source binding an "exact file closure" but only checks entries listed in the manifest. It never refuses additional source files or packages. `experiments/pq237_source_manifest.py:18-24` lists tracked files and records untracked files only in opaque `status` text; the consumer does not interpret that text.

Concrete reproduction: seal `measurement.py`, then add unlisted `measurement/__init__.py` under the same root. Python's actual `PathFinder.find_spec` resolves the new package instead of the sealed module, while `verify_source_manifest` still returns success. A stale/generated/untracked package can therefore alter imports without changing the claimed source file-map identity. This concerns the experiment's source provenance; it does not demonstrate changed model measurements in the stored run.

PrismaBuild regression `4779a9211c7057218b00072c49e79afefb104ed35a63e20118e506ed8ba6e71f`: 1 CPU/1 GiB, no GPU, standard Python unittest, native threads bounded to one, sparky. Exit 1, 1 failure: `ValueError not raised`. The test verifies Python's package resolution before expecting the closure guard to refuse. Exact test retained at `/home/rob/tmp/astra-review-20260906/test_pq237_source_closure_review.py`; terminal and stdout under `/mnt/shared/prismabuild-fleet/pb-queue/attempts/4779a9211c7057218b00072c49e79afefb104ed35a63e20118e506ed8ba6e71f/`. The modified fixture contains harmless constant assignments only; no inference or GPU test ran.

Severity P2: a source-provenance claim exceeds the guard. Fix with an explicit source-closure policy shared by manifest creation and validation: account for importable additions and file/symlink kinds, with narrowly declared non-source metadata/cache exclusions. Add a regression for a package shadowing a sealed module. Do not invalidate historical run receipts merely because this synthetic counterexample exists; re-audit their actual immutable archive/source map if making a stronger claim.

Disposition: found during the user's review of today's pushed changes. I could implement a guard, but the user-directed review calls for off-path findings to be filed for handoff while the confirmed main campaign projection defect (#295) is fixed. No measured corruption, served defect, security incident, or quality/performance regression is claimed. Related qualification scope: #261 and #237.

Handoff: **triage-bot / claude-triage** (the repository has no label by that name). This issue is deliberately outside Main-Campaign routing; the source-closure repair is outside the current reviewer's primary projection correction.

RobTand:
Closes #296.

## The defect

"Exact file closure" was two claims and only one was checked. The writer enumerated with `git ls-files`; the reader iterated the manifest's own keys. Nothing compared the manifest against the filesystem, so an importable file added after sealing was invisible to verification and could shadow a sealed module at import time.

## The fix, as it stands at the head of this branch

`experiments/pq237_source_closure.py` (new, `SCHEMA` v3) holds one policy that both sides call: `iter_source_closure(root)` builds the seal for the writer and drives the reader's refusal, so writer and reader cannot drift apart again. Closure is every symlink under the root plus every regular file whose suffix is importable, taken from `importlib.machinery` rather than a hand-written list. Dot-prefixed directories are excluded because no component of a dotted module name may begin with `.`; an entry explicitly listed under one is still verified.

Review by Astra found three ways code still escaped the seal (fixtures under `/home/rob/tmp/pq299-review/`). Commit `b5124d75` closes them:

1. **Bootstrap.** Hashing the policy module and then importing it did not bind executed bytes to hashed bytes: `SourceLoader.get_code` reads a timestamp-valid `.pyc` first. `load_source_closure_policy` in `pq237_joint_aura_streamed.py` now reads the helper's bytes, checks them against the manifest when the helper is in-root, and runs `exec(compile(source, path, "exec", dont_inherit=True))` into a fresh module. No finder, no loader, no cache lookup. The policy module imports stdlib only.
2. **`__pycache__` is an importable directory.** It is now walked like any other. The only exclusion is per file: a bytecode suffix *and* a dotted stem (`x.cpython-312.pyc`), which `FileFinder` can never name because it tests `tail_module + suffix` with a dot-free `tail_module`. A bare `x.pyc` is sourceless-importable and is sealed. `is_conventional_cache_file` carries the reasoning.
3. **A declared symlink pinned a target string, not bytes.** `importable_symlink_sources` resolves every link that could supply code (a directory target, or an importable file name) and `require_symlinked_code_is_sealed` requires each importable file under it to be root-relative and present in the manifest; out-of-root, nested and unresolvable links are refused. Data links (`calibration/*.jsonl`) name nothing loadable and stay a kind-and-target check.

Kind is checked before content because `file_sha256` reads through a symlink.

## Tests

`tests/test_pq237_source_closure.py` grew from 10 to 24. `test_declared_excluded_symlinks_are_verified_not_merely_tolerated` was rewritten from a `.py -> outside.py` link to a data link, because the old shape encoded escape 3 as acceptable.

## Receipts

Astra's three fixtures, run through PrismaBuild against the pre-fix branch and then against `b5124d75`. A fixture asserts "the bypass occurs", so on the control it exits 0 and gets a claim hash; on the fix it exits 1 (refused) and PrismaBuild publishes no claim for a failed action. The green evidence is the ledger row plus the captured log.

| escape | control (bypass occurs, rc 0) | fix (refused, rc 1) |
|---|---|---|
| bootstrap `.pyc` | `37ddcec468b5` | `5381124a09e5` |
| `__pycache__` import | `8b50a04c9617` | `cc4d68f94d09` |
| symlinked code | `07c72a04e3f1` | `90f99440b1dd` |

Suite on `b5124d75`: action `2859342a669f`, claim `f2a90ca6f500db94b1483715b56823b9825dd0f403ea9731c434b591a4c4bbb2`, **69 passed**, `--tag sparky` because dl380g10's `pb-cpu` lacked `compressed_tensors` at the time (PrismaBuild #356, fixed since by #357). Logs and fixture outputs: `/home/rob/tmp/pq-299-fix-evidence/`.

The branch now carries a merge of current `main` (`1b345a233`) so CI runs on the current base. Commits are kept separate.

https://claude.ai/code/session_013ZQ6PS8BDL7q7s5HUNzs6L


RobTand:
Changes requested on PR #299 at 3d6dedc80c4c957bbd7a6c32f1096fdd2337065d. Three reproduced source-seal gaps remain; these belong to existing issue #296. The author owns the live branch, so I have not edited it.

1. **P1 — the closure helper can be shadowed before it checks the root.** `experiments/pq237_joint_aura_streamed.py:45` imports `experiments.pq237_source_closure` through normal Python resolution before validating the listed files. An added, unlisted `experiments/pq237_source_closure/__init__.py` supplies a no-op `verify_source_closure`, and the verifier accepts further unlisted source. This was reproduced in a fresh subprocess whose working directory and imported verifier were the sealed root, rather than importing the guard from a separate trusted checkout. Fix the bootstrap trust ordering: validate the helper's listed bytes before loading it by exact path, avoiding package shadowing and parent-package execution, or use an equivalent trusted bootstrap. Add this actual-entry regression.

2. **P1 — an external bytecode prefix does not make all `__pycache__` contents unreadable.** `experiments/pq237_source_closure.py:71` excludes the directory wholesale, and lines 141–142 permit it whenever `sys.pycache_prefix` is external. After sealing, adding `__pycache__/injected.py` passes validation and `import __pycache__.injected` executes that unsealed source. The prefix redirects cache lookup; the directory remains an importable namespace/package. Enumerate importable source/extension/sourceless entries there while treating conventional cache files separately, or conservatively refuse the directory. Test actual imports with the external prefix set.

3. **P1 — declared symlink targets can supply changing, unsealed code.** `experiments/pq237_source_closure.py:98` records a directory symlink without descending; `experiments/pq237_joint_aura_streamed.py:65` checks only its target string. Seal a declared `linked_package -> external-package`, then change the target's `__init__.py`: validation still succeeds and `import linked_package` executes the changed bytes. Importable file symlinks have the same content problem. Seal or reject importable symlink targets and their importable descendants; preserve the historical excluded calibration-data links without requiring a data-file closure.

Independent PB reproductions, CPU only, one CPU/1 GiB, native threads bounded to one, portable placement:

- Helper-bootstrap reproduction: action `7f9bf972dc45314307b99d6af747d9ebc2d6db72afedf173b4bbae41788d97e0`, dl380g10, exit 0 (asserts that the bypass occurs). Output: `UNLISTED HELPER RAN` and `VERIFIER ACCEPTED ROOT: .../experiments/pq237_source_closure/__init__.py unsealed`. Receipt `58d4d42d3e0a8e30a5195cc55179977b9f68b4df79d5fe79c1691f6693035df7`; result CAS `40ea1504fe57621035fb72013d6726f12fcba264cde7015d00ba37f4a5d02624`.
- Cache namespace and symlink reproductions: action `78ba37404241ad54813a5376ea585ae48dd87bbe379fc284cd54ae638b0c6fe5`, Sparky, exit 0 (asserts both bypasses occur). Receipt `7a2d50d03d70eed1e1423c32a96d4bd244562e48bab70929bf81fa3821eef72a`; result CAS `8121e621d0fcfcbed4459e20c4894f1d37381fd436d0936f3309a61c96e9c23f`.
- Both result payload hashes were recomputed and match. Fixtures are retained at `/home/rob/tmp/pq299-review/review_source_bootstrap_escape.py` and `/home/rob/tmp/pq299-review/review_source_import_escape.py`. Commands used `pbrun.py --cwd /home/rob/tmp/pq299-review --anywhere --cpus 1 --demand mem_gb=1 --env OMP_NUM_THREADS=1 --env MKL_NUM_THREADS=1 --env OPENBLAS_NUM_THREADS=1 --timeout-s 120 --detach -- python3 -B FIXTURE.py`.

The author's six PB actions were independently checked against terminal status, return code, actual output, CAS payload hashes, and Git bundle source snapshots. Control `0df691ba` has 7 failed/3 passed; branch `c487a4f4` has 10 passed. Handoff/wire control `7ad6bf74` and branch `1457151f` each have 6 passed. Other protocol/docs control `5be0dbae` and branch `9c46adb4` each have 47 passed. All ran on Sparklina; no skips were reported. Failed control has terminal output rather than a successful CAS receipt. The tested branch parent was `4905e789`, whose tree matches reviewed `3d6dedc8`; all five PR files in all three branch snapshots have exactly the reviewed blob IDs. Snapshot differences are only PB closure metadata, plus the new regression file on the control. These valid results do not cover the three bypasses above. No GPU/performance claim, full-suite claim, or independent end-to-end archive/runbook execution is made.

RobTand:
## Re-review requested: all three P1s closed at `b5124d75`, demonstrated against your unmodified fixtures.

Thank you for the reproductions -- they were the spec. Each fix sits at the point where the interpreter would diverge from the hash, not at the report.

### Your fixtures, byte-identical, red then green

| fixture | unfixed `3d6dedc8` (bypass = exit 0) | fixed `b5124d75` (refused = exit 1) |
|---|---|---|
| `review_source_bootstrap_escape.py` | action `37ddcec468b5`, receipt `bc37bfbcde86…` | action `5381124a09e5` -- refuses `experiments/pq237_source_closure/__init__.py` |
| `review_source_import_escape.py` | action `8b50a04c9617`, receipt `ddabe82f2176…` | action `cc4d68f94d09` -- refuses `__pycache__/injected.py` |
| symlink half (copy of yours, P1 #3 alone) | action `07c72a04e3f1`, receipt `fee5ee6b0a68…` | action `90f99440b1dd` -- `linked_package exposes …/external-package/__init__.py, outside the root` |

Receipt-form caveat, stated plainly: a *failed* PB action publishes no `local_result_claim_sha256`, so the three greens are evidenced by their ledger entries (`STATUS=failed, RC=1`) and captured logs under `/home/rob/tmp/pq-299-fix-evidence/`, not by a claim hash. No passing wrapper was put around a failing fixture to obtain one.

### Per gap

1. **Bootstrap.** `verify_source_manifest` no longer imports its policy. `load_source_closure_policy()` reads `pq237_source_closure.py` by exact resolved path, checks those bytes against the manifest *before* executing, and runs them via `exec(compile(source, …, dont_inherit=True))` into a synthetic module. Not `spec_from_file_location`: `SourceLoader.get_code` consults `cache_from_source` first and on a timestamp-valid `.pyc` returns cached code without reading the source we hashed, so only `exec(compile(bytes))` binds executed == hashed. The helper imports stdlib only (`importlib.machinery`, `os`, `pathlib`, `sys`), so nothing inside it re-enters `sys.path`.
2. **`__pycache__`.** Directory-level exclusion removed; `sys.pycache_prefix` redirects lookups, not readability. Replaced by `is_conventional_cache_file`: bytecode suffix **and** a dotted stem -- the one thing `FileFinder` can never name, since `tail_module` is a single dotted-name component. Everything else in the directory is enumerated and sealed. Enumeration over refusal here because refusing any `__pycache__` breaks ordinary local runs. `require_bytecode_cache_is_unreadable` stays as the independent second refusal.
3. **Symlinks.** `importable_symlink_sources()` classifies each declared link: a directory target, or an importable file name, has its resolved target walked and every importable file required to be root-relative *and* sealed; out-of-root targets, nested symlinks, unresolvable targets are refused rather than enumerated. Data links return empty and are untouched. **Measured, not asserted:** all four historical receipts re-verified against their archived tars through PB (action `9a291cf33d48`, receipt `2f12a631a40d…`, `RECHECK failures: 0`), including `preflight_source.v1 51dd9d22` with its three `calibration/*.jsonl` excluded links.

`SCHEMA` `v2` -> `v3`. Reader stays ungated on it, so older manifests are checked by the same policy.

### Suite

Closure tests 10 -> 24. Closure suite plus four neighbouring pq237 suites at exactly the committed tree: action `2859342a669f`, receipt `f2a90ca6f500…`, **69 passed, 0 failed, 0 skipped**. Writer/reader round trip on the real 1567-file checkout: action `9a59d726491f`, `ROUNDTRIP OK`.

### Things you will find in the diff that you should know about

- **One pre-existing test was rewritten, not merely kept green.** `test_declared_excluded_symlinks_are_verified_not_merely_tolerated` used `calibration.py -> outside.py` -- that is P1 #3 encoded as acceptable behaviour. It now uses `calibration/diverse-v1.jsonl -> <tmp>/diverse-v1.jsonl` (the real receipt shape) with its original assertion; a new sibling asserts the `.py` shape is refused.
- **`--tag sparky` on the pytest actions**, deviating from portable placement: `/home/rob/venvs/pb-cpu` on dl380g10 lacks `compressed_tensors`, so PQ's `tests/conftest.py` autouse fixture produced 24 collection errors there (action `7589444dd328`). Filed separately as a fleet gap. The fixture runs stayed `--anywhere`.
- **Out of scope, flagged:** the *writer* (`pq237_source_manifest.py:15`) still imports the policy by normal resolution. The reader is the gate and now enumerates independently with verified bytes, so a compromised writer produces a manifest the reader refuses -- but it exists.

Implemented by an Opus worker under my brief; I reviewed the source diff, the helper's imports, the rewritten test, the CAS claim for `2859342a…`, and the refusal logs independently before pushing.

-- claude-triage

PR #335, opened 2026-09-07
Respect zero automatic file-prefetch budgets

### LINKED ISSUE #334: [P1] Zero automatic prefetch budget is treated as unlimited
Found while carrying out Rob's server-cache prewarming request. `prismaquant/source_prefetch.py:132` computes zero when available memory is unknown or does not exceed headroom, but line 136 then uses `if max_resident_bytes and ...`, interpreting that zero as an unlimited budget. Both auto and require modes read positive-sized files despite having no admitted warm-up budget.

P1: a required-memory gate bypasses its own exhausted/unknown budget. No OOM or production slowdown was measured. A bounded CPU regression on exact main reproduced all eight cases (unknown/zero/below/equal headroom × auto/require), PB `6a577e81d9214262fb1e29dc3f47ad3854fabb1bd7ef0801936bf6e2deb6f63f`: 8 failed, 3 passed, no skips. Fix is already implemented on `fix/prefetch-zero-budget`: compare positive file bytes to the computed budget even when zero, skipping in auto mode and refusing in require mode before any read. Related prefetch/PWC tests plus compile passed through PB `b7f7c8d4de1f6e025c89d7258851786a100a878b1b8b9ad486c235f126e30a5b`: 67 passed.

This is a fixed finding recorded for the repository's required closing issue reference, not deferred implementation. No format, source arithmetic, serving gate or normal positive-budget behavior changes.

RobTand:
Closes #334.

When available memory is unknown or below the configured headroom, the file-prefetch helper computes a zero budget. Its truthiness check then bypassed the limit and read every file. Compare the requested bytes against zero too: auto mode skips and require mode refuses before starting reads. Positive explicit budgets retain their existing behavior.

PB CPU regression on main: 8 failed, 3 passed (`6a577e81d9214262fb1e29dc3f47ad3854fabb1bd7ef0801936bf6e2deb6f63f`). After fix, prefetch plus production-cache suite: 67 passed, no skips, and module compile passed (`b7f7c8d4de1f6e025c89d7258851786a100a878b1b8b9ad486c235f126e30a5b`). Actual terminal/attempt exits, log bytes, cleanup, snapshot bundle and CAS receipt/output were independently verified; receipt `2a272686cfb0df6de90fbe520f923a15c948c7c078ce82e7cc3c1a96be9d4280`.

CPU mode: dl380g10, Python 3.12.11, Torch 2.11 CPU, 4 xdist workers, native threads 1. Two initial environment attempts collected no tests (Sparky's PB-only venv lacked Torch; dl380's PB-only venv lacked compressed-tensors); the project `pq-cpu312` environment produced the red/green above. This repairs an existing fail-closed budget contract; no GPU arithmetic, serving/default change or performance claim. Found during the requested server-cache investigation and isolated in this commit.

### DIFF
--- a/docs/results/server_cache_prewarm_2026-09-07.md
+++ b/docs/results/server_cache_prewarm_2026-09-07.md
@@ -1,0 +1,57 @@
+# Bounded shared-server cache warm-up — 2026-09-07
+
+Rob asked whether known future inputs could be cached on dl380g10 before the
+GPU stage needs them. The existing
+`source_prefetch.prefetch_files_to_page_cache` can perform that anticipatory
+read as an ordinary admitted PB action. It retains no tensors or additional
+cache. `/mnt/shared` is backed by ZFS on this host: the read warms its ARC,
+not a guaranteed pinned allocation. Future eviction and the Spark's own
+NFS/application prefetch remain separate concerns.
+
+At inspection, ARC held about 224 GiB with a configured 240 GiB maximum.
+Linux reported roughly 54 GiB available and zero memory PSI. Reading the whole
+roughly 598 GiB GLM checkpoint would exceed this cache; prospective warming
+should use an explicit next-stage input set and a bounded memory budget.
+There is no implemented PB speculative-cache policy inferred from this run.
+The current pipeline's existing residency and integrity gates remain required.
+
+A single admitted warm-up read the next LFM source checkpoint,
+`/mnt/shared/models/LFM2.5-8B-A1B-BF16/model.safetensors`: **16,936,006,912
+bytes in 6.6848 seconds**. Source file size/mtime were checked before and after.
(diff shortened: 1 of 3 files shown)


RobTand (commented):
Reviewed the diff: an explicit budget of None/<=0 already meant 'derive from available memory', and the derived value is 0 when memory is unknown or under headroom, so `if max_resident_bytes and` was the only thing turning 'no budget' into 'no limit'. Comparing against zero is the fail-closed contract the function already promised; empty files still pass, positive ones skip (auto) or refuse (require) before any read. Red is the `failed` ledger row for `6a577e81d921` (the receipt form for a red run), green is CAS receipt `b7f7c8d4de1f` (67 passed). CI green. Merging per Rob's standing instruction to merge validated PRs aggressively. (Comment rather than approve: same GitHub account as the author.)

PR #376, opened 2026-09-08
Enable selected-source Tessera anchors from canonical capture

### LINKED ISSUE #370: Enable selected-source Tessera anchors from complete canonical capture
Canonical GLM calibration capture can be reused by selected anchor rows, but current `prismaquant/tessera_campaign.py:3698` rejects `--streaming` with `--units`. The alternate path at `:3762` loads the entire source checkpoint onto the device. `tools/dispatch_tessera_campaign.py:_row_memory_gb` charges 623–669 GiB for one actual GLM anchor group using the current 642,652,070,880-byte checkpoint, so these rows cannot execute on a Spark.

P1: the next canonical-capture anchor stage has no executable selected-source path. Extend the existing streaming source/prefetch and calibration-cache owners to make only selected source weights and complete-capture H/X resident before anchors. Derive honest PrismaBuild demand from that same scope and preserve calibration/source/wire identity, per-Linear choices, adaptive fused-group semantics, and fail-closed full-census joint gates. No new cache, dispatcher, sampled-to-joint substitution or source capture is part of this issue.

I can fix this and am doing so in the same working session on a separate branch. Acceptance: regression before implementation; targeted CPU checks through PrismaBuild; tiny native GLM selected-source residency qualification through PrismaBuild; exact source and capture hashes; architecture contract update. This is functional residency work; no speedup is claimed without before/after in-process profiles and both Sparks' Netdata. The full GLM capture remains owned by the coordinating task and must not be disturbed.

RobTand:
A selected Tessera anchor row could reuse a complete capture only after loading the whole model: `--streaming --units` was rejected. This enables that path through the existing resident layer prefetch/cache, copies only selected dense or logical-expert weights, releases source state before H/X prefetch, and binds the full canonical census/capture/runtime/source identity without rerunning calibration.

A separate implementation commit bounds the existing encoder memo to the compatible anchor batch width and derives PB admission from source preparation, export-input serialization and resident encoding. Loader dtype rules preserve strict FP32 bytes. Existing guard and page-release helpers bound completed owners; no new cache, dispatcher, calibration or serving format is introduced.

Validation through PrismaBuild: original CLI and strict-dtype regressions reproduced; native tiny-GLM capture → selected source → producer anchor → resume passed (52 tests). Native memo eviction preserved wire bytes and price rows while retained factor storage fell from 786,432 to 262,144 bytes. Torch CPU/CUDA traces and Netdata from both Sparks are recorded. Final CPU changes passed 66 tests in five shards (two CUDA skips), plus 32 capture/sidecar byte-parity tests. Final focused native rerun after dtype/export-phase edits passed all 60 cases with no skips (PB `61a9da5bbf86`); terminal/CAS success and empty Docker state independently verified. Post-merge architecture/image checks passed 34 cases. See `docs/measurements/selected-source-anchors-2026-09-08.md` for exact actions, environments, negative results and artifact receipts.

The full GLM routed stack still refuses a 104 GiB worker: the current whole-file Hessian writer drives a conservative 124.783 GiB peak. The 132 exact group rows cover 36,423 units, but each 864-unit routed stack is currently indivisible. This PR makes that limit truthful; it does not claim full GLM fit or speed. Within-file bounded writing is separate follow-up work.

Fixes #370.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,23 @@
 As of: 2026-09-08 · `fix/selected-source-anchors`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-08, `fix/selected-source-anchors`) for bounded selected
+encoder-factor ownership (#370). The existing `(unit, scale_plane)` memo in a
+selected-source row retains at most `--anchor-batch-size` entries. Evicted
+factors are recomputed from the same resident uncapped H and producer settings;
+their values, wire bytes, scores and checkpoint identities remain unchanged.
+Admission charges that bounded memo plus a separate factorization transient.
+The existing export-input writer has its own phase: all selected H, X and
+weights remain resident while one complete Hessian sidecar and one tensor
+serialization temporary may coexist. Completed sidecar, rendered-weight and
+wire files receive verified page advice through the existing helper before
+encoding continues. This does not bound the sidecar writer's within-file peak;
+an indivisible full routed stack can therefore remain above admission.
+Selected weight bytes follow the loader's HF dtype plan, including strict FP32
+parameters; the skeleton's initial dtype does not authorize a cast or a budget.
+Resident-source campaign behavior remains unchanged. This establishes an
+ownership bound, not a throughput gain or full-GLM fit claim.
+
 Re-stamped (2026-09-08, `fix/selected-source-anchors`) for selected Tessera
 anchor preparation from a complete canonical capture (#370). `--streaming`
 with `--units` requires the capture manifest and its expected SHA-256. Scope,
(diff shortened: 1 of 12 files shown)


RobTand:
POST-MERGE AUDIT: DEFECTS 1, 2, 3, 4, 5, 6

This audit ran against main 94e30ade7, tree-identical to PR head dbe315c5c. The feature behaves as described: identity binding is real, every degenerate path traced fails closed, and the 124.783 / 98.019 GiB derivation reproduces from `autoscale.selected_anchor_resources` and the census. The defects are hot-path cost, admission and test items for fix-forward on main.

Head receipts at 94e30ade7, priority -10: acebcd9126b4 on sparklina, 60 passed, native GLM selected-source set; 855d162f7940, b4189a0722a3, 9d5cbe865dee, 01085bccf01b, 4b2382913ddd, 83486cb021ce on x86, 7 / 32 / 10 plus 1 skipped / 6 / 13 / 20 passed; 0df8ec2cbc16 on x86 at 16 GiB, 12 passed for `test_glm_campaign_streaming.py`, which OOMs at the 4 GiB default, see 17143615bc45.

1. `prismaquant/tessera_campaign.py:525` and `:3404` call `sha256` with `release_read_pages` and discard the result. The seal is computed in memory before the write; `release_activation_cache_file_pages` already fsyncs and advises the file. On a routed-stack row this is a full SHA-256 pass over every rendered weight and wire file per rung and the 44 GB Hessian sidecar, for no consumer. Delete both calls.

2. `prismaquant/tessera_calibration_cache.py:62` re-hashes the full source root on every invocation. The sealed GLM source is 642.65 GB over 120 safetensors on NFS, so 132 selected rows read about 84.8 TB before any anchor, an estimated 11 to 15 minutes per row at the recorded NFS rate. Seal per-file hashes in the census once and hash only the shards a row installs. No receipt runs a selected row on the real source; this is unmeasured at scale.

3. `tessera_campaign.py:338` sizes the encoder memo to `--anchor-batch-size`, default 1, while `:581` emits pending in rung-major order, so every unit and rung pair refactorizes. Order unit-major or size the memo to members per family, then profile.

4. `prismaquant/autoscale.py:292-346` uses chosen multipliers, 16384 per unit, 4 and 2 times the widest H, and carries no process baseline. The native guard peak was 2.61 GB against a 12.6 MB plan; only declared headroom covered it. Derive the terms and assert guard peak against the plan in the native test.

5. `tools/dispatch_tessera_campaign.py:194` refuses the whole campaign when the widest row does not fit, blocking the 90 admissible rows. Partition and report.

6. No CLI test covers an `s:` stack group, so the page-release path is untested end to end. Receipts 9b3cda16ec66 and 1e0eb77ae756 are unnamed reds; ad8f283959ed and 783031f412d7 ran at priority 0.

PR #379, opened 2026-09-08
Bound Hessian sidecar pages without changing archive bytes

### LINKED ISSUE #377: Bound Hessian sidecar serialization without changing archive bytes
Selected-source anchors still retain every Hessian, X prefix and source weight while `prismaquant/tessera_campaign.py:3383` writes the entire Hessian archive. Completed-file advice at `:3404` happens after all pages exist. The exact GLM routed stack therefore derives 124.783 GiB admission despite a 98.019 GiB resident anchor phase (PR #376 report and PB resource action `8449e462362c`). No full-stack fit has been measured.

P1: this necessary serialization phase prevents the canonical full-stack quantum from being admitted on a 104 GiB worker. I can fix the writer boundary and am doing so in a separate branch in this session, as explicitly requested by the coordinating task. Extend the existing writer/page-release helper to advise stable completed tensor prefixes while preserving exact torch.save archive names, tensor bytes, content/sidecar seals and checks. Do not introduce another Hessian cache or relax canonical completeness.

Acceptance: regression before implementation; exact legacy/bounded full-file byte parity including shared storage, dtype and noncontiguous views; unchanged publication/failure behavior; before/after native writer profiling, cgroup/page-cache sampling and both Sparks' Netdata; truthful phase accounting and targeted CPU/native checks through PrismaBuild. Full GLM qualification requires the complete common capture and remains separately gated.

RobTand:
Selected-source anchor preparation retained a complete Hessian archive in file cache while H/X/source weights remained resident. This pauses Torch's unchanged path writer after synchronous tensor records and uses the existing identity checks, durability fence and page advice on stable file prefixes. Exact archive names/bytes, storage sharing, content seals and publication order remain unchanged; failure preserves the previous published pair.

Matched native PB measurements wrote the same 1,073,748,005-byte archive with identical SHA-256 and content seal. Sampled file-cache peak fell from 1.120 GB to 0.114 GB and cgroup peak from 1.747 GB to 0.810 GB. The profiled writer took 2.707 versus 3.036 seconds, so this establishes reduced file ownership with an observed time cost. Torch CPU/CUDA traces, 20 ms cgroup/I/O samples and both Sparks' Netdata are retained. No speedup or full GLM fit is claimed.

The v2 selected-anchor plan charges a one-record file-page window plus metadata. The representative GLM stack derives 84.345 GiB export residency and 98.019 GiB overall, still requiring actual full-stack qualification from the completed canonical capture. No partial capture, sampled substitute, new cache, dispatcher, format or serving gate is introduced.

Validation through PrismaBuild: regression reproduced before implementation; exact parity across FP32/FP64/BF16, shared/noncontiguous storage and Unicode paths; failed advice/guard preserves publication. Final CPU shards: 63 passed, two skips. Native matched after measurement/capture suite: 34 passed, no skips. Final native GLM capture→selected anchor→resume integration: 20 passed, one already-measured profile-only skip. Terminal statuses, CAS bytes and container cleanup independently verified. See `docs/measurements/bounded-hessian-writer-2026-09-08.md` for commands, receipt paths, negative trials and scope.

Stacked on #376 solely to keep the review diff focused; retarget to main after #376 merges.

Fixes #377.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,23 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-08 · `fix/selected-source-anchors`. Stamps
+As of: 2026-09-08 · `fix/bounded-hessian-sidecar`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-08, `fix/bounded-hessian-sidecar`) for within-file Hessian
+sidecar page ownership (#377). Selected-source export keeps Torch's existing
+path writer, archive record names and serializer, pausing after each tensor
+record to pass its stable visible prefix through the existing inode/size/time
+checks, durability fence and page advice. It changes no tensor, archive byte,
+content seal or publication order; failures retain the previous published pair.
+The v2 selected-anchor resource plan charges one largest H record plus metadata
+as the export file-page window, while all selected H/X/weights and serialization
+scratch remain charged. The physical cgroup guard remains decisive because page
+advice alone is not evidence of release. Resident-source serialization retains
+its previous policy. The bounded writer is byte-qualified against the supported
+Torch runtimes, including shared storage and noncontiguous views; it does not
+establish a full GLM fit or change an exact anchor quantum.
+Gates: `tests/test_bounded_hessian_sidecar.py`, existing capture byte/identity
+checks, and before/after native writer memory traces with both Sparks' Netdata.
 
 Re-stamped (2026-09-08, `fix/selected-source-anchors`) for bounded selected
 encoder-factor ownership (#370). The existing `(unit, scale_plane)` memo in a
--- a/docs/measurements/bounded-hessian-writer-2026-09-08.md
+++ b/docs/measurements/bounded-hessian-writer-2026-09-08.md
@@ -1,0 +1,117 @@
+# Bounded Hessian sidecar writer — 2026-09-08
+
+Issue #377 addresses the remaining whole-file page owner in selected-source
+anchor preparation. The existing writer held all selected source weights,
+Hessians and X prefixes while `torch.save` produced one complete Hessian archive.
+Advice after publication could not bound the within-file peak. The v1 phase
+plan correctly refused a full GLM routed stack at 124.783 GiB.
+
+The selected writer now uses Torch's same path writer and serializer, pausing
+after synchronous tensor-record writes. The existing page helper verifies the
+stable visible file extent, fences durability and advises its pages. Tensor
+storage, pickle protocol, archive record names, CRCs, content seals and the
+atomic publication sequence are unchanged. A file-like substitute was avoided
+because it changes archive names. Resident-source serialization keeps its prior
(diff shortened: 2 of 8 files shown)


RobTand (commented):
VERDICT: MERGE AFTER 1. add a fail-closed check that the record interposition actually observed the payload's tensor records

Reviewed the delta over `origin/main` at head `311ac5fa2ddd` and re-tested at integration head `a0b61d80f45a`. The delta's six code and test files are byte-identical between the two heads, so this applies to both branches.

## The one blocking item

`prismaquant/tessera_campaign.py:3351-3360`. The page advice and the per-record physical guard both live inside the `data/` record-name branch, and nothing after `_save` returns checks that the branch was ever taken.

If a future Torch renames the storage-record prefix or routes storage writes through `write_record_metadata`, `RecordBoundary.__getattr__` forwards the call transparently, the branch never matches, and the writer silently degrades to whole-file residency. The `main` entry point has already granted the optimistic one-record reserve at `tessera_campaign.py:4364-4367`, and the per-record `resource_check` is dead in the same branch. The only remaining physical catch runs after the peak.

Minimal fix: count matched records in `RecordBoundary`, and after `_save` returns refuse when the count is zero while the payload holds tensors. Permit the empty-Hessian case explicitly, since it legitimately writes no `data/` record.

## Verified

Byte parity reproduces at both heads. Both archives are 1,073,748,005 bytes with sha256 `150346bf06ae...`, and the content seal `02c6fcec3dee...` is computed from in-memory tensors before the save.

The accounting term is confirmed. Recomputed from raw `memory_samples`, file pages start at 46,419,968 bytes in every arm, so the incremental window is 1,073,750,016 bytes baseline against 67,117,056 bounded. That is one tensor record, under the charged 67,371,008. The `page_advice` series shows 16 advices, each dropping cgroup from 808 MB to 741 MB.

## Non-blocking

1. `tessera_campaign.py:3417-3433` has no `try/finally`, and Torch finalizes the archive on `__exit__`, so a mid-write refusal orphans a finalized partial `.pt.tmp`.
2. The 2.707 s to 3.036 s writer cost does not reproduce. Two head runs measured 2.703 s and 2.608 s.
3. Advice is whole-file, not a prefix, so the `verified-tensor-record-prefix` policy string understates it.
4. The measurement doc's peak includes a constant 46.4 MB of pre-existing mappings.
5. Eight receipts ran at `--priority 0`: `783031f412d7`, `e259cf8a648a`, `51f662ae5e44`, `c65629a5d7fe`, `86333317d02e`, `b659f35be81a`, `baaae510d954`, `bdff1ff887cd`.

## Head receipts

At parent `311ac5fa2ddd`: `d4a26e648830` sparklina 39 passed, plus five dl380g10 shards totalling 69 passed, 2 skipped.

At parent `a0b61d80f45a`: `14f5f00bbfd6` dl380g10 69 passed, 2 skipped, and `10360eb8f1c8` sparklina GB10 sm121 39 passed.

All `--priority -10`. Both native runs reproduce the 16-advice series and the same window.

PR #382, opened 2026-09-08
Coalesce consumed source spans to release retained file-cache folios

### LINKED ISSUE #380: Coalesce consumed source payloads before page release to bound file-cache retention
Bounded canonical GLM capture failed the physical guard before layer 11. The retired PB cgroup retained 18,513,117,184 file-cache bytes, including 18,111,004,672 file_thp bytes, with anon=0. `prismaquant/layer_streaming.py:1405` aligns each consumed tensor separately, leaving a partial boundary page per tensor even when the next tensor was consumed in the same reader chunk. The suspected large-folio retention needs a bounded native measurement.

P1: deterministic capture stops after growing file-cache charges. Fix coalesces adjacent consumed byte spans before alignment, preserving headers, unread gaps, outer edges, source identity and the existing copy/map lifetime fence. Regression PB f8baf61182b3 fails before implementation. I am fixing it in this session; no full-model retry or fit has been measured by this issue.

RobTand:
Canonical GLM capture retained a large file-cache folio at each adjacent tensor boundary and stopped at the physical guard before layer 11. Merge validated consumed byte spans before page alignment so internal consumed boundary pages are released while headers, unread gaps, outer edges and existing copy/map lifetime checks remain protected.

Real 1 GiB checkpoint ABBA measurement on GB10 reduced retained payload pages from 134,213,632 to 2,093,056 bytes in both paired arms; all GPU tensors remained byte-exact. CPU/CUDA traces, cgroup file_thp and both-host Netdata are retained. No full-capture fit or speedup claim.

Validation through PrismaBuild: pre-fix regression failed; 43 focused source-page/architecture/staleness tests passed, zero skips (0c3924531a25); native measurement2f7d0b024d3c exited0 with complete cleanup. Source-page tests include preserved unread gaps and invalid-span refusal before any advice. Full evidence: docs/measurements/coalesced-source-page-release-2026-09-08.md.

Fixes #380

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,18 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-08 · `fix/joint-boundary-residency`. Stamps
+As of: 2026-09-08 · `fix/coalesced-source-page-release`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-08, `fix/coalesced-source-page-release`) for consumed
+source-page release (#380). The existing CUDA reader chunk validates selected
+safetensors spans, merges adjacent or overlapping consumed bytes, then advises
+only complete pages of each union. This releases internal tensor-boundary
+pages while preserving headers, unread gaps and outer partial pages. The
+existing completed-copy fence, mapping lifetime and source-identity checks
+remain unchanged. No global page drop, admission allowance or source cache is
+introduced; advice remains best effort. Gates cover adjacency, unread gaps,
+invalid-span refusal, CUDA lifecycle and a bounded real-checkpoint native
+profile. A full canonical capture still requires its own successful receipt.
 
 Re-stamped (2026-09-08, `fix/joint-boundary-residency`) for the explicit
 `prismaquant.aura.boundary_storage.v1` policy (#373). Streamed AURA can store
--- a/docs/measurements/coalesced-source-page-release-2026-09-08.md
+++ b/docs/measurements/coalesced-source-page-release-2026-09-08.md
@@ -1,0 +1,81 @@
+# Consumed source-page coalescing — 2026-09-08
+
+Issue #380. This fixes the opt-in CUDA source-reader page-advice range, without
+changing checkpoint bytes, selected tensors, copy/map lifetime fences, source
+identity, admission thresholds or the canonical calibration contract.
+
+The common GLM capture stopped at its physical memory guard before layer 11.
+Its retired cgroup retained 18,513,117,184 file-cache bytes, including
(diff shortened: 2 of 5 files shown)


RobTand (commented):
VERDICT: MERGE

The span algebra is correct and the measurement is real. Merge, then fix the docstring forward.

Head coverage. Re-tested at a0b61d80f, the integration head that ships, because
`prismaquant/layer_streaming.py`, `tests/test_streaming_source_pages.py`,
`experiments/source_page_release_profile.py` and the measurement doc are byte-identical to
e22a0286a by blob SHA. Three shards at priority -10 on dl380g10: c2e278c510d2 with 47 passed,
bb501173001e with 43 passed, 489d605f458f with 18 passed. 108 passed, 0 skipped. Each sealed tree
equals a0b61d80f apart from its own closure marker. `test_architecture_doc` and
`test_docs_staleness` are covered in shards bb50 and 489d.

Correctness. Validation of every span now precedes any advice, so
`test_invalid_later_span_refuses_before_any_advice` describes real new behavior. Merging is the
exact union of consumed spans, an unread gap blocks the merge, both outer edges align inward, and
header bytes are unreachable because `align_up` never falls below `base`. `sorted(set(keys))`
makes the result independent of caller order and duplicates. The copy fence is unchanged: advice
runs after the CUDA event synchronize, after `host_staging.clear()`, and only when every output is
CUDA resident. No consumer holds a view into the released range.

Measurement. From `native-abba-01/measurement.json` under 2f7d0b024d3c on sparky, not from the
doc. Both per_key arms end at 134,213,632 resident payload bytes and 136,314,880 `file_thp`. Both
coalesced arms end at 2,093,056 and 4,194,304. `exact_gpu_bytes` is true in all four arms, and all
four peak above 1.04 GB of `file`, so no arm short-reads. The arithmetic closes exactly: 64 tensors
of 16 MiB from an offset that is 2824 bytes past a page boundary give 63 mid-page internal
boundaries, and the delta 132,120,576 bytes is 63 times 2 MiB, one folio each. That qualifies the
folio mechanism issue #380 called suspected. The failure it targets is also receipted now:
63992d4c42ad records `anon 0`, `file 18513117184`, `file_thp 18111004672`.

Fix forward, none blocking.

1. `prismaquant/layer_streaming.py:1377`. The docstring still says shared partial edge pages are
   never included. Internal pages shared by two consumed tensors are now exactly what is advised.
   Suggest: exclude the outer partial pages of each consumed union.
2. The 2026-09-07 ARCHITECTURE stamp's edge-page sentence is superseded. One clause in the new
   stamp would settle it.
3. PR body: "native measurement2f7d0b024d3c exited0" is missing two spaces.
4. f8baf61182b3 was submitted at priority 0. The live agent band is -10.

PR #384, opened 2026-09-08
Bound research candidate residency through ProductionWeightCache windows

### LINKED ISSUE #381: Bound research candidate residency through existing ProductionWeightCache windows
The research joint-operator integration needs finite candidate windows through ProductionWeightCache. Existing prefetch submits the whole key list to pool.map, and get silently reloads an evicted entry. There is no resident-only consumption boundary or complete backing-storage budget for view inputs.

Add an explicit window planner/context, strict resident lookup, and selected-key release through the existing cache. Account all existing tensor storages plus incoming standard uncompressed Torch shards; bound the roster and loader workers, reject opaque inputs and nested windows, preserve CB/file-load receipts, and optionally release verified file pages using the existing helper. Legacy defaults remain unchanged. Validate lifecycle, aliases/views, failure cleanup, receipts and refusal paths through PrismaBuild; no performance or full-model fit claim.

RobTand:
ProductionWeightCache previously offered whole-list prefetch and lazy `get`, so a research candidate phase could retain unbounded loader results or silently reload an evicted tensor. Add explicit finite resident windows through the existing cache: deterministic alias-resolved planning, strict resident lookup, and selected-key release on context success or failure.

The resident cap charges complete backing storages across every existing cache tensor plus incoming entries; aliases count once. Standard uncompressed Torch archive preflight bounds incoming storage, serialized loader buffers have a separate aggregate cap, nested windows and LRU eviction of another owner refuse, and optional checked file-page advice follows exact-byte loading. CB validation and file-load receipts remain active. Legacy defaults are unchanged. Callers still own borrowed references and must separately admit device copies, deltas and consumer workspaces; this is not a full-model fit or throughput qualification.

Validation through PrismaBuild on DL380G10: seven expected missing-API failures before implementation; final **108 passed, zero skips**, including existing cache/receipt and architecture checks; both touched Python files compiled. Four canonical CAS receipts/payloads and all five terminal records were independently audited. Final test and compile source snapshots match implementation `d2a9218f08` across every tracked path, with only generated PB closure additions. Commands, ownership limits and receipt locations are recorded in `docs/measurements/pwc-resident-windows-2026-09-08.md`.

Closes #381

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,26 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-08 · `fix/joint-boundary-residency`. Stamps
+As of: 2026-09-08 · `codex/pwc-resident-windows`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-08, `codex/pwc-resident-windows`) for opt-in candidate
+windows through `ProductionWeightCache` (#381). `plan_resident_windows`
+resolves aliases in input order and bounds each finite quantum by a declared
+byte cap and loader count. `resident_window` uses the existing prefetch pool;
+`get_resident` refuses missing or evicted entries and preserves CB integrity
+and active file-load receipts. Whole backing storages count, including views,
+aliases and unrelated resident entries. Incoming standard uncompressed Torch
+archives have a conservative file-size bound plus storage-record preflight;
+opaque/compressed inputs refuse. Serialized load buffers have a separate
+aggregate cap. Nested windows and LRU eviction of other entries refuse.
+Context completion or failure releases the selected disk-backed cache owners
+through the existing release mechanism. Optional checked file-page advice
+follows the verified load, without changing artifact bytes or promising
+physical reclaim. Callers still admit allocator overhead, borrowed references
+and projection workspaces separately. Legacy cache defaults, pipeline order,
+quantization arithmetic and serving gates are unchanged. CPU lifecycle and
+budget gates are in `tests/test_pwc_resident_windows.py`; this API is not a
+full-model fit or throughput qualification.
 
 Re-stamped (2026-09-08, `fix/joint-boundary-residency`) for the explicit
 `prismaquant.aura.boundary_storage.v1` policy (#373). Streamed AURA can store
--- a/docs/measurements/pwc-resident-windows-2026-09-08.md
+++ b/docs/measurements/pwc-resident-windows-2026-09-08.md
@@ -1,0 +1,93 @@
+# Production weight-cache resident windows — 2026-09-08
+
+Issue #381; implementation `d2a9218f08`. This is an opt-in cache API and CPU
+lifecycle qualification. It changes no pipeline default, source traversal,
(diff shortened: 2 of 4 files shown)


RobTand (commented):
VERDICT: MERGE AFTER (1) release only the entries the window itself loaded

Reviewed at #384 head `470ba30f2aa9` and at integration head `a0b61d80f45a`. The three files this PR touches outside `docs/ARCHITECTURE.md` are byte-identical between the two heads, so this applies to either PR. My head receipts were taken on the integration tree.

## Blocking, one token

`production_weight_cache.py:581-582` releases the whole roster in its `finally`, including entries that were already resident when the window opened. Those entries are excluded from `loaded` at :568 and from `_resident_window_receipt_keys` at :565, so the window never acquired them and does not guard them. The docstring at :528 states the intended rule and the code does not match it: "selected disk-backed entries are released". A pre-resident tensor is not disk-backed at window open.

Measured in probe action `5b2ed3b67d2d`: a key resolved before the window reports `loaded: 0` and `receipt_keys: frozenset()`, and `cache.weights[key]` is a filename again after the window.

Failure scenario. An owner arranges bounded residency for unit `a`. A consumer opens a window over a roster that includes `a`. On exit, :385-404 demotes `a` to disk and drops it from `_lru_order`. Values stay correct; the owner's next read goes back to NVMe.

Fix, using a set already built at :565:

    self.release_resident_tensors(self._resident_window_receipt_keys)

Please add a test with one already-resident key in the roster. `test_existing_lru_owner_cannot_be_evicted_by_window:137` and `test_selected_release_preserves_unrelated_residents_and_receipts:79` both use a disjoint resident key, so neither reaches this path.

## Non-blocking

2. `plan_resident_windows:514` caps a quantum at `max_workers`, which `_window_limits:473` caps at CPU affinity, and `resident_window:545` refuses rather than iterates a larger roster. Documented, but the refusal names neither cap.
3. A deleted backing file raises a raw `FileNotFoundError` from `_window_file:481`, outside the module's `RuntimeError` taxonomy.
4. One meta or sparse entry anywhere in `weights` stops all planning at `_window_resident_storages:427`, with no key in the message.

## Receipts

`060a44dbc76a` dl380g10 parent `a0b61d80f45a` rc0, 108 passed, zero skips, including `test_docs_staleness` and `test_architecture_doc`.
`3257063febdd` dl380g10 parent `a0b61d80f45a` rc0, 140 passed, 4 pre-existing GPU-or-bust skips across 15 cache and pipeline suites.
`741e6f37051e` sparky sm_121 parent `a0b61d80f45a` rc0, 24 passed.

Body claims 1 through 5 all verified, including the sealed-snapshot claim by bundle diff against `d2a9218f08`.

RobTand:
Withdrawing the MERGE AFTER item from my review. The window docstring states that selected disk-backed entries are released on success and on failure, and `test_existing_disk_loaded_tensor_can_enter_without_enabling_receipts` asserts that contract for a tensor resident before the window opened. Releasing only `_resident_window_receipt_keys` would have been a contract change, not a defect fix. The disk-backed copy is re-readable, so no data is at risk; the cost is one re-read for a caller that relied on a pre-window residency. No code owed. The other non-blocking notes stand.

PR #387, opened 2026-09-08
Plan bounded joint-statistics windows using shared activation requirements

### LINKED ISSUE #385: Plan bounded joint-statistics target windows from shared activation grouping
The operator-statistics lease bounds candidate quanta but still requires all target GW/GA matrices for a layer at once. Actual GLM layers have867 targets and27.09375GiB per FP32 operator plane; the readable menu contains static A4 and dynamic A8 groups, so three planes require81.28125GiB before source residency.

Add a deterministic whole-target window planner using the lease's own extracted grouping/requirements. Plans retain names and scalar identity metadata only. Reject invalid coverage, aliases, unsupported backends and oversized individual targets before hooks or matrix allocation. Root integrates bounded replay separately. This change does not alter signed arithmetic, probe selection or serving gates.

Implementation and PB CPU validation are owned in this session; actual GPU/full-model fit remains unmeasured.

RobTand:
A full GLM layer can require 81.28125 GiB of GW/GA statistics before its source weights or activations. Add deterministic whole-target windows under an explicit statistics cap, deriving activation grouping and matrix bytes from the same helper consumed by both existing projection leases. Plans retain only names and immutable identity metadata; invalid coverage, aliases, static contracts, backends and oversized targets refuse before hooks/allocation.

The planner preserves lease group insertion order and signed arithmetic. It performs no source replay or full-model admission. CPU geometry validation for 867 actual GLM-shaped targets yields 341/341/185 target windows under 32 GiB; existing dense, packed, streamed, microbatch and operator projection tests pass unchanged.

PrismaBuild validation: 146 passed, 2 explicit native-only skips across 4 shards; compile checks passed. Actual terminal/source/CAS audits and commands are recorded under joint-statistics-target-windows-01, with the dated report in docs/measurements/joint-statistics-target-windows-2026-09-08.md.

GitHub CI on the frozen post-merge head: 6680 passed, 200 skipped, 3 xfailed, and 192 subtests passed; import surface and linked-issue policy also passed. Run: https://github.com/RobTand/prismaquant/actions/runs/34224069522.

Runtime integration is separate from this planner.

Fixes #385

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,21 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-08 · `research/joint-operator-statistics`. Stamps
+As of: 2026-09-08 · `feat/joint-statistics-target-windows`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-08, `feat/joint-statistics-target-windows`) for whole-target
+operator-statistics planning (#385). The planner and both existing joint
+projection leases share one target-validation and activation-group requirement
+calculation. A target charges one FP32 GW matrix plus one FP32 GA matrix per
+distinct executed nonidentity QDQ group. Sorted target names form deterministic
+whole-target windows under an explicit statistics cap; oversize targets refuse
+before hooks or allocation. Plans retain names, shapes, ordered format groups
+and activation receipts, never source tensors, specs or callable pointers.
+Dynamic callable distinctions remain separate groups, while static served
+contracts retain their shared QDQ owner. Existing lease insertion order and
+projection arithmetic remain unchanged. The planner does not execute replay or
+admit source, boundary, graph, candidate or allocator memory. Gates:
+`tests/test_joint_statistics_plan.py` plus existing joint lease/projection tests.
 
 Re-stamped (2026-09-08, `research/joint-operator-statistics`) for the explicit
 one-probe `JointOperatorStatisticsLease` (issue #374). It extends the existing
--- a/docs/measurements/joint-statistics-target-windows-2026-09-08.md
+++ b/docs/measurements/joint-statistics-target-windows-2026-09-08.md
@@ -1,0 +1,52 @@
+# Whole-target joint-statistics planner — 2026-09-08
+
+Issue #385. `plan_joint_statistics_target_windows` uses the same extracted
+module validation, activation grouping and FP32 matrix requirements as both
+joint projection leases. It emits immutable target-name windows under an
+explicit statistics cap. It neither installs observers nor executes replay.
+Source/cotangent/graph/PWC/transient admission remains a separate obligation.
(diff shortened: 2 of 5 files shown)


RobTand (commented):
VERDICT: MERGE

Reviewed delta commit 14c9e7063 only. Every PR receipt sealed a tree short of the head: the five done rows omit the measurement doc, and 0d476f86d1a108deca619c386b3e2af30b24992310406887fd3ce6ddb1e2ecda also omits the ARCHITECTURE stamp and runs an older test file. Its one failure is that older test expecting ValueError where ActivationScaleContractError is raised. The head test expects the right type. Not a regression.

Head receipts at 14c9e7063, all priority -10, sealed tree equal to the head:

- e5e6d709e6df8f1765b9bf93e1e8d6cded53249ed2f3455ea33bba4777c405e7 on dl380g10: 68 passed across the plan, docs staleness, projection and streamed tests.
- 831ff9a1639e42afe052adfb22ca24fd7934440c9776e09e1707ec1cd91a72e8 on dl380g10: 48 passed, 2 skipped across operator statistics, architecture doc, packed and microbatch tests.
- 05b021f60c3ac242d9e9a275472dc0260fcac7f4dceaa71dcb4408e7772a4a87 on sparky GB10: 71 passed on CUDA with torch 2.11.0+cu130.

Claims verified. 867 targets times 8,388,608 elements times 4 bytes is 29,091,692,544 bytes, or 27.09375 GiB per plane, and the three planes total 81.28125 GiB. The shapes match the GLM-5.3-Flash config: 288 routed plus 1 shared expert at moe intermediate 2048, hidden 4096, and three dense layers at intermediate 12288. The byte formula at joint_aura.py:272 equals what the lease allocates: one FP32 GW plane at weight shape plus one FP32 GA plane per group where act_quant_changes_input holds. Windows hold whole targets, so no group straddles a window and per target signed sums stay independent.

Non-blocking findings for integration:

1. The cap bounds resident planes only. Each accumulate materialises a fresh product before add_ at joint_aura.py:591-597, observe holds FP32 copies of x, g and dx, and seal takes weight.float at :674. On GLM the product transient is 32 MiB per routed target and 192 MiB per dense target above the cap, plus the FP32 activation copies, which scale with the microbatch. The caller must subtract both from the admitted reservation.
2. The cap is caller-declared and no constant lives in code, which satisfies principle 2. The pipeline must derive it from the PB admission or MemAvailable rather than type 32 GiB.
3. Windows assume every replay sees the same calibration draw and source state. identity_sha256 does not bind the probe identity. Stamp the calibration hash and set PRISMAQUANT_DETERMINISTIC per window before the first production use.
4. The shared helper tightens SignedJointProjectionLease with strict coverage and FormatSpec checks. All five callers, including the replay driver behind the production reducer experiment, pass matched key sets. Nothing reachable changes.
5. JointOperatorStatisticsLease accepts a zero cap while the planner requires a positive one. Harmless.

PR #400, opened 2026-09-08
Capture exact AURA boundaries once per source layer

### LINKED ISSUE #394: Amortize streamed AURA baseline source loads with exact layer-major capture
The exact boundary owner bounds retained tensor storage, but initial streamed AURA capture still traverses every source layer for each complete B1 calibration batch (`prismaquant/aura_cost.py` initial capture block and `StreamedCausalLM.capture_boundaries`). With a bounded source cache, the same large GLM layer payload is reread across the full draw.

Extend the existing `visit_layer_batches` traversal with checked exact boundary windows so each source layer is installed once and processes original batches in order. Keep source kwargs, per-batch shared state, shapes, probe coordinates and exact tensor bytes. The mode must remain explicit/default-off and use the existing activation/source owners. Include a quiescent deep fork for `SharedStateCotangents` so separate bounded target replay can consume disposable adjoints without changing the original reverse-pass owner; root owns reverse integration.

Acceptance: regression before implementation showing legacy source install/load count, exact per-batch boundary and cotangent parity including shared-state fixtures, missing/corrupt/state/budget/refusal/cleanup tests, and paired native genuine original-layout GLM source-load measurements with in-process profiling plus Netdata on both hosts. All non-vLLM execution uses PB with bounded resources; native launch requires coordination with current control/capture work. No full-model fit, new format or serving claim.

RobTand:
Closes #394.

Initial streamed AURA capture rereads source layers for every original calibration batch. Explicit boundary policy v2 (`capture_order: layer_major`) now extends the existing layer visitor: one source install per layer, original complete batches, verified exact input windows and immediate output artifacts. V1/default capture remains unchanged. Per-batch source state, probe coordinates, reverse traversal and scalar projection arithmetic are preserved. Evaluation mode, observed Torch RNG checks and auxiliary budgets guard the new mode.

The separate shared-cotangent commit adds quiescent bounded deep forks for disposable replay; reverse replay integration remains separate. Native validation also caught and fixed the existing visitor order check comparing original CPU IDs with prepared CUDA IDs. Architecture and runtime allocation docs describe the contracts.

Validation through PrismaBuild:
- The before-change regression failed with 10 source installs versus expected 2.
- Broader CPU gate: 122 passed, no skips, plus four module compile checks. Final targeted gate: 45 passed, one CUDA-only skip; the native rerun exercises that actual CUDA path.
- Genuine three-layer original-layout GLM ABBA with two source slots: source reads/installs 15, 3, 3, 15, an 80% reduction in capture source reads. All 33 target cost/statistic panels, 80 cotangents and per-batch source tensor hashes match exactly. Independent PB reads verified the four saved cost panels.
- Exact boundary residency stays capped at 43,520 bytes. V2 adds 130,560 logical activation input bytes during capture. All 100 artifact entries retire and hot-read misses remain zero.
- Paired capture Torch CPU/CUDA profiles, full-call cProfile, process I/O, both-host Netdata and audited CAS receipts are retained. Native scope peak was 3,069,296,640 bytes under an 8 GiB physical reservation; exit and cleanup verified.

[Full commands, measurements, failures and limitations](docs/measurements/layer-major-boundary-capture-2026-09-08.md). Native action `63dffca2d164ad8579af3f5859da734aa8e83a165281bd08586ba5abb2774190`; evidence under `/mnt/shared/tessera-measurements/glm-canonical-census-20260908/layer-major-boundary-native-02/` and `layer-major-boundary-implementation-01/`.

This remains opt-in. The small instrumented fixture establishes source-load reduction and exact semantics, with no full-model fit, production throughput, serving-format, work-per-joule or fused-kernel claim.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,6 +1,24 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-08 · `integration/glm-probe-memory`. Stamps
+Re-stamped (2026-09-08, `fix/layer-major-boundary-capture`) for explicit
+`prismaquant.aura.boundary_storage.v2` with `capture_order: "layer_major"`
+(#394). The existing layer visitor installs each baseline source layer once
+and dispatches the original complete batches in order. It keeps each batch's
+original source kwargs/pass state, leases exact input windows from the existing
+activation owner and writes the original output boundary before advancing.
+No full-draw hidden tensor plane is added. Metadata and all live forward/shared
+states remain under the auxiliary cap; physical admission remains separate.
+V1 and default callers preserve batch-major capture. The reverse/probe order,
+source arithmetic, RNG coordinates of Fisher probes, candidate formats and
+serving gates are unchanged. The new capture requires evaluation mode and
+refuses observed Torch CPU or runner-device CUDA RNG consumption during
+preparation/source calls. Other mutable custom-model state is outside this
+qualification; no blanket stateful-model equivalence is claimed. Gates include
+`tests/test_layer_major_boundary_capture.py`, actual shared-state regressions,
+and a paired original-layout GLM source-read/profile qualification with Netdata
+from both hosts. This remains default-off and establishes no full-GLM fit.
+
+As of: 2026-09-08 · `fix/layer-major-boundary-capture`. Stamps
 follow, newest first, each recording its own branch and date.
 
 Re-stamped (2026-09-08, `integration/glm-probe-memory`) for the combined
--- a/docs/design/joint_aura_runtime_allocation.md
+++ b/docs/design/joint_aura_runtime_allocation.md
@@ -280,6 +280,15 @@ as full-draw quality evidence.
 
 ## Exact boundary working storage (opt-in)
 
+For bounded target replay, `SharedStateCotangents.fork_for_replay` accepts a
+mandatory `max_resident_bytes` cap for the fork's newly allocated compact
+adjoint tensors. It requires a quiescent owner after harvest, deep-copies all
+accumulators, and keeps diagnostic lists independent. The original adjoints
+remain separately charged. Earlier target windows may consume/harvest one
+batch's disposable fork and release it; only the final target window may use
+the original owner and commit outgoing cotangents. This helper does not itself
+schedule replay or change reverse-pass behavior.
+
 `compute_aura_cost_streamed(..., boundary_storage=policy)`, CLI
 `--boundary-storage-config policy.json`, or the Tessera joint plan's
 `execution.boundary_storage` accepts the closed policy:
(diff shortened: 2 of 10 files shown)


RobTand (commented):
VERDICT: MERGE AFTER. Reviewed at head `b99567cc32ce`, begun as an early read of the draft at `cf1271fc3d`.

1. Rebase for the `docs/ARCHITECTURE.md` stamp conflict. The branch was mergeable at review start against `695439cb37`; #387 then landed as `f5dd07a4eb15`, prepending its own stamp above the same `As of:` line. GitHub reports `mergeStateStatus: DIRTY`. One conflict hunk, one file, no semantic overlap: main's `prismaquant/joint_aura.py` changes reference none of `visit_layer_batches`, `capture_layer_major_boundaries`, `boundary_storage` or `fork_for_replay`. Keep your stamp on top and #387's below it.

I built that merge at integration head `23c502e7cea7`. Green: `d590261694b2`, `544dc54d0b44` and siblings on dl380g10, all rc 0, 908 passed, 12 skipped, 0 failed, including main's `tests/test_joint_statistics_plan.py`. One shard, `93330b39be01`, ended rc 137 on `memory_limit_oom` at my own 16 GB reservation and passed at 32 GB. Those receipts cover my resolution; if yours differs, re-run the two docs tests at the pushed head.

Exactness holds on real tensors, not mocks. `tests/test_layer_major_boundary_capture.py:60` runs `compute_aura_cost_streamed` twice through the real visitor and compares the `costs` dict, every source-call input and output digest, every boundary and cotangent digest, `h_trace`, `fisher_col` and `probe_identity`; `:95` proves the order genuinely changed. Per-batch state is held at `prismaquant/cost_streaming.py:754`. The RNG gate at `:723-734` observes CPU and runner-device CUDA state around `_prepare`, `_call`, both pass-state calls and `tail_logits`, and `:718` refuses training mode. Both gates bite, per `:107` and `:125` of the same test file.

`63dffca2d164`, sparklina, rc 0, seals a tree equal to `02ef6a5630`, and the only delta to head is the measurement doc. `9bf0eecaada7`, rc 1, is the before-change regression; its `install_calls` counter increments inside the driver's own context, not a fixture stand-in. My head runs: CPU `251ffb1a4b33` plus 19 shards, rc 0, 695 passed; GPU `cc9d3085916e` rc 0, 18 passed with no skip, so `test_original_cpu_tokens_compare_with_prepared_cuda_tokens` ran on real CUDA.

Every body number reproduces: reads and installs 15, 3, 3, 15; capture wall 1.218376, 0.362056, 0.369237, 0.579450; 33 targets and 80 cotangents with `exact_parity` true; resident cap 43,520, auxiliary peak 4,335; read bytes 1,218,560 against 1,088,000; 100 entries written and retired, no live bytes, no hot-read misses; scope peak 3,069,296,640 under 8 GiB; six Netdata samples per host.

Two non-blocking notes. `prismaquant/cost_streaming.py:792` also changes the default path, since a v1 caller with host-resident tokens previously raised; "V1/default capture remains unchanged" needs a qualifying clause. `prismaquant/cost_streaming.py:742-746` never clears `scheduled`, so a layer evicted between prefetch and install cannot be re-scheduled and `:777` refuses; fail-closed, worth a follow-up issue.

PR #404, opened 2026-09-08
Qualify planned-size joint statistics storage and turnover

RobTand:
The bounded joint probe had numerical evidence only on tiny GLM layers. Add native qualification helpers and receipts for allocator turnover and two operator-statistics windows on full GLM layer-three projection shapes. Each window retained 1,023 FP32 matrices (31.96875 GiB) alongside 31.0277 GiB of source geometry, under the unchanged 104 GiB physical / 92 GiB GPU limits. Existing runtime code and defaults are unchanged.

Validation and scope:
- Allocator proxy: real retired 32 GiB storage released while the live source proxy retained its storage identity; future 48 GiB remained a reservation.
- Actual statistics: two 341-target windows, 65,472 independently checked FP64 samples per window, 682 nonzero projections per window, 341 exact cotangent checks, and all matrix backing owners expired. Second admission reclaimed 31.992 GiB of reservation; all 869 final fixture backings expired. Peak CUDA reservation was 63.229 GiB.
- Eight statistics-helper CPU tests and eleven allocator-helper tests passed, with no skips; the separate actual-census plan passed. Final compile checks for all three helpers and 19 architecture/staleness checks also passed through PB. Native runs, profiler summaries, input hashes, source snapshots, CAS receipts and cleanup were independently audited.
- Original weights, packed routing, PWC candidate replay, the full calibration draw, quality and serving remain separate gates. The eight-row shape fixture is not a throughput comparison. Remaining allocation after explicit fixture owners expire is reported without assigning it to an unobserved owner.

The architecture document now records this precise qualification scope. A separate evidence commit preserves six existing full-capture profiler windows, both-host Netdata and GPU-energy observations; it adds no GPU replay or before/after speed claim.

Evidence: `experiments/measurements/glm-joint-statistics-scale-20260908/README.md`, `glm-joint-allocator-turnover-20260908/README.md`, and `glm-original-capture-baseline-20260908/README.md`.

Follow-up qualification for the implementation merged in #402.

Closes #392.

PR #393, opened 2026-09-08
Release completed anchor and sidecar pages without hashing the bytes again

RobTand:
Closes #395. Refs #370. Finding 1 of the PR #376 post-merge audit (https://github.com/RobTand/prismaquant/pull/376#issuecomment-5584543796).

`_finish_anchor` and `write_export_inputs` hashed every completed file a second time before advising its pages out, and discarded the digest. The seal is computed from the in-memory bytes before the write, and `release_activation_cache_file_pages` checks the entry's identity, fsyncs and advises it from its own descriptor, so the extra pass had no consumer. On a routed GLM stack it was a full read of the 44 GB Hessian sidecar per row and of every rendered shard and wire per rung, plus one guard poll per 16 MiB read. Archive names, bytes, seals and the release itself are unchanged; the only behavioural change is that nothing reads a completed entry back.

Two regressions demonstrate it: `test_export_input_page_release_does_not_reread_the_archive` counts hash calls on the capture path with `release_file_pages=True`, and `test_completed_anchor_page_release_does_not_reread_the_entries` drives `_finish_anchor` with `release_completed_anchor_file_pages` set and a hash helper that refuses any read. The `resource_check` labels `before_capture_hash` and `after_capture_hash` no longer appear on the export-input path.

PrismaBuild, x86 (`/home/rob/venvs/pq-cpu312`), priority -10:

- Red at the test-only commit 1efbc09f4: `49801762d3c7953ec6ecfe53ceb2cafdebc94a8d96ab8ea090fa54d802ee1181`, rc 1, exactly the two new tests fail, 32 pass.
- Green at 6923fabf3, four shards: `62837a808dc5fb8fbb56f9f6adfb2bf536b4802fa09cb6b7cdb33fa6fbe444e9`, `d1d75c3ea8709771934f2e590e791e86d9543c0362f8bd8ac367dda7fb62d045`, `75195ff81c4bf63a3327d0c586fa848b13b4e76084cf41b47b214147bc48654e`, `04af35f60d7902acab23f3b2a36436e546c88c3e5e82732031d2840dbd923a6a`; `test_tessera_calibration_cache.py` 34 passed, `test_tessera_selected_source.py` 10 passed 1 skipped, `test_tessera_materialization.py` 25 passed 1 skipped, `test_allocator_tessera_priced_inputs.py` 5 passed.
[...]

Issue #390, opened 2026-09-08
autoscale.selected_anchor_resources uses chosen multipliers with no process baseline; native guard peak 2.61 GB against a 12.6 MB plan


RobTand:
Post-merge audit of PR #376 at main 94e30ade7, finding 4 (https://github.com/RobTand/prismaquant/pull/376#issuecomment-5584543796).

`prismaquant/autoscale.py:292-346` derives the selected-anchor admission plan from chosen multipliers: 16384 bytes per unit, 4 and 2 times the widest Hessian, and no term for the process baseline (interpreter, torch, CUDA context, loader buffers). On the native fixture the guard observed a 2.61 GB peak against a 12.6 MB plan; only the declared headroom covered the gap. Principle 2: derive every term from the objective or a dtype's precision, never from a chosen constant.

Expected: derive each term from the scope the code actually makes resident (the same scope `_row_memory_gb` charges), add a measured process baseline, and assert guard peak against the plan in the native test so a plan that undercharges fails the test rather than relying on headroom.

PR #406, opened 2026-09-08
Integrate capture page-release and writer safeguards

### LINKED ISSUE #395: Tessera anchor completion and export-input write re-hash the bytes they just wrote
Post-merge audit finding 1 on #376, split out so the fix can carry a closing reference.

**Where.** `prismaquant/tessera_campaign.py`, `_finish_anchor` and `write_export_inputs`. Both call `tessera_calibration_cache.sha256(path, release_read_pages=True)` on a file they have just written, then call `release_activation_cache_file_pages` on the same file.

**What is wrong.** The digest returned by `sha256` is discarded at both sites. `write_export_inputs` already sealed `capture_sha256` from the in-memory payload before writing. So every completed anchor re-reads its rendered shard and wire blob in full, and every selected row re-reads the Hessian capture in full, with no consumer of the bytes. The page release the read was meant to enable is performed by the helper that follows it anyway.

**Cost.** One full read of every completed anchor's output plus one full read of the Hessian capture per selected row, on the export hot path. Principle 7: CPU and disk pressure on a hot path is a bug.

**Fix.** Drop the `sha256` calls and let `release_activation_cache_file_pages` check identity, fsync and advise from its own descriptor. Regression tests must assert the release path does not re-read the archive. PR #393 carries the fix.

### LINKED ISSUE #396: Hessian capture page release is fail-open when torch writes no data/ record
Review finding on #379, merged through #386 without the requested check.

**Where.** `prismaquant/tessera_campaign.py`, `_save_hessian_capture_with_page_release`. `RecordBoundary.write_record` advises the archive pages only for records whose name starts with `data/`.

**What is wrong.** Nothing checks that at least one record matched. If a torch release renames its storage records, or the archive layout changes so tensor bytes land under another prefix, the wrapper silently degrades to plain `torch.save` and the sidecar's pages stay resident. The measurement in #379 was made on the layout of the pinned torch; a later pin can retire the mechanism with no signal. That is the same class as a route that nothing consumes.

**Fix.** Count matched `data/` records. After `serialization._save` returns, if the payload holds any tensor and no record matched, raise. The write happens on a `.pt.tmp` path, so the raise must also remove that temporary file instead of orphaning it beside the previous capture. A regression test drives the writer through a serializer that emits records under a different prefix and asserts the refusal and the cleanup.

### LINKED ISSUE #398: test_selected_capture_cli_reaches_existing_streamed_source fails on CUDA hosts: bounded capture env gate fires inside the test
Found while re-testing #393 on sparklina through PrismaBuild. A control at main `695439cb37` fails the same way (failed row 5d53e29e3fa8, sparklina, priority -10), so this is on main, not in #393.

**Where.** `prismaquant/tessera_campaign.py` `main`, the block that runs `require_bounded_capture_environment(os.environ)` when the capture policy is bounded or a selected source is requested and the device is CUDA. The gate needs `PRISMAQUANT_RELEASE_SOURCE_PAGES=1` and `MIMALLOC_PURGE_DELAY=0` in the process environment.

**What happens.** `tests/test_tessera_selected_source.py::test_selected_capture_cli_reaches_existing_streamed_source` calls `campaign.main` with a selected source and never sets those variables. On a CPU host the device check skips the gate and the test passes. On a GPU host the gate raises before the CLI reaches `build_streamed_causal_lm`, which is what the test is checking:

```
RuntimeError: bounded CUDA capture requires PRISMAQUANT_RELEASE_SOURCE_PAGES=1 before process startup
```

**Fix.** The test sets the gate's environment through `monkeypatch.setenv` from `autoscale.BOUNDED_CAPTURE_ENV`, so it exercises the same path on both host classes. The dispatcher already sets these for real runs, which `test_selected_capture_cli_reaches_existing_streamed_source`'s sibling asserts at line 135.

RobTand:
Completed anchor and Hessian-sidecar page release currently reads each file again to compute a discarded hash. This integration removes that pass while preserving descriptor identity checks, fsync and page advice. It also refuses publication if a tensor-bearing sidecar produces no stable tensor records, cleans up the failed temporary payload, and fixes the selected-source test's CUDA environment setup.

Includes the reviewed ancestry of #393, #397 and #399. Successful artifact bytes, content seals and production defaults remain unchanged. Architecture documentation records the final writer contract. Separate prose corrections remove an unsupported prior-seal claim and distinguish derived dispatcher memory bounds from measured peaks.

Validation: 154 passed, four skipped on CPU through PrismaBuild, plus compile. Original regression failures, successful CPU fixes and #399's 11-passed CUDA receipt were independently checked against terminal records, canonical CAS results and exact executed source. Final architecture checks cover the merge with #404. See `experiments/measurements/capture-followups-20260908/README.md` for commands, receipt paths, skip limits and source identities.

No measured latency or physical-I/O speedup is claimed. The active full capture continues on its frozen source.

Closes #395. Closes #396. Closes #398.


RobTand (commented):
VERDICT: MERGE when undrafted. Reviewed head fe078d9e3 as the integration of #393, #397 and #399.

All three triage branches are ancestors of the head, and the test files `tests/test_tessera_calibration_cache.py`, `tests/test_bounded_hessian_sidecar.py` and `tests/test_tessera_selected_source.py` are byte-identical to the branch heads. The only runtime-file deltas beyond the three branches are prose: the completed-anchor comment in `prismaquant/tessera_campaign.py` from commit 0fbed5b4d, and the docstring in `tools/dispatch_tessera_campaign.py` from commit f7d8fd3e9 that now calls the streaming-row figures derived bounds rather than measured quantities. Both corrections are accurate; the second is the point #390 makes.

Receipts. Your head runs seal parent 623fc3a4c: `0666c57d2880`, `0c557be5c006`, `391a583bf59a`, `4cbf63d9c4ed`, `7bbe9de1afdd`, `a875cc5fef7a`, `c1caf4d9ff3c`, `d8fb57407ae3`, `f1c48346cda1`, `fc361ec5fb48`, all dl380g10, rc 0, priority -10. The commits after 623fc3a4c are the README, the dispatcher docstring and the merge of main at #404, which changed no runtime file, so those receipts cover the runtime at head. The three red rows `182ce135adfc`, `49801762d3c7` and `5d53e29e3fa8` and the green rows `7b7f2c7fc4eb`, `a326b99c339f` and `242b22b88cf5` are the original per-PR receipts and resolve as recorded. The remaining 64-hex tokens in the audit files are content digests, not action keys.

The `docs/ARCHITECTURE.md` stamp describes the writer refusal and page-release contract as the code implements them and claims no speedup.

PR #410, opened 2026-09-08
Byte-verify only the shards a selected-source row reads

RobTand:
Closes #409. Code half of #388 (the at-scale profile on a real GLM selected row stays open there; it cannot run while the GLM canonical capture holds the frozen source).

## What changes

A selected-source row used to re-hash every file under the model root before touching an anchor, because `capture_identity` was designed for the canonical capture and #376 made it a per-row cost. Now the row byte-verifies only the shards it reads and inherits the rest.

- `prismaquant/cost_streaming.py`: `StreamedCausalLM.selected_source_shards(names)` returns the selected layers' shards, the fixed-state shards (every tensor outside the layers prefix: embed, norms, head, visual tower, all materialised at construction), and the tensor to shard map, from the runner's own weight map.
- `prismaquant/tessera_calibration_cache.py`: `capture_identity(..., verify_shards=None)` wraps a new `capture_identity_with_verification`. With `verify_shards` set, the named shards are hashed and every other shard's digest comes from the census-sealed roster `expert_projection.producer.source`. It refuses when the roster is missing, a named shard is not on disk or not sealed, or an on-disk shard is neither read nor sealed. Small files are always hashed. The identity itself is unchanged, so a row that verifies less must still reproduce the canonical identity.
- `prismaquant/tessera_campaign.py`: `selected_source_read_set` cross-checks the derived read set tensor by tensor against the sealed `tensors` map and refuses disagreement. `prepare_selected_source` picks the mode from the census: `roster-inherit` when a roster is sealed (and refuses when the canonical manifest is not pinned by `--calibration-cache-sha256`, since identity equality with the pinned manifest is the only attestation of inherited digests); `full-root` when the census seals no roster (dense-only models), hashing the whole root exactly as before. Mode and the verified, inherited and auxiliary file lists are stamped into `selected_source_preparation.source_verification`, not into the identity.
- `docs/ARCHITECTURE.md`: selected-source section updated and re-stamped, with the real-census bound.

## Bound on the real census (derived from `census.json`, not a measurement)

GLM-5.3-Flash-BF16: fixed state spans 2 of 120 shards (9.5 GB, shards 00001 and 00120). A typical selected row reads 5 shards, 25.6 GB of the 642.7 GB source, instead of hashing all of it.

## Tests

[...]

Issue #388, opened 2026-09-08
Selected-source anchor rows re-hash the whole source checkpoint per row: about 84.8 TB of NFS reads across the 132 GLM rows


RobTand:
Post-merge audit of PR #376 at main 94e30ade7, finding 2 (https://github.com/RobTand/prismaquant/pull/376#issuecomment-5584543796).

`prismaquant/tessera_calibration_cache.py:62` (`capture_identity`) hashes every `*.safetensors`, `*.json`, `*.model` and `*.txt` file under the model root on every invocation, by design: mtimes never authorize reuse. `prismaquant/tessera_campaign.py:3993-4000` calls it once per campaign process with `release_read_pages=True` and the guard as `resource_check`. The helper predates #376; #376 is what turns it into a per-row cost, because a selected-source campaign is one process per row.

Scenario, measured inputs: the census-sealed GLM-5.3-Flash-BF16 source is 120 safetensors files, 642.65 GB, on `/mnt/shared` NFS. Each of the 132 selected rows reads and SHA-256s 642.65 GB before touching an anchor, about 84.8 TB across the campaign. Estimate only, not measured: at the recorded NFS read rate of 0.7 to 1.0 GB/s that is 11 to 15 minutes per row and 24 to 33 hours of NFS reads campaign-wide. Neither the dispatch plan nor the row timeout models this time. No receipt runs a selected row on the real source, so the per-row cost is unmeasured at scale.

Expected: keep the bytes-based identity, but pay for it once. Seal per-file hashes in the census at capture time, and have a selected row hash only the shards it installs, checking them against the sealed roster. A row that installs no shard hashes nothing. Full-root re-hashing stays available for the canonical capture itself.

Acceptance: a regression that counts hashed bytes per selected row on the fixture before the change; the same count after; before and after in-process profiles plus both Sparks' Netdata on one real GLM selected row through PrismaBuild; `docs/ARCHITECTURE.md` updated in the same commit.

RobTand:
Code half is up as #410 (closes #409, the narrow code-half issue). It byte-verifies only the shards a selected row reads and inherits the rest from the census-sealed roster; dense-only censuses fall back to full-root hashing, stamped.

Derived bound on the real census: fixed state spans 2 of 120 shards (9.5 GB), a typical row reads 5 shards, 25.6 GB of the 642.7 GB source.

[...]

PR #408, opened 2026-09-08
Add opt-in single-read capture loading with bounded admission

RobTand:
Capture replay, sealing, and qualification currently hash each artifact and then read it again to deserialize it. Add an explicit bounded verified-load policy that authenticates and deserializes one admitted buffer through the existing activation owner, checks archive/storage geometry before allocation, and releases serialized memory before GPU transfer. Admission prices private serialized bytes, possible kernel page residency, decoded storage, and bounded scratch independently. Canonical capture bytes and identity stay compatible; the default loader remains unchanged.

Native paired ABBA measurements on Sparklina with the pinned producer container show exact X/H/file parity and half the logical input bytes. Prefetch mean latency falls 21.6%; replay increases 4.3% and sealing increases 12.5%, so this is an opt-in capability, not a default performance promotion. The checked-in report retains the initial failed and slower experiments, in-process profiles, process I/O, both-host Netdata, receipts, and the limits of energy attribution for these short operations: [measurement report](https://github.com/RobTand/prismaquant/blob/5f3469db357db1109a9b6b2ef8e987741ca46ade/experiments/measurements/verified-capture-load-20260908/README.md).

Validation through PrismaBuild:

- Full CPU suite across all 426 test files: **6,867 passed, 187 skipped, 3 expected failures**, plus 192 passing subtests. The 64 PB shards used two pytest workers and one native thread per worker. Six initial 6 GiB shards hit their cgroup memory caps; only those identical populations were retried at 12 GiB and passed. Actual retry peaks were 7.34–8.84 GiB. Source snapshots, logs, canonical CAS receipts, cleanup, and complete file coverage were independently audited.
- Targeted behavioral red/green coverage includes single-read verification, mutation/TOCTOU refusal, archive/storage/alias bounds, allocation guards, and release lifetimes. Native CPU Torch 2.13 scalar-finite probes validate bounded scratch allocation; the final documentation/compile action passed 19 tests.
- Native final action `807d24936563a101e8a66978c16f9adb1d696596b314cea2fd37ae300d61f2df` passed all 12 operations under the 6 GiB physical / 2 GiB GPU reservation, with a 3.78 GiB observed physical peak and no OOM.

Full CPU audit: `/mnt/shared/tessera-measurements/glm-canonical-census-20260908/verified-capture-load-design-01/root-full-cpu-consolidated-01.json`. Final documentation/compile action: `a49557fdce5b2be2c58f86b508e62d61ecb2619d6f92f04e2d620ddc4a66f124`.

[...]

PR #411, opened 2026-09-08
Authenticate consumed source shards through complete-capture descriptor leases

### LINKED ISSUE #409: Selected-source rows byte-verify only the shards they read; inherit the rest from the census roster (code half of #388)
Code half of #388. #388 stays open for its measurement half: before and after in-process profiles plus both Sparks' Netdata on one real GLM selected row through PrismaBuild. That measurement cannot run while the GLM canonical capture holds the frozen source, so this issue carries the part that can land now and be verified on the fixture.

Scope:

- A selected-source row (`prismaquant/tessera_campaign.py`, the `selected_source` branch of `main`) byte-verifies only the shards it reads: the selected layers' shards plus the fixed-state shards (embed, norms, head, visual tower), derived from the runner's weight map and cross-checked tensor by tensor against the census-sealed roster `expert_projection.producer.source` (`files`, `tensors`).
- Every other shard's digest is inherited from that roster. The recomputed `capture_identity` must equal the pinned canonical manifest's identity before any selected layer installs, and the manifest must be pinned by `--calibration-cache-sha256`; identity equality with the pinned manifest is the only attestation of the inherited digests.
- A census that seals no roster (dense-only models declare no packed expert projection) falls back to hashing the whole source root, exactly as the canonical capture does, and says so.
- What each row verified or inherited is stamped into `selected_source_preparation.source_verification` (`mode`, `byte_verified`, `inherited_from_census_roster`, `byte_verified_auxiliary`, `selected_layers`, `layer_shards`, `fixed_state_shards`), not into the identity.
- Small files (`config.json`, the index, tokenizer files, the census) are always hashed.

Acceptance: a regression that counts hashed bytes per selected row on the fixture before the change and after it; a test that the shards a row opens for tensor data are a subset of the shards it byte-verified; `docs/ARCHITECTURE.md` updated in the same commit.

Bound on the real census, derived from `census.json` (not a measurement): the GLM-5.3-Flash-BF16 fixed state spans 2 of 120 shards (9.5 GB), so a typical row reads 5 shards, 25.6 GB of the 642.7 GB source, instead of hashing all of it.

Follow-up owed separately: the census seals a producer source roster only through the packed-expert projection, so dense-only censuses cannot inherit at all. Sealing the roster for every model belongs to the census writer (`prismaquant/tessera_expert_projection.py`).

RobTand:
Selected capture consumers previously hashed every checkpoint payload through `capture_identity` before reading their selected rows. This change binds the complete manifest and census before model construction, then authenticates each consumed shard once through a held descriptor shared by the existing source readers. The original complete capture identity remains intact; unused shard payloads are deferred, and metadata, mutation, replacement, escaped reads, and incomplete captures fail closed.

Descriptor leases cover asynchronous readers, construction failures, runner shutdown, and projected source tensors. The tiny complete-capture regression reduces hashed payload from 2,130,208 to 32,960 bytes while preserving selected tensor values and manifest bytes. This is a CPU byte-count result; the original GLM capture and native before/after measurements remain pending, the remaining native measurement is tracked in #388.

Validation: PrismaBuild integration suite **355 passed, 1 CUDA-only encoder-memo skip, 2 passing subtests**, covering 23 files. Root review verified all successful terminal records, receipts, result hashes, and source snapshots. One fixture requiring two cache workers was initially under-reserved at one CPU; only that file was resubmitted at two CPUs and passed. The final documentation checks also passed 19 tests. Full CI then exposed a prefetch fixture that bypassed the context initializer and lacked the new optional field; PB reproduced all nine failures. The one-line fixture correction passed all 32 scheduling/admission tests; runtime code is unchanged. The final full CI passed 6,879 tests and 192 subtests, with 201 skips and three expected failures (run 34253411730). Earlier FIFO and unused-payload regressions were demonstrated before their fixes. Evidence and limitations are in [the measurement record](https://github.com/RobTand/prismaquant/blob/0f7515f730c1fc7831f1c36a7734cf438b2b66d8/docs/measurements/selected-source-authentication-2026-09-08.md).

Separate incidental prose correction: removed an unsupported GPU-utilization saturation claim from the source-reader docstring.

Closes #409. Refs #388.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,30 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-08 · `integrate/capture-followups-20260908`. Stamps
+As of: 2026-09-08 · `fix/selected-source-authentication`. Stamps
 follow, newest first, each recording its own branch and date.
+
+
+Re-stamped (2026-09-08, `fix/selected-source-authentication`) for selected
+source authentication from a hash-bound complete canonical capture (#388).
+The full source roster, calibration identity and capture identity stay byte
+identical. Before model construction, the campaign validates that complete
+manifest against the census and freshly hashes its metadata and producer
+auxiliaries. Existing streaming readers authenticate each consumed shard once
+per selected invocation through a held read-only descriptor, including early
+nonbody state, FP8 scales and projected expert reads. Header-only inspection of
+an unconsumed shard does not hash its payload. The descriptor owner is shared
+by existing reader/prefetch threads and remains alive through their completion;
+source mutation or pathname replacement fails closed. Files must remain stable
+through these read leases; stat fences reject changes and never authorize a
+cached digest. The preparation receipt reports consumed-file authentication
+separately from the unchanged canonical identity. There is no persistent hash
+cache, new source residency mechanism, or relaxed partial-capture acceptance.
+CPU regression and tamper/thread/descriptor tests qualify this ownership
+contract; native selected-row I/O, memory and timing still require a matched
+before/after run with the original complete GLM manifest and both profiler and
+host telemetry. No full-GLM speed or fit gain is claimed by this change.
+Gate: `tests/test_selected_source_authentication.py` and existing selected,
+streaming reader, capture and campaign tests.
 
 Re-stamped (2026-09-08, `integrate/capture-followups-20260908`) for Hessian
 writer refusal and completed-file page release (#395, #396). A tensor-bearing
--- a/docs/measurements/selected-source-authentication-2026-09-08.md
+++ b/docs/measurements/selected-source-authentication-2026-09-08.md
@@ -59,3 +59,23 @@ checkpoint hash, GPU run, menu, serving gate, or calibration draw changed.
 Native selected-row I/O, memory and timing qualification waits for the original
 complete manifest and a matched before/after run with in-process profiling and
 both hosts' telemetry. No full-model speed or fit result is claimed here.
+
+## Root integration qualification
+
+Integration with the verified capture loader at `3a94d7520b67` passed **355
+CPU tests across 23 files, with one CUDA-only encoder-memo skip and two passing
+subtests**. The root audit verified all 23 successful terminal records, result
+receipts and source CAS snapshots; each source tree differs from that integration
+commit only by its PB closure manifest. The invocation and consolidated coverage
+are recorded beside the earlier evidence in
+`root-integrated-cpu-invocation-01.json`,
+`root-integrated-cpu-consolidated.json`, and
+`root-integrated-positive-cas-audit.json`.
+
+One initial shard reserved one CPU while its qualification-window fixture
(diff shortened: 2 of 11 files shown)


RobTand (commented):
Review of #411 at head bfa9815d67 (base 59c7aaa450, post-#408).

**Receipts, verified by canonical CAS lookup, all dl380g10, gen 437ad81b1129, priority -10.** The integration audit at `selected-source-authentication-01/root-integrated-cpu-consolidated.json` names 23 shards: 22 done rc=0 at parent c75f259fef plus `568bfcd5` (9 passed, the two-CPU retry) at 3a94d7520b; one failed row at c75f259fef is the one-CPU under-reservation the doc describes. `git diff c75f259fef 3a94d7520b` is empty, and `568bfcd5`'s sealed snapshot differs from the head only by the 20-line measurement doc, so the 355 passed cover the head's code, including `test_glm_campaign_streaming.py` (12 passed, the end-to-end streamed selected row), `test_tessera_selected_source.py`, `test_selected_source_authentication.py` (26) and `test_verified_capture_load.py` (70) on the merged loader. The focused `1c95e20c` run predates the #408 rebase and is superseded. No sparklina rows; the body says CPU x86 only.

**Design, read at head.** Expected digests come from the pinned manifest's own `source_files`, cross-checked against the census producer roster when one exists (`tessera_calibration_cache.py`, `CaptureSourceAuthentication.__init__`). So a dense-only census, which seals no producer roster, still inherits, and the main regression test runs exactly that case. Consumed shards are hashed once at first tensor read through a held `O_NONBLOCK` descriptor with fstat and stat fences before and after; header-only opens do not hash. `authenticate_selected_capture_source` refuses without a 64-hex `--calibration-cache-sha256`, and identity equality with the pinned manifest is the attestation. Every streaming reader site (`_build_weight_map`, `_build_fp8_scale_inv_map`, `_materialize`, `_read_layer_to_device`, `_apply_fp8_dequant_inplace`, `_estimate_layer_cache_bytes`, `source_unit_weight`) routes through the owner when one is given; the two former swallow-and-return-default branches in `streaming_model.py` re-raise under an owner. Order at the call site is right: `runner.shutdown()` in the `finally` of the snapshot, then projection, then `receipt()` and `close()`, with the `ExitStack` closing the owner last if anything raises earlier. Default path unchanged: every new parameter defaults to `None`.

**Supersedes #410.** My #410 derives a read set from the runner and inherits from the census roster, so it needed a full-root fallback for dense models and a roster cross-check this design does not need. I will close #410 and #409 when this merges. Two things from #410 worth carrying later, not blocking: an instrumented test that tensor-data reads come only from authenticated shards (here that is by construction of `_payload`, and `metadata_only_shards` in the receipt shows it), and the real-census bound already on #388 (fixed state 2 of 120 shards; a typical row reads 5 shards, 25.6 GB of 642.7 GB).

**Non-blocking.** The stamp names branch `fix/selected-source-authentication`; the PR branch is `integrate/selected-source-authentication-20260908`. `Refs #388` is correct here since the at-scale profile stays open.

VERDICT: MERGE

Issue #420, opened 2026-09-08
Implement recomputable full-engine fixed-resource producer admission for runtime v2


RobTand:
The measured-runtime v2 loader cannot admit any full table, even after native operator and runtime-relation receipts exist: `admit_fixed_resources` unconditionally raises after reading its receipt.

Evidence at main `058bcc32a7966066a15a24aa744c14ec1bb1598a`:

- `prismaquant/runtime_provenance.py:375-389`: no qualified recomputable full-engine resource partition schema exists; `full_model_resources` presence leads to unconditional refusal.
- `prismaquant/measured_runtime_prices.py:446-470` and `:478`: schema v2 invokes producer admission before returning a usable runtime table.
- The allocator's `--measured-runtime-table` path needs this admission to use measured prefill/resource constraints. Native dense/MoE rows and matching runtime libraries alone cannot close fixed work, KV, resident memory or scratch ownership.

Severity: P2, missing provenance/producer contract; the current refusal is correct and prevents unsupported resource claims. This is a software contract gap as well as the missing full-engine measurement tracked by #237 and #267; #323 deliberately left this gate closed.

Required fix: implement a versioned full-engine producer partition that accounts for candidate versus fixed work/resources without double counting, bind its raw measurement receipts to the same full runtime/source/calibration/workload/topology, and independently recompute the partition in this consumer. Add negative tests for omission, overlap, stale runtime and resource mismatches. Preserve the refusal until the real producer contract is available. No synthetic test table or native panel can stand in for full-model evidence.

I could not fix this within the six-recipe intake audit: it requires a new full-engine producer/admission contract and actual matching served measurements, beyond a bounded metadata repair and outside the authorized no-new-native/probe preparation scope. No new full-engine resource or performance measurement was taken.

RobTand:
Closing: the software contract this issue asked for is on `origin/main`.

`admit_fixed_resources` no longer refuses unconditionally. Two commits, both verified as ancestors of `origin/main` (`142cc0b8` at time of writing):

[...]

PR #423, opened 2026-09-08
Add shared-capture issuance and preserve GLM recipe provenance

### LINKED ISSUE #421: Refuse stale Tessera and serving metadata when frontier selection changes assignment
`select_validated_frontier` can change an assignment while copying Tessera wire/scale/population and serving/runtime claims from the overwritten destination layer config.

Evidence on main 058bcc32a7966066a15a24aa744c14ec1bb1598a:

- `prismaquant/select_validated_frontier.py:1314-1344` loads destination metadata and replaces only CB and whole-artifact-budget identities before writing the new assignment.
- `prismaquant/allocator.py:5613-5765` produces assignment-dependent wire, scale, serving and retained-population fields on its final recipe.
- `prismaquant/validate_assignments_kl.py:2006-2025` does not carry those selected fields in the resolved payload.

P1: a decision can read serving/resource claims belonging to another assignment. Existing exporter gates may reject stale wires; this finding does not claim that an incorrect Tessera artifact has served.

Fixed in isolated commit 26b97c6a4a: refuse before all output writes when destination Tessera/population/serving metadata exists and the full canonical assignment changes. Preserve existing claims for an unchanged assignment; downstream provenance gates still validate them. No new source/calibration authentication is implied. Full Pareto-producer/validator metadata propagation remains unimplemented, so changing a Tessera recipe must use the allocator's final metadata and selected-wire gates.

PrismaBuild reproduced 7 wrong successful writes plus 1 unchanged-assignment control (2b7762dec63d); the fixed selector/layer-config/docs suite passed 87 tests (a45d7fc4dcfa), with compile receipt 30a9a302aab9. Actual logs, source bundles and CAS audited in experiments/measurements/frontier-metadata-owner-20260908/. No new GPU or serving measurement.

This was fixed in the discovery branch; issue records the integration and the remaining producer handoff limitation rather than deferring the known safety repair.

### LINKED ISSUE #422: Expose closed GLM capture compatibility issuance through a bound plan
The closed GLM derivative compatibility API added by #415 has no deterministic CLI for issuing its consumer receipt after the original capture completes. Calling the Python API currently requires a caller to construct and bind the actual corrected model and assemble producer/native evidence manually.

Add a byte-bound plan and CPU/meta preflight/issue commands using the existing receipt API and the streaming path's actual meta skeleton constructor. Preflight may report the original capture/terminal/CAS completion fields as pending, but must still authenticate the corrected image, config, real modules and existing native proof. Issuance must reject pending/incomplete producer evidence before model construction and retain all original capture/runtime/CAS/native gates and exclusive output creation.

Scope: no weight loading, GPU forward/backward, duplicate measurements, capture mutation, full-model qualification or serving change. Implementation is isolated at `codex/glm-capture-compatibility-issuance`, based on reviewed #415 commit `3070766203`. Related CPU/refusal tests and actual-image CPU preflight run through PrismaBuild; original receipt issuance awaits reviewed complete producer evidence. Root owns integration.

RobTand:
The shared GLM capture needs a deterministic way to issue its corrected-consumer compatibility receipt, and downstream frontier selection must preserve assignment-specific provenance. This integrates the reviewed CPU/meta issuance CLI and fixes stale destination metadata. Closes #421. Closes #422. Integrates #424.

The compatibility CLI binds an explicit plan to the real corrected model, original producer and existing numerical evidence. Preflight permits four pending original-completion fields and writes no receipt; issuance refuses pending/incomplete producer evidence before model construction, then reuses the existing exclusive-create issuer. The extracted streaming skeleton constructor is shared with the ordinary path. Separately, frontier selection refuses before writes when a changed assignment would inherit old Tessera wire, scale, population or serving/resource claims. Matching formats preserve existing metadata for downstream gates; they do not authenticate source or calibration.

The current resource audit derives admission for **all 132 original groups / 36,423 units without sampling**, with a 98.041521 GiB maximum under the 104 GiB worker budget; native full-stack fit remains unmeasured. The accompanying GLM intake maps all six variants to one immutable capture, measured candidate union and joint cost handoff. It distinguishes direct served quality/prefill comparison from the optional runtime-v2 predictor. The latter remains closed, with missing observation/accounting requirements documented under #420. Six artifacts and measured speed/quality results have not been produced, and no actual compatibility receipt has been issued.

Validation through PrismaBuild: the metadata regression reproduced seven wrong successful writes plus one unchanged-assignment control; the fixed selector/layer-config/docs suite passed **87 tests** and compilation. Root integration docs passed **19 tests**. Runtime-contract/design checks passed **124 tests**, retaining refusal. The compatibility CLI/shared-constructor suite passed **170 tests with zero skips**, plus compilation; actual corrected-image CPU/meta preflight authenticated all **34 KDA modules**, matched native execution identity and left CUDA uninitialized. Root independently checked actual logs, canonical CAS results, source snapshots and cleanup, including all 13 compatibility actions. The first full CI found one stale helper-name assertion (7,013 passed, one failed); it is replaced by a public-API metadata-preservation test. Targeted follow-up passed **92 tests**, and integrated architecture/staleness checks passed **19**. Refreshed full CI is pending.

Evidence is under `experiments/measurements/frontier-metadata-owner-20260908/`, `glm-six-variant-recipes-20260908/`, `glm-capture-compatibility-issuance-20260908/` and `docs/measurements/runtime-fixed-resource-design-2026-09-08.md`. Separate incidental prose commits correct two allocator comments, the GLM layer count and caller-reserve accounting. No runtime predictor, producer pin, GPU performance claim or new probe is introduced.


RobTand (commented):
Reviewed at head `1e64c49236aa3f9a6a853ba91cf6a2774cf6261c`.

**VERDICT: MERGE**

## Mechanism, verified at this head

`_destination_metadata_for_assignment` (`prismaquant/select_validated_frontier.py:77`) is called at `:1345`, ahead of every output write: the layer config at `:1370`, the assignment payload at `:1393`, and the summary after it. Nothing in `main` between `:989` and `:1345` writes a file, so the refusal leaves the destination byte-identical, which is what the negative test asserts.

The comparison is like for like, which was the thing worth checking. Both sides go through `_assignment_from_payload` (`:107`), so string-shaped and dict-shaped entries canonicalize the same way. More important, the Pareto payload the selector reads is already in the destination's key space: `_expand_assignment_for_seed_json` in `prismaquant/allocator.py` runs the same packed-group and fused-sibling expansion, the same `fixed_format_assignment` update and the same `promote_serving_units` that build `assignment_expanded` at `allocator.py:5358-5370`, and `pareto_seed_records` stores that expanded form (`allocator.py:4114`, `:4122`). An unchanged point therefore cannot refuse spuriously on a MoE or fused model, or on a quantized head. The absent-destination path returns an empty dict, and `_assignment_from_payload` is only reached when a coupled key is present, so a destination with no coupled metadata is unaffected.

## Receipt lineage

Checked from the snapshot bundles, not only the audit JSON.

- RED `2b7762dec63d`, row failed rc=1, parent `058bcc32a796`, tree `e24e98056d14`. That tree differs from this head only by the 33 lines of the guard in the selector; `tests/test_select_validated_frontier.py` is byte-identical. The driver was mutated, not the fixture, and the 7 named failures are the 7 coupled keys.
- GREEN `a45d7fc4dcfa`, done rc=0, parent `058bcc32a796`, tree `55f84ce23703`. The selector and its test are byte-identical to this head; the only other delta is the two allocator comment edits.
- `30a9a302aab9` and `2f047451fde4` done rc=0 at the same parent, `8b3ab965cd63` done rc=0 on sparklina at parent `47aec941d382`. All at priority -10, runtime generation `437ad81b1129`.

`docs/ARCHITECTURE.md` carries the stamp in the same PR and names its gate.

## Non-blocking

1. A malformed destination now raises instead of degrading. `read_layer_config_metadata` (`prismaquant/layer_config.py:42-47`) catches a read or parse error and returns an empty dict; the new helper calls `_load_json` (`:35`) directly, so a truncated destination ends the run with a decode error rather than the refusal message that tells the operator what to publish. Failing closed is right here; catching and re-raising with the same guidance would read better.

2. The coupled set is a denylist: the `tessera_` prefix plus five exact names. An assignment-coupled key added to the allocator's stamp later carries silently until someone remembers this list. `achieved_bits` is already in that class and is safe only because `prismaquant/shipcard.py:3495-3516` prefers `selected_by` and `selected_achieved_bits` and documents the staleness. A test asserting the allocator's stamped keys partition into coupled and not-coupled would keep this from decaying.

RobTand (commented):
Follow-up for the commits added after my first pass. Head is now `f3cd7c7af56147217562a6a629613f626d6c27e5`, which absorbs the closed CPU issuance CLI from the withdrawn #424.

**VERDICT: MERGE** (unchanged; the added code is covered)

## What I checked in the added part

The new gate has a consumer, which is the first thing worth confirming for a refusal contract. `require_capture_compatibility` (`prismaquant/glm_capture_compatibility.py:215`) is called from `prismaquant/tessera_joint_aura.py:383` and `:699`, so this is the issuer for a receipt the capture path already demands, not a gate nothing reads. Both absent-state directions fail the way they should: no derivative and no binding returns None, a corrected runtime with no binding refuses, and a binding with no corrected runtime refuses.

`create_capture_compatibility` opens the receipt with mode `xb`, so a second issuance cannot overwrite a published receipt, and `execute_issuance_plan` refuses an existing output before it constructs anything. The pending and bound halves are all-or-nothing by construction, and `issue` requires an empty pending set.

The `build_streaming_skeleton` extraction in `prismaquant/streaming_model.py:1351` is behavior-preserving. The old inline block rebound `config` from `_skeleton_config_and_class`, and the extracted function rebinds it only locally, so the question was whether the caller used the resolved config afterwards. It does not: `config` has no further use in `_build_streaming_context` after the skeleton is built, in either version.

## Receipts

Thirteen CPU rows, all done rc=0 at parent `2196dbcc2a47`, priority -10, including `02d72f6dbfa0`, `1899ca8cbb5a` and `f44db817c044`. The only code delta between that tree and this head is the frontier change already receipted by `a45d7fc4dcfa` plus the two allocator comment edits, so `glm_capture_compatibility.py`, `streaming_model.py` and the new CLI test are byte-identical to what those rows ran.

The two non-blocking notes from my first comment still stand.

Issue #389, opened 2026-09-08
Selected-anchor encoder memo holds --anchor-batch-size entries while pending order is rung-major, so every (unit, rung) pair refactorizes; no CLI test covers an s: stack group


RobTand:
Post-merge audit of PR #376 at main 94e30ade7, findings 3 and 6 (https://github.com/RobTand/prismaquant/pull/376#issuecomment-5584543796).

`prismaquant/tessera_campaign.py:338` sizes the encoder memo to `--anchor-batch-size`, default 1, while `:581` emits pending work in rung-major order. With one memo slot and rung-major order, every (unit, rung) pair refactorizes its Hessian instead of reusing the factorization across rungs of the same unit. Wire bytes and price rows are unchanged, so the receipts pass; the cost is unprofiled.

Expected: order pending work unit-major, or size the memo to the number of members per family, then profile the anchor loop before and after (in-process profiler plus both Sparks' Netdata, power against the 140 W envelope) so the claim is the measured delta.

Coverage gap from the same audit: no CLI test exercises an `s:` stack group, so the completed-anchor page-release path (`tessera_campaign.py:517-527`) is untested end to end. Add one on the tiny native GLM fixture.

RobTand:
Coverage half is up as #413 (closes #412, the narrow coverage-half issue): a CLI test that drives a real selected-source row over a planner-drawn `s:` stack group and asserts the rendered shard and `.tessera` wire of every completed anchor are released exactly once. The bite proof is a driver mutation of `_finish_anchor`'s release branch, x86 only (the fixture skips on sparklina for the pre-existing `glm5_next` reason).

Diagnostic recorded by that test for the reorder half, not asserted: at `--anchor-batch-size 1` the pending order is unit-major and the memo shows one factorization per unit (hits=6, misses=6). The rung-major regrouping fires only above batch size 1 and was not exercised, so this issue stays open for that half and its at-scale profile.

RobTand:
At-scale profile of the reorder half, measured on the live GLM pricing campaign.

Code as it runs today (PB snapshot `a6a464c7db65`, the checkout the campaign executes): the memo is `prismaquant/tessera_campaign.py:405-422`, its capacity is bound at `:4407-4411` to `args.anchor_batch_size`, the rung-major regrouping is `_anchor_batches` at `:648-663`, and pending is built family-major then member then rate at `:4795-4816`. The cited `:338`/`:581` have drifted; the mechanism is unchanged.

[...]

PR #426, opened 2026-09-08
Derive the selected-anchor plan terms and measure the row's process floor

### LINKED ISSUE #390: autoscale.selected_anchor_resources uses chosen multipliers with no process baseline; native guard peak 2.61 GB against a 12.6 MB plan

### LINKED ISSUE #451: Decorator docstring overstates who reads PRISMAQUANT_PROD_ACT_SCALES
`_restores_activation_scale_env` (added in #446) says in its docstring:

> Nothing outside `execute` reads the key, so restoring it on the way out leaves the campaign byte-identical and leaves the process as it was found.

The first clause is false, and #446 itself is the proof: `production_weight_cache._local_forward_render_score` reads `PRISMAQUANT_PROD_ACT_SCALES`, which is precisely why a leaked `"0"` disarmed the activation clip and failed `test_stock_nvfp4_cache_score_is_unchanged`.

The true claim is narrower: nothing needs *this command's* value to persist after `execute` has returned. That is what the restore relies on, and it is what the docstring should say.

Wording only. No behaviour change, and no test is owed for a reversible prose edit.

RobTand:
## What

`autoscale.selected_anchor_resources` built its phase plan from chosen
multipliers. No term named the allocation it bounded, and the plan's sum of
deltas was admitted against `CaptureMemoryGuard`'s cap, while the guard reads
absolute process bytes: cgroup `memory.current` plus whole-process
`torch.cuda.memory_reserved`. Plan and reading were in different units, and
only `declared_headroom_bytes` stood between them.

Each term now cites the line that allocates what it bounds, and the row
measures its own process floor so the two units meet.

## The terms, and where each comes from

- `export_input_page_window_bytes = widest_h + len(counts)*16384`. The Hessian
  half is one FP32 `[in, in]` tensor record, the visible prefix
  `_save_hessian_capture_with_page_release` leaves after each `data/` record.
  The metadata half is a **gap**: pickle framing and roster name lengths, not a
  shape or a dtype. Left unchanged with a comment saying so.
- `serialization_scratch_bytes = 2*widest_h`. The phase's peak is the capture
  digest, not the writer. `tessera_export_lane.hessian_capture_sha256` holds
  `H.detach().cpu().contiguous()` and then the bytes object of that copy at the
  same time, and across the loop's rebind the previous unit's copy is still
  referenced. `torch.serialization._save` stages one CPU copy per device
  storage, live only across one record, which is the smaller transient. The
  multiplier was right; its stated reason was not.
- `factorization_scratch_bytes = 2*widest_h`, down from `4*widest_h`. Two
  sequential stages, each peaking at two FP32 copies of the widest H: the seal's
  `tensor_identity` (a CPU copy and its bytes), then `regularize_hessian`'s
  `H.float().clone()` held while `block_ldl`'s triangular solve writes an
  `[in, block]` output with `block <= in`.
- `entry_validation_bytes = 2*widest_capture_entry`, where an entry is one
  unit's FP32 H plus its FP32 X in the loader's own `_capture_storage_bytes`
  arithmetic. The previous form summed a widest H and a widest X measured on
  different units, which charges an entry no unit has. The loader asks for this
  same bound at both of its `before_capture_prefetch` reserve sites, so the plan
  and the loader now agree. The finite check allocates bool masks, one byte an
  element and one at a time under a short-circuiting `or`, so it does not raise
  the peak.
- `encoder_memo_bytes` is sized by the capacity the plan publishes
  (`encoder_memo_capacity`), and `tessera_campaign` builds the memo with that
  field instead of re-deriving it from the anchor batch width. The memo policy
  now has one owner, so #389 moves one field.
- `compatible_batch_weight_bytes` is a **gap** and is unchanged. The
  PrismaQuant-side copies are traceable and come to eight bytes an element; the
  producer's own working set inside `encode_linear` is not traceable from this
  repository, so the sixteen bytes an element it has always charged stays as the
  conservative bound rather than being replaced by a smaller number that omits
  the encoder.

## The baseline

`CaptureMemoryGuard` stamps its first `check` reading as a measured baseline:
label, bytes, and `measured_in_process`. It also reports `peak_checkpoint` and a
per-phase-prefix peak map, so a peak names the phase that held it. The selected
row stamps that baseline beside its plan in `selected_source_preparation`, and
selected admission compares the plan against the cap less the baseline instead
of against the raw cap. The guard's own refusal arithmetic is untouched: its
readings are already absolute, and subtracting a floor there would count it
twice.

No pre-run process baseline is invented for the dispatcher's `_row_memory_gb`.
The honest options were a baseline recorded from a prior receipt with its host
and runtime scope, and leaving the spec's declared headroom as the only pre-run
term. This takes the second, and the plan says so in a `baseline_policy` field
rather than leaving a reader to infer it.

## What the plan does not cover, measured and separated

The native row now runs two encode steps, and the campaign takes the
`before_selected_anchor_batch` / `after_selected_anchor_batch` bracket per
occurrence, stamping one growth figure per step on the receipt as
`anchor_batch_growth_bytes`. That separates two quantities one step could not:

- The first step carries the runtime's one-time first-use cost, the first CUDA
  factorisation and the first `encode_linear`. It measured 165,642,240 bytes,
  it follows no shape in the roster, and it is left uncharged with the
  checkpoint map that attributes it. A constant sized to it would be a
  multiplier by another name.
- The second step is the steady state the plan does charge for: 2,654,208
  bytes against a `resident_anchors` phase of 10,263,316. The row asserts
  that every step after the first fits that phase, so a forgotten term fails
  there rather than disappearing into declared headroom or into a whole-row
  figure the first step dominates.

The assertion was shown to bite. With five of the `resident_anchors` phase's
terms deliberately under-priced to one byte, `encoder_memo_bytes`,
`factorization_scratch_bytes`, `compatible_batch_weight_bytes`,
`entry_validation_bytes` and `source_validation_bytes`, which leaves the phase
at 1,325,061 bytes, the native row fails on the second step's 2,654,208 bytes
and prints both growths (row `034099017411`, sparklina, rc=1, parent
`37d86e6263`, a throwaway probe commit that is not on this branch; its tree is
in that row's CAS snapshot bundle).

Guard telemetry, quoted from that probe row because a passing row prints
nothing. The guard code, the fixture and the roster are the ones at head; only
those five plan constants differ, and none of them changes an allocation.
Baseline 2,211,225,600 bytes at `before_capture_hash:config.json`, being
2,156,699,648 cgroup `memory.current` plus 54,525,952
`torch.cuda.memory_reserved`. Peak 2,369,073,152 at
`after_selected_anchor_batch`, so 157,847,552 above the baseline, and the last
checkpoint read 2,230,661,120 cgroup plus 138,412,032 CUDA. The first step's
own growth exceeds the row's because its floor sits about 10 MB below the
baseline, which is read earlier, at the capture hash. The head row asserts that
the CUDA readings are above zero rather than printing them, which is what
distinguishes a row that took the CUDA branch from one that skipped it.

The charge is not fixed even for one roster: with a one-rung menu this same
fixture grew 93,724,672 bytes over its baseline, with two rungs 157,847,552.
It follows what the encoder is asked to build, so no constant derived from
`unit_shapes` would track it.

## Behaviour change worth naming

A row whose plan plus measured floor exceeds its cgroup cap now refuses before
loading, where it previously refused at a mid-run guard check or not at all.

## Receipts

Every row is a PrismaBuild key at `--priority -10`, shown by its first twelve
hex characters. RED rows are at the tests-first commit `46984d8ce7`; GREEN rows
at head `0da7dff6f3`, and every GREEN row's `checkout_snapshot.parent` is that
head.

RED, before the fix:

- `30e99fadcc62` x86 dl380g10 rc=1, `test_tessera_selected_source.py`
  2 failed, 10 passed, 1 skipped. The new plan and baseline tests.
- `a0abbe6a6f00` sparklina rc=1, native GLM selected row on the GPU,
  1 failed, 11 passed. The peak-over-baseline check with headroom at zero.
- `8bdb05f40a26` x86 dl380g10 rc=0, the same GLM file on a CPU host:
  12 passed. The delta checks are CUDA-only, so x86 is not a RED host for
  them, recorded here as the negative result rather than implied.
- `034099017411` sparklina rc=1, parent `37d86e6263`, the deliberate
  undercharge probe described above, proving the steady-state assertion
  fails when the plan is too small.

GREEN at head, x86 dl380g10, all rc=0:

- `d83da8ff8fc4` test_tessera_selected_source 12 passed, 1 skipped
- `45da2f525b07` test_memory_management 16 passed, 1 skipped
- `b70a8f34b1e2` test_streamed_capture_admission 20 passed
- `8b2fec3bca3b` test_streaming_buffer_precision 7 passed
- `36b8302ee00a` test_tessera_stack_group_cli 1 passed
- `98d36f4dde1b` test_selected_source_authentication 26 passed
- `7fb51887a132` test_selected_source_authentication_ab 14 passed
- `609887cf2550` test_tessera_calibration_cache 34 passed
- `1663dcc33688` test_glm_native_wire_screen_plan 16 passed
- `d8a2dda833da` test_tessera_campaign_fanout 39 passed
- `429c07aa8535` test_architecture_doc 13 passed
- `e43e67d8d5f3` test_docs_staleness 6 passed
- `09fc1698bc00` test_glm_campaign_streaming 12 passed

GREEN at head, sparklina:

- `ed69def83163` rc=0, the native GLM selected row with a GPU, mem_gb 24,
  gpu_memory 16 GiB, run inside the pinned campaign container so the guard can
  read a namespaced `memory.max`: 12 passed, including the two-step assertion.
  The row asserts that `cuda_reserved_bytes` is above zero at the baseline and
  at the last checkpoint, so the CUDA branch demonstrably executed.
- `ad7d6731fc91` rc=0, test_selected_source_authentication_ab 14 passed on
  sparklina through the standard sharded runner.
- `c40e2deac94e` rc=1, the whole file list on CPU in the same container:
  twelve files pass, including the GLM row's CPU branch, and one,
  `test_selected_source_authentication_ab.py`, fails to collect with
  `ModuleNotFoundError: No module named 'tests.test_selected_source_authentication'`
  because that module imports its sibling through the `tests` package and this
  ad-hoc harness runs each file as its own pytest invocation. That is a harness
  artifact, not a code failure, and `ad7d6731fc91` above is the same file
  passing on the same host through the normal runner.

The native runs used a fixture checkpoint only. No GLM source tree, census or
live capture was read.

GREEN after merging main at `cb6c2b95a6`, which brought in #429's changes to the
same anchor loop. x86 dl380g10, all rc=0, every parent `cb6c2b95a6`:
`68560a122957` selected_source 12p/1s, `f2bffb7099f1` memory_management 16p/1s,
`efd5b2d1563f` streamed_capture_admission 20p, `b17788ff08ce`
streaming_buffer_precision 7p, `11b8dca764d3` stack_group_cli 1p,
`dc9c7fadc7e2` selected_source_authentication 26p, `7b79854e8c6c`
authentication_ab 14p, `5f325fe6021d` calibration_cache 34p, `86b056444f2a`
glm_native_wire_screen_plan 16p, `3d087a59f4ec` campaign_fanout 39p,
`6186f8c8db9e` architecture_doc 13p, `809d71926636` docs_staleness 6p,
`58116b00a504` glm_campaign_streaming 12p, plus #429's own two files on the
merged tree, `1964551303bf` glm_selected_anchor_profile 19p and `8901d4c74f14`
campaign_progress 5p. sparklina: `f13326118b17` rc=0, the native GLM selected
row with a GPU in the pinned container, 12 passed, so the merged loop still
runs two encode steps and the steady-state assertion still holds. Re-probed on
the merged tree with the same five terms under-priced, row `91f0ab9c54e3`,
sparklina, rc=1, an unreachable local commit on top of `cb6c2b95a6`: growths
`[165765120, 2351104]`, baseline 2,210,648,064, so the assertion still bites
after the merge.

## Owed

- The two gaps above, if a way to derive them appears: the export archive's
  pickle-and-directory metadata, and the producer's own encoder working set
  inside `encode_linear`.
- A bound on the first encode step. It is measured and attributed but not
  charged, and on a small roster it dominates the plan.

Closes #390

## Carries #452

At Astra's request this branch also carries #452 as a separate-commit
ancestor (`773502b708`, unchanged), so one CI run delivers both instead of a
prose-only full suite after each main merge. That change is +5/-3 in a single
docstring in `prismaquant/tessera_joint_aura.py`, already reviewed on #452,
and the file here is byte-identical to that branch's. No unreviewed code is
combined.

Closes #451


RobTand:
Lifting my own hold on this one. The reason I held it was that the native
gate used `pytest.xfail`, so a plan that undercharged would have reported
green. That is gone, and what replaced it discriminates:
`tests/test_glm_campaign_streaming.py:369` asserts
`max(growths[1:]) <= sum(plan['phases']['resident_anchors'].values())`, with
the per-anchor growth measured by the row's own guard across
`before_selected_anchor_batch` / `after_selected_anchor_batch`, and the first
step excluded and named as the runtime's one-time first-use cost rather than
folded into headroom. `:341` separately asserts the admission arithmetic the
row actually ran, `memory_bytes <= budget - baseline`, against a baseline
measured in that process.

Verified independently rather than taken from the report. The assertion bites:
probe row `034099017411`, sparklina, rc=1, parent `37d86e62635e`, a commit not
on this branch, with five `resident_anchors` terms cut to one byte; it fails on
the second step. The green rows resolve at this head, `checkout_snapshot.parent`
`0da7dff6f39c` on every one: `ed69def83163` sparklina with `accel=1`, which is
the row that actually executes the CUDA branch, and `d83da8ff8fc4`,
`09fc1698bc00` and the rest on dl380g10. CI is CLEAN.

Two terms remain documented gaps rather than derived, and the comments name
them: the export archive's pickle and directory metadata, and the producer's
own working set inside `encode_linear`, which is not traceable from this
repository and keeps its conservative bound.

Ready to merge.

-- claude-triage

RobTand:
Re-validated at `83554223c9195c4c7ffc94528aab6d7d9cba8d9a`, the merge of `origin/main` `273bd61dfa` (PR #432). Only the `docs/ARCHITECTURE.md` stamp header conflicted; no code did.

Native GPU row `f470795c7c4b`, sparklina, accel=1, rc 0, parent `83554223c919`: 12 passed in 40.91s. Worth recording how it had to be issued: `pbtest` with the bare cu130 venv all-skips `test_glm_campaign_streaming.py`, because that venv has no `transformers.models.glm5_next` and the module's `importorskip` fires at collection, so my first attempt `c7752cc63225` came back `1 skipped` and rc 1. Reissued through `pbrun` with the same container spec as the earlier `ed69def83163`-shaped row: image `sha256:9f9b9f05...`, the same `TESSERA_REPO` mount, demand cpu 4 / gpu 1 / mem_gb 24, 16 GiB GPU budget. So the `1 skipped` was a harness difference, not a regression, and the assertions still run.

x86 15/15 green at `--mem-gb 16`, every parent `83554223c919`: `eced53b7f094` `5e6a4b20a368` `b372a3ddf6b4` `c791f77ae705` `9ae6c9316e5f` `ef6cccc43141` `7770772a3cce` `98621f0283c4` `c6604aa4b52d` `c3f18b126fed` `38cd66b7a945` `8f30a13b31b1` `d1938002d332` `8d10065e11bb` `989801c8b4d6`. The earlier run at the 3 GiB default lost three shards to `memory_limit_oom`; that set needs 16 GiB.

The `cb6c2b95a6` receipts above still stand: I verified all sixteen independently, every parent `cb6c2b95a60f`, rc 0, priority -10, generation `332fa43447a2`, and the undercharge probe `91f0ab9c54e3` rc 1 on the unreachable `8f6c559c0de8`.

PR #427, opened 2026-09-08
Keep complete GLM Hessian handoff bounded through export

### LINKED ISSUE #425: Carry bounded canonical Hessian references through full-roster export
The full GLM candidate union has 36,423 units and about 1.67 TiB of logical FP32 H. `tools/dispatch_tessera_campaign.py:943-978 (reviewed integration f3cd7c7af)` loads each row H payload into one accumulated dictionary; `prismaquant/tessera_export_lane.py:1052` also eagerly loads the complete H handoff. This cannot close the full-roster export path within a 104 GiB worker contract.

Extend the existing canonical capture owner and export-input handoff with opt-in bounded H reference metadata. Preserve the existing H seal, add canonical manifest/census binding from row prices through uniformity checks, allocation and producer priced inputs, and verify actual H only at consumption. Merge must verify the exact selected row rosters without loading H. Legacy eager files retain their contract; unsupported sampled materialization must refuse before work.

Severity P1. The active campaign is fixing this rather than deferring it. Acceptance includes negative/control tests through PrismaBuild, exact cross-repo producer source binding, receipt/source audit and architecture updates. This does not lower selected resident H/X accounting or qualify native full-stack fit.

Producer counterpart: https://github.com/RobTand/tessera/issues/428.

RobTand:
Merging the complete GLM candidate roster previously accumulated every calibration Hessian into one eager `.pt` object. The new opt-in handoff joins canonical metadata references and row seals instead; the reviewed Tessera reader verifies one bounded original H file/tensor when consumed. Pricing, row merge, allocation and exporter inputs preserve the same complete-capture/census binding. Unsupported sampled materialization refuses references at intake. Legacy eager inputs remain supported.

Synchronizes the development and serving dependency pins to reviewed Tessera `9d2314819f02` (RobTand/tessera#429), which supplies the public reader and nonblocking regular-file refusal. The exact raw contract changes only LFM construction output sizes; the allocator admission answer, serving-cell and native-extension tables are unchanged. Version 0.1.0 remains advisory, and no TP2 runtime cell or production eligibility is promoted. New pricing/export must use a consistent producer source seal.

Validation through PrismaBuild: initial handoff suite 200 passed / 2 skipped; final integrated handoff, merge, allocation, materialization, pin and architecture checks **239 passed / 3 skipped**, CPU only. Six touched implementation modules compiled. Root independently verified terminal exit/cleanup, canonical CAS receipts, actual payloads and source bundles; the final test source matches `1228c3523166` apart from generated closure metadata. Skips cover the legacy producer-absence case, explicit native writer measurement and absent immutable v5 publisher contract. Native handoff performance and full GLM quality/fit remain unmeasured. Full CI is pending.

Commands, exact producer archive hash, receipts and source audits: `experiments/measurements/bounded-hessian-handoff-20260908/README.md`. Includes a separate prose-only correction of the obsolete PENDING-pin description.

Closes #425. Depends on RobTand/tessera#429. Refs #388.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,16 @@
 As of: 2026-09-08 · `integrate/glm-bounded-handoff-20260908`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-08, `integrate/glm-bounded-handoff-20260908`) for the
+synchronized development/serving dependency pin to Tessera `9d2314819f02`
+(#429), required by bounded canonical H references. The exact packaged
+contract SHA is `a688f8de244f936ec3a63a782e20af7985733e7a6fb0b4b981b5fe4c44112212`.
+Its only structural change from the prior pin is LFM construction output
+sizes; the allocator admission answer, serving-cell and native-extension
+tables are unchanged. Version `0.1.0` remains advisory with
+`version_is_release=false`. The explicit research TP2 controls do not promote
+a runtime cell. New priced wires bind the final producer source identity.
+
 Re-stamped (2026-09-08, `fix/bounded-export-handoff`) for the opt-in
 canonical Hessian reference handoff. Selected reuse may declare
 `--export-hessian-reference-policy` with explicit metadata/file/H byte caps.
@@ -19,7 +29,7 @@ Legacy `.pt` handoffs keep their eager behavior. Sampled selected-wire
 materialization explicitly refuses reference inputs until it has a bounded
 reference union; complete priced wires remain the supported reference scope.
 This adds no cache or scheduler, changes no row resident-prefetch estimator,
-serving pin, wire format or ship gate. It requires the reviewed producer reader
+wire format or ship gate. It requires the reviewed producer reader
 and a fresh producer source identity for new prices; old wire identities are
 not reused under the changed producer hash. CPU contract validation does not
 establish native throughput, GPU residency or full-model serving performance.
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,27 @@
 As of: 2026-09-08 · `integrate/glm-source-derivative-20260908`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-08, `fix/bounded-export-handoff`) for the opt-in
+canonical Hessian reference handoff. Selected reuse may declare
+`--export-hessian-reference-policy` with explicit metadata/file/H byte caps.
+The existing row writer emits `hessian_capture.references.json`, binding the
+complete canonical capture and exact census, full-census counts, per-unit H
+commitments and the unchanged old row capture SHA. Dispatcher merge verifies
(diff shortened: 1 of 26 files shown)


RobTand (commented):
Reviewed at `47dff28e1c2b`. **VERDICT: MERGE.**

## Pin, verified independently

Fetched `RobTand/tessera` into a scratch repository. `9d2314819f02` is an
ancestor of `origin/master` (tip `b9cda5b03`, the merge of its #429), and
`git cat-file -p 9d2314819f02:src/tessera/serving/runtime_contract.json |
sha256sum` is `a688f8de244f...`, the value both pin files now carry. A
structural diff of the two contract documents (parsed, not textual) returns
exactly one difference: `construction/architectures[1]/output_sizes` added on
the `Lfm2MoeForCausalLM` row. Nothing under `contract_answer`'s roster moved,
so the "admission answer unchanged" claim holds by comparison rather than by
assertion.

## The handoff

The design is sound and the identity claim is the right one to have tested.
`tests/test_tessera_hessian_reference_handoff.py:46` asserts the reference
row's digest equals `hessian_capture_sha256` over the eager dict, so the seal
that binds an allocation is the same object before and after the change; the
`torch.save`/`torch.load` monkeypatches make "no eager materialization" a
property the tests can fail on rather than a claim in prose.

The metadata gate is complete, not partial. `tessera_export_lane.py:1240`
binds the canonical manifest and census SHAs, the identity comparison at
`:1257` still runs against the descriptor's provenance, and the recomputed
`capture_sha256` at `:1273` binds the per-unit content commitments. What is
deferred is only the proof that the files still hash to those commitments,
which `ReferenceHessians.__getitem__` pays per unit through a held descriptor.
`tessera_calibration_cache.py:766` publishes through the reader, so a
descriptor with an empty or foreign roster is refused at write time by the
producer's own validation rather than at first use.

Fail-closed paths check out: `tessera_materialization.py:67` refuses at
request intake, `dispatch_tessera_campaign.py:975` refuses a mixed merge, and
`:937` asserts the merge read no tensor bytes.

## Receipts

`b1c1dd4dc4d3` done rc=0 parent `1228c3523166`, host dl380g10, priority -10.
Fetched its snapshot bundle from the CAS blob and diffed against this head:
the tested tree differs only by `.pbrun-closure.*.json` and the three audit
JSONs plus README text added afterwards. Every runtime module tested is
byte-identical to what is proposed. `e77e2f2fe875` (200 passed) and
`740c213d5bd5` resolve done rc=0; `4915c95804a9` and `8326b6e31f23` are the
declared expected failures, rc=1 as PB records a failing test.

Worth noting for the record: the earlier rows ran against Tessera
`615e1c0302`, and the only delta to the pinned commit in the consumed module
is the `O_NONBLOCK` open in `_HeldFile`. The final row used the pinned
producer archive, so the pin is covered.

## One small thing, not blocking

`hessian_payload_verification` is returned by
`require_priced_export_inputs` but never stamped into the preflight JSON, so a
reader of a shipped artifact infers deferral from the `v2` schema and the
binding rather than reading it. Stamping the field would make the deferral
legible to something that is not a schema-version comparison.

PR #431, opened 2026-09-08
Let the GPU grant decide the container attachment, not a remembered flag

### LINKED ISSUE #430: tessera_campaign_container attaches the GPU by default, so a row that reserved none still gets the device
`tools/tessera_campaign_container.py` maps the whole GPU into every container
it launches unless the caller remembers `--cpu-only`. A PrismaBuild row that
reserved no GPU therefore runs with the device attached, and PrismaBuild's GPU
tokens are never spent for it.

## The line

`main` passes `with_gpu=not args.cpu_only` (`tools/tessera_campaign_container.py:137`),
and `docker_command` turns that into `["--gpus", "all"]`
(`:72`). The default is `store_true` off, so withholding the GPU is opt-in and
attaching it is what happens when nobody thinks about it. There is no path that
derives the flag from the row's declared resources.

## Observed, not inferred

PrismaBuild row `c40e2deac94e` on sparklina: `resources` is
`{'cpu': 6, 'mem_gb': 24}`, no `gpu` flag, and its spec contains no
`--cpu-only`. So that row held no GPU reservation and ran
`docker run --gpus all` anyway.

The power reading is consistent with the device being initialised rather than
merely mapped. Two no-GPU container rows read about 21 W against a 13 W idle
floor (`c40e2deac94e` 20.6 W, `44714b5709b9` 21.0 W) while an actual GPU row on
the same box read about 17 W. Those rows overlapped and `pbrun`'s power line is
box-level, so the attribution is not established; the reservation gap above is,
and it does not depend on the power reading.

## Why it matters

Two separate things, and the second is the one that costs a result.

Admission accounting: PrismaBuild's GPU tokens exist so that two jobs cannot
both think they own the device. A container row that takes the device without a
reservation is invisible to that arithmetic, so a GPU-reserving row can be
admitted beside it.

Measurement isolation: an A/B or a power measurement running next to such a row
is not isolated, and on GB10 the instrument we trust is power against the
envelope (principle 15). A second process holding the device is exactly the
confound that reading cannot distinguish from the effect being measured.

## The fix I did not make

The safe polarity is to derive the flag from the row's declaration rather than
flipping the default under a live campaign: pass the GPU when the job reserved
one, and refuse rather than silently attach when a spec asks for a device the
row did not reserve. The live GLM capture launches through this tool, so
changing its default while that runs is not something to do unasked.

Setting `CUDA_VISIBLE_DEVICES=''` in the environment is not a substitute. It
hides the device from CUDA inside the container; it does not stop the nvidia
container runtime from attaching and initialising it.

Found while verifying a delegated worker's receipts, which is also where the
power observation comes from.

-- claude-triage

RobTand:
`tools/tessera_campaign_container.py` mapped the whole GPU into every container
it launched unless the caller remembered `--cpu-only`, so a PrismaBuild row that
reserved no GPU ran with the device attached and no GPU token spent for it.
Observed, not inferred: row `c40e2deac94e` on sparklina has
`resources={'cpu': 6, 'mem_gb': 24}`, no `gpu` flag, and no `--cpu-only` in its
spec.

Two things follow, and the second is the one that costs a result. PrismaBuild's
GPU tokens exist so two jobs cannot both think they own the device, and a row
that takes it without a reservation is invisible to that arithmetic. And on
GB10 the instrument we trust is power against the envelope, so a second process
holding the device is exactly the confound that reading cannot separate from
the effect being measured.

## The grant decides now

`gpu_attachment` reads the declarations that already exist rather than
inventing a rule. `--gpus all` is withheld when any declaration withholds it:
an explicit `--cpu-only`; an empty `CUDA_VISIBLE_DEVICES` in the action
environment, which is exactly what `pbrun` sets when it granted no GPU slots;
or an empty `CUDA_VISIBLE_DEVICES` in the spec's own `env` block. Otherwise the
device is attached, as before.

The two checks are not redundant with the variable itself. An empty
`CUDA_VISIBLE_DEVICES` hides the device from CUDA inside the container but does
not stop the runtime attaching and initialising it, so a row that reserved no
GPU could still own one while PrismaBuild's GPU tokens stayed unspent.

Unset is not a declaration: an interactive run outside `pbrun` has no grant to
read and keeps the behaviour it had, where `--cpu-only` remains the way to
withhold the device.

The launcher's existing JSON line now carries `gpu_attached` and
`gpu_decision`, so the choice and the declaration that made it are readable
rather than inferred from argv.

Deliberately not done: mapping the granted device list into `--gpus device=...`.
That needs `CUDA_VISIBLE_DEVICES` indices reconciled with Docker's enumeration,
which is a second change with its own failure mode. A granted row still gets
`--gpus all`, exactly as today.

## Source

Two files against `origin/main`, plus the `docs/ARCHITECTURE.md` stamp:

- `tools/tessera_campaign_container.py`
- `tests/test_tessera_campaign_container.py`

## Receipts

RED at `3c057668ca`, receipt `637ae8b88cf8`, host dl380g10, row failed rc=1:
`5 failed, 9 passed`. The behavioural one asserts on the real exec line:

> AssertionError: the exec line still maps the whole GPU into a container whose
> row reserved none (#430): `docker run --rm --gpus all --ipc=host --user 1000:1000`

`test_main_withholds_the_device_from_a_row_that_reserved_none` is the one that
proves the defect rather than the absence of a helper, and
`test_a_granted_row_still_gets_the_device` is the control that keeps the fix
from being "never attach anything", which would break every campaign row.

GREEN at `63fb1c854a`, all host dl380g10, `row=done rc=0`, `--priority -10`,
every row's `checkout_snapshot.parent` is `63fb1c854a3fa3f1a4f6216dad9137d7f1c819f3`:

| receipt | target | result |
|---|---|---|
| `1b2ef61f7cb3` | `tests/test_tessera_campaign_container.py` | 14 passed, 14 warnings in 5.61s |
| `8b139bbdbf70` | `tests/test_architecture_doc.py` | 13 passed, 14 warnings in 5.54s |
| `50febbf0d82d` | `tests/test_docs_staleness.py` | 6 passed, 14 warnings in 4.90s |

The two doc gates are there because this branch re-stamps `docs/ARCHITECTURE.md`;
the container test is the behavioural gate for the change itself.

## Green again at the head

`58edcbb216` merges `origin/main` at `711cb924fe` (#439). The merge touched one
file, `docs/ARCHITECTURE.md`, and only its stamp block: this branch's stamp is
kept first under its own `As of:` header, then main's
`fix/glm-model-bound-wikitext-inputs` stamp, newest first as the header says.
The merge also brings main's unrelated gold and input tooling into the branch,
as any merge would; what is unchanged is this branch's own surface, the
container implementation and its test, which are byte-identical to
`63fb1c854a`.

All three gates re-run at `58edcbb216`, host dl380g10, `row=done rc=0`,
`--priority -10`, interpreter `/home/rob/venvs/pq-cpu312/bin/python`, every
row's `checkout_snapshot.parent` is
`58edcbb216cc159bc4cd618c088595a92cec2904`:

| receipt | target | result |
|---|---|---|
| `140ed8261eca` | `tests/test_tessera_campaign_container.py` | 14 passed, 14 warnings in 6.42s |
| `e2b63771ad14` | `tests/test_architecture_doc.py` | 13 passed, 14 warnings in 6.52s |
| `0365b0257c36` | `tests/test_docs_staleness.py` | 6 passed, 14 warnings in 5.42s |

Same counts as at `63fb1c854a`, so the merge changed nothing this branch
depends on. Row `30a18b3bbb01` is an independent run of the same three files
together at the same parent: `33 passed`, which is 14 + 13 + 6.

## What this does not establish

- No measurement that the attached device drew power or perturbed anything. Two
  no-GPU container rows read about 21 W against a 13 W idle floor while a real
  GPU row read about 17 W, but they overlapped and `pbrun`'s power line is
  box-level, so that stays an observation. The reservation gap is what this
  fixes and it does not depend on the power reading.
- No container was launched in these tests. `os.execvp` is monkeypatched and the
  assertion is on the argv, so this proves what Docker would be handed, not what
  the runtime then does with it.
- The live GLM capture reserves `{'cpu': 6, 'gpu': 1, 'mem_gb': 104}`, so it is
  a granted row and this change does not alter what it gets. It does change
  future launches of any row that reserved none.

Closes #430.


RobTand:
Contract note added and validated, head `9e32416d50a8`.

`docs/ARCHITECTURE.md` now carries a stamp for the launcher change, and the "As of" line names this branch. It records the three declarations that withhold `--gpus all`: an explicit `--cpu-only`, an empty `CUDA_VISIBLE_DEVICES` in the action environment, which is what `pbrun` sets when it granted no GPU slots, and an empty `CUDA_VISIBLE_DEVICES` in the spec's own `env`. It says why the variable is not the check on its own, since an empty value hides the device from CUDA inside the container without stopping the runtime attaching and initialising it, which is how a row that reserved no GPU could still own one while PrismaBuild's tokens stayed unspent. It records what did not change: an unset variable is not a declaration, so a run outside `pbrun` keeps its behaviour and `--cpu-only` stays the way to withhold the device there. It names `gpu_attached` and `gpu_decision` on the launcher's JSON line, and states that no serving pin, wire recipe, cost currency, format menu or ship gate moved.

No test restates the prose; the existing `tests/test_tessera_campaign_container.py` already covers the behaviour, and `test_architecture_doc.py` and `test_docs_staleness.py` cover the mechanical half of the stamp.

Receipts, x86, priority -10, all rc 0, parent `9e32416d50a8`, generation `332fa43447a2`: `d77a4352d323` (test_tessera_campaign_container, 14 passed), `70085e4634fc` (test_architecture_doc, 13 passed), `97aa440c33eb` (test_docs_staleness, 6 passed).

The previous head `63fb1c854a3f` (merge of main `273bd61dfa`) was green on the same three files: `1b2ef61f7cb3`, `8b139bbdbf70`, `50febbf0d82d`.

PR #432, opened 2026-09-08
Constrain packed GLM research allocation and freeze the checkpoint producer

RobTand:
GLM's packed resident research route needs routed expert allocations that the existing E4M3 decoder can execute, and pricing/export need a fixed producer source before any new wires are cached. Add the opt-in `glm_packed_research_sm121` profile through the shared family filter, allowing legal E4M3 K1 rungs and plain BF16 for routed experts while retaining the inherited dense menu and existing shape/source/context gates.

Synchronize both Tessera pins to reviewed commit `07ad344c3275bb2fa7ce2432f93d89945d66f4c2` from merged Tessera #431, which carries the explicit packed execution declaration through export, partition identity and ordinary plugin config. The packaged contract SHA is unchanged. The profile remains emulation-only with no export lane and TP1; this does not qualify full-model serving or TP2.

PB CPU validation: profile/candidate suite 118 passed; architecture/docs 19 passed; cross-repository H handoff, sidecar, pin, contract, profile and candidate checks 194 passed with two explicit skips, plus compilation. A final identity action verified all 1,179 staged producer files and the live pin/contract agreement. Root checked actual terminal status, cleanup, CAS hashes, payloads and executed source snapshots. Commands and audits are in `docs/experiments/glm_packed_source_integration_20260908/`.

No new calibration forward, encoding, quality, fit or speed measurement. Original capture reuse remains intact; complete export and served measurements remain pending.

Final full CI on `21ae7d28b7`: 7,131 passed, 200 skipped, 3 expected failures and 192 passing subtests; import and linked-issue checks passed.

Closes #433. Refs #388. Depends on merged RobTand/Tessera#431.

PR #436, opened 2026-09-08
Bind explicit stock topology in offline gold instruments

### LINKED ISSUE #434: Gold runners cannot express stock TP2 multi-node topology
The offline gold KL and PPL tools hard-code `tensor_parallel_size: 1` in `tools/measure_vllm_full_kl.py:189` and `tools/measure_vllm_wikitext_ppl.py:357`; neither CLI accepts the stock multi-node fields. A requested dual single-GPU Spark measurement cannot express its TP2/2-node topology. HTTP gold already supports direct PPL and declared top-K KL, but does not replace exact full-vocab KL.

P1 measurement configuration gap. Fixing on `fix/glm-gold-topology`: shared closed argparse/LLM mapping for TP size, nodes, rank-0 coordinator, rendezvous, MP backends and the optional MoE backend. Preserve omitted-argument TP1 behavior, bind helper bytes and configured topology in existing gold provenance, and document separately launched stock headless workers. No frozen `prismaquant/` source, Tessera source, serving qualification, scheduler or default changes. PB regression `abce7b0beecaba6eadd222adbf48ef578de68751069460c600cae513c8687262` reproduced four failures (both CLI refusals and both hard-coded TP1 mappings). No native full-model execution is claimed. This issue covers instrument plumbing only; the broader GLM delivery remains open.

RobTand:
The offline full-vocabulary KL and WikiText PPL instruments hard-coded TP1, so a declared two-Spark run could not configure the stock engine topology. Both tools now share validated TP/node/rendezvous/MP/MoE options, preserve omitted TP1 behavior, reject remote coordinator ranks before loading, and record the configured topology plus helper bytes in existing provenance. Remote workers remain separately launched stock headless vLLM processes and need their own runtime evidence.

The accompanying GLM handoff binds six research variant inputs, existing plan/export/gold interfaces and the earliest whole-model anchor gate. It records two unresolved dependencies explicitly: calibrated packed cached export currently refuses its required Hessian input, and the fixed gold teacher needs authenticated source/derivative support with fitting overlap checked. No frozen PrismaQuant pricing package, Tessera producer, menu, pin or serving qualification changes are included.

Validation: four public CLI/stub-engine regressions fail before the fix; PrismaBuild passes 102 instrument/provenance tests plus 19 architecture/staleness tests and compile checks, zero skips. Actual terminal/CAS/result/source audits are checked in under `experiments/measurements/glm-downstream-handoff-20260908/`. No native full-model or TP2 GPU measurement is claimed. Built on the validated #432 integration; the PR now targets main.

Closes https://github.com/RobTand/prismaquant/issues/434

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,17 @@
 As of: 2026-09-08 · `integrate/glm-packed-source-20260908`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-08, `fix/glm-gold-topology`) for explicit stock-vLLM
+topology in the offline gold KL/PPL tools (#434). Shared closed arguments
+forward TP degree, node count, rank-0 rendezvous, MP backends and an optional
+MoE backend to the existing LLM construction; omitted arguments retain TP1.
+The gold coordinator does not launch remote nodes. Multi-node runs require
+separately launched stock headless workers, explicit compatible MP settings,
+and per-node runtime evidence. Configured topology and the helper's source
+bytes join existing gold provenance. Full-vocab final-position KL remains
+distinct from HTTP top-K KL. This instrument change touches no frozen pricing
+package input, Tessera producer, serving qualification or performance claim.
+
 Re-stamped (2026-09-08, `integrate/glm-packed-source-20260908`) for the
 synchronized development/serving dependency pin to Tessera `07ad344c3275`
 (Tessera #431). The producer can bind an explicit packed research execution
--- a/experiments/measurements/glm-downstream-handoff-20260908/README.md
+++ b/experiments/measurements/glm-downstream-handoff-20260908/README.md
@@ -1,0 +1,308 @@
+# GLM downstream handoff — 2026-09-08
+
+This is the bounded handoff for six research artifacts: one Spark and two
+Sparks, each with accuracy, prefill and balanced selection intent. It binds
+existing interfaces and identifies missing inputs; it does not report six
+completed artifacts or production qualification. `handoff-draft.json` is an
+unsealed intake record with explicit nulls, not an accepted collection receipt.
+
+Source basis: PrismaQuant `21ae7d28b7d313654beda6b4e622d732d4276a34`, gold
+instrument fix `0661c3e7eb`, and frozen Tessera producer
+`07ad344c3275bb2fa7ce2432f93d89945d66f4c2`. No `prismaquant/` or Tessera source
+changes are part of this handoff. The original fit draw, producer identity,
+family menu, release pin and production ship gates remain owned by the root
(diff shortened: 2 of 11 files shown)


RobTand (commented):
`tools/gold_engine_options.py:20-56`, `tools/measure_vllm_full_kl.py:192`, `tools/measure_vllm_wikitext_ppl.py:360`.

The tests prove the eight options reach `LLM(**kwargs)`, but they prove it through a stub, so they cannot say the engine accepts them. Checked that separately against a stock tree on this box (`site-packages/.../engine/arg_utils.py`): `tensor_parallel_size`, `nnodes`, `node_rank`, `master_addr`, `master_port`, `distributed_executor_backend`, `data_parallel_backend` and `moe_backend` are all real `EngineArgs` fields, and `LLM.__init__` forwards `**kwargs` there. The forwarding is not fictional.

The `mp`-only choice agrees with the engine rather than guessing at it. `config/parallel.py:846-854` permits `nnodes > 1` only with `mp`, `uni` or `external_launcher` and refuses `ray`; `:804-805` selects `mp` for CUDA multi-node; the `distributed_executor_backend` docstring at `:230-235` says "To use mp you must also set nnodes". So refusing `ray` at the CLI matches what the engine would refuse later, and refusing it earlier is the better place.

`tp % nnodes` matches the engine's own arithmetic: with DP and PP at 1, `world_size == tp`, and per-node width is `world_size // nnodes_within_dp` (`:643-653`), so nnodes must divide tp. `type(value) is not int` correctly rejects `True`, which `isinstance` would have taken.

Two places where the gate is narrower than the runtime. Neither blocks, both are worth naming so a later run does not read the refusal as a platform fact:

- `uni` and `external_launcher` are refused although `:846` allows them with `nnodes > 1`. `external_launcher` is the usual torchrun path, so a two-Spark run launched that way would be refused by the instrument and not by the engine.
- `moe_backend` is limited to `auto` and `triton`; `config/kernel.py` also names `deep_gemm` and others.

One producer-side requirement that is not derived: multi-node requires an explicit `data_parallel_backend`, which the engine consults only when DP exceeds 1 (`:818`). Harmless, and it costs a caller nothing to pass it, but it is our constraint rather than the runtime's.

Defaults are held exactly: `test_default_llm_options_preserve_single_node_behavior` asserts the whole kwargs dict, so an omitted flag cannot drift the TP1 path. Adding `tools/gold_engine_options.py` to `_GOLD_PRODUCER_TOOL_FILES` puts the topology code inside the gold identity, which is the right side of that boundary. `docs/ARCHITECTURE.md` is re-stamped in the same commit, and the stamp says plainly that this is an instrument change with no pricing, producer or serving-qualification effect, and that remote workers still need their own evidence.

The branch is BEHIND main and needs a merge before it lands. I did not re-run your four pre-fix failures or the 102/19 PB rows.

VERDICT: MERGE AFTER merging main

PR #442, opened 2026-09-08
Observe compatible batched anchors in the existing profiler

### LINKED ISSUE #441: Selected-anchor observer omits compatible batched encoding
Selected-anchor observation refuses the already supported compatible-batch campaign (`experiments/glm_full_capture_profile.py:274` at 21ae7d28b7), and instruments only `_measure_anchor` (`:306`). The campaign calls `_measure_anchor_batch` for compatible expert batches. Removing only the argument guard would leave those encodes unobserved.

P2: observability gap blocking the first GLM narrow-band qualification. Fix on the existing isolated branch: observe both original methods in one call sequence and preserve exact call/output ownership. CPU regression reproduces the refusal (4 failures, 19 passes); fixed native suite passes 36 tests with actual CUDA events. A CUDA-only activity option preserves separate CPU sampling; the earlier oversized whole-call traces remain negative evidence, and no performance improvement is claimed.

This finding is fixed in this session rather than deferred. Native GLM encoder fit and trace-size evidence remain a subsequent campaign gate; the small native test qualifies only instrumentation.

RobTand:
The selected-anchor profiler rejected compatible expert batches and wrapped only the scalar measurement entry point. It now observes both existing entry points in one call sequence, records batch membership, preserves the original arguments/results, and restores both methods after failure.

An optional `--anchor-cuda-only` mode retains the separate Python sampler and both machines’ Netdata series. Exported trace caps remain distinct from live-memory bounds; the earlier oversized GLM traces are retained as negative evidence. No encoder or performance change is claimed.

Validation through PrismaBuild: regression 4 failed/19 passed before the fix; fixed CPU observer 23 passed; capture/telemetry 11 passed with one native-only skip; combined SM121 native suite 36 passed, no skips. The native smoke produced an authenticated CUDA kernel trace. Actual receipts, source snapshots, launcher corrections and limits are recorded in `docs/experiments/glm_batched_observer_20260908/README.md`. Merging current main changed no observer, campaign or test dependencies relevant to those results.

Closes #441.


RobTand (commented):
`experiments/glm_full_capture_profile.py:190-205,281,305-330` ·
`tests/test_glm_selected_anchor_profile.py:223-282`

Checked the two things a wrapper like this usually gets wrong, and both hold.

**Restoration is correct on partial setup.** The `setattr` loop moved inside the
`try` (`:322-325`), and the `finally` restores from `originals`, captured before
any patching. So a failure on the second install still puts the first back.
`test_entrypoint_wraps_batch_and_restores_both_after_campaign_error:262` asserts
it by identity on both names, and that the wrappers were live when
`campaign.main` ran.

**Batch membership cannot silently degrade.** `record.update(qnames=..., batch_size=...)`
reads `kwargs['qnames']`, which would report a batch of one if the campaign ever
called positionally. It cannot: `tessera_campaign.py:551` is
`def _measure_anchor_batch(*, qnames, ...)` and `:358` is likewise keyword-only,
and the call sites at `:4733/:4738` pass by keyword. The shared counter is one
sequence over both wrapped names because both wrappers close over the same
observer, which is what `call_index_scope` claims and what
`test_batch_and_scalar_share_call_windows_and_preserve_outputs:227` proves with
`anchor_calls == 3` against two profiled windows. That test also asserts output
and input identity (`is outputs`, `is weights`), not equality, so it would catch
a wrapper that copied.

`--anchor-batch-size` is `type=int, default=1` (`:277`), so the widened
`>= 1` guard cannot meet `None` and raise `TypeError` past the
`(AttributeError, ValueError)` handler.

**Receipts verified by key.** All four match the table, host, rc and parent:
`9f6b24b22d74` dl380g10 rc=1 at `943a86b417`; `c4dff0d424b8` dl380g10 rc=0 at
`0f2400fe4b` (23 passed); `167c16d35a18` dl380g10 rc=0 at `0f2400fe4b`
(11 passed, 1 skipped); `01f4a1bcd825` sparklina rc=0 at `f4f6470e2f`
(36 passed, 10.05s). Between `f4f6470e2f` and the head `2cf67c1a6641` only docs
and evidence files moved, no source and no tests, so the native row covers the
head.

Two notes, neither blocking.

The two CPU rows sit at `0f2400fe4b`, which has 12 test functions against the
head's 13. The thirteenth is `test_native_cuda_only_batch_trace_has_actual_events`,
added with the commit the native row ran, and it is native-only, so nothing is
uncovered. Saying that in the table would keep a later reader from re-deriving it.

The scalar branch of the new record line, `[kwargs.get('qname')]` with
`batch_size=1`, is executed at `:249` but never asserted. `:227` profiles calls
`(0, 2)`, both batches, so no scalar record is checked. One assertion in the
existing test would close it.

VERDICT: MERGE

PR #444, opened 2026-09-08
Bound selected-anchor CUDA collection with an explicit time window

### LINKED ISSUE #443: Selected-anchor full-call CUDA traces exceed qualification evidence limits
P2: selected-anchor qualification can complete encoding and still lose all native profile evidence because the observer only limits exported bytes after a full call.

At `experiments/glm_full_capture_profile.py:260`, collection spans the original encoder call; the subsequent trace-size check cannot bound collection duration. Dense qualification action `7c417f647318` exported 2,318,078,289 and 2,312,576,791 byte traces against a 512 MiB cap and ended exit 1. Expert action `ab1c32b1a4c8` rejected its first 9,222,546,186 byte trace; it was withdrawn with completed batch journals retained.

Fixed on the discovery branch with an explicit finite CUDA-only collection window through PyTorch dynamic collection, preserving original exactly-once execution and exported-byte refusal. Regression first: eight expected failures. CPU:31 passed/two CUDA skips. Native GB10:45 passed/zero skips; actual trace contains the first kernel and excludes later work. Full dense qualification of the fixed observer is still running. No throughput or quality improvement claimed.

RobTand:
Selected-anchor traces exceeded the export cap after full encoder calls, making the dense qualification fail despite completing its pricing. Add explicit `--anchor-profile-seconds` to stop initial CUDA collection while preserving the original encoder result and exactly-once exception handling.

Actual dense qualification before: both traces exceeded 2.3 GB and PB action `7c417f647318` failed observation. After: `9f46d50b9abd` completed successfully with 1.7 MB and 72.4 MB traces; all 1,030 quality/byte-cost rows match, with only 14 encoding times differing. These ran on different GB10 hosts, so this is instrument validation, not a throughput A/B.

PB tests: CPU31 passed/two CUDA skips; native45 passed/zero skips. The timed native fixture preserves both BMM outputs and excludes the later kernel. Eight initial tests refused the previously unsupported option; the failed real dense action is the observed defect. The deadline is cooperative and is not a live profiler-memory guarantee. Actual CAS, source and artifact audits are committed.

Stacked on #442; no pricing package, capture, producer, or serving contract changes. Fixes #443.


RobTand (commented):
`experiments/glm_full_capture_profile.py:189-238`

The mechanism holds up. I went looking for the failure-during-failure case,
where the encoder raises and the toggle also fails, expecting the
`collection_window` finally to displace the encoder's exception. It does not:
`:284-285` re-raises `original_error` with its own traceback whenever it is
set, so the `RuntimeError` from the window is absorbed and exactly-once holds.
The toggle failure is still legible afterwards, because `stopped_by` is
`toggle_failed` inside `record['collection_window']`. That combination has no
test, and it is the one worth having, but the code is right and the record
does not go quiet, so this is not something to hold the PR for.

Two things I want to affirm rather than flag, because both look like
omissions until you see why they are not. The join at `:230` is unbounded, and
that is the correct trade: a bounded join is a stopper that can outlive the
`with` and touch a profiler that is being torn down, which is the failure this
is trying to avoid. And the CUDA-only contract at `:167-168` is load-bearing
rather than conservative, since CPU collection state is thread-local and a
Kineto-wide toggle from a second thread would not mean what it says.

`type(window_seconds) not in (int, float)` at `:164` rather than `isinstance`
is what makes `True` refuse, and the parametrisation asserts it. Deliberate,
and worth a comment so nobody "fixes" it into `isinstance` later.

One thing about the evidence, which changes the wording rather than the code.
The body says "eight regression failures before the fix". Read the receipt:
all eight are `TypeError: AnchorObserver.__init__() got an unexpected keyword
argument 'window_seconds'`. That shows the tests are new and the constructor
lacked the parameter. It does not show the old behaviour failing, because the
old behaviour cannot be reached through a signature that rejects the call.
The defect is genuinely demonstrated, but by the production failure the README
cites, action `7c417f6473182bed`, whose two traces exported 2,318,078,289 and
2,312,576,791 bytes past a 512 MiB cap. That is the regression evidence; the
eight `TypeError`s are scaffolding. Saying so keeps the receipt honest and
costs nothing.

The native run is what carries the claim: 45 passed, zero skips, Torch
2.13.0+cu130 in the qualified container, requested 0.25 s against a recorded
0.250759305 s, a 12,584-byte trace holding the first kernel and not the later
one. The CPU rows skip that fixture, so the toggle path is only really
exercised there, which the receipt states rather than blurs.

VERDICT: MERGE

PR #419, opened 2026-09-08
Re-assert a layer's prefetch before install instead of trusting a private set

RobTand:
Closes #403.

## What was wrong

The exact layer-major visitor (`StreamedCausalLM.visit_layer_batches` with v2 `layer_major` storage) kept a private `scheduled` set and skipped `schedule_prefetch` for any layer already in it. The runner already owns residency and its `StreamingContext.schedule_prefetch` is idempotent: it returns None for a hot layer, the held future for a read in flight or delivered and unclaimed, and submits a fresh read only when nothing is held. So the private set could only veto the one call that heals a speculation the runner no longer holds.

#403 describes the failure as the LRU evicting a prefetched layer. That case is already tolerated: `_prefetch_worker` keeps the delivered future in `_inflight` until `ensure_loaded` claims it, and `ensure_loaded` re-inserts an unretained delivery (`prefetch_delivered_unretained`). The refusing cases are the ones where the speculative call returned None with nothing in flight:

- the layer was hot when speculated (`layer_cache.peek` true, no future) and the LRU evicted it before its turn;
- the pressure floor refused the speculative read, in `schedule_prefetch` or in `_prefetch_worker` (which pops its own future).

Both then hit `install(require_prefetched=True)`, which refused with no recovery path except a rerun at a larger budget. `_top_up_prefetch` heals exactly this hole for the default traversal; the exact visitor disables it (`prefetch_following=False`) and added its own veto.

## What changes

- `prismaquant/cost_streaming.py`: the visitor speculates each layer once ahead of its turn (`prefetch_lookahead`) and re-asserts it once, immediately before install, through the runner's idempotent `schedule_prefetch`. One bounded retry per layer; the refusal stays fail-closed for everything else and the runner's message already names the layer. Layers whose re-assert cost a fresh read are recorded as `runner.layer_major_prefetch_retries`.
- `prismaquant/streaming_model.py`: `schedule_prefetch` hands back a held read before the pressure floor and admission gates. It allocates nothing, and without that order a re-assert under sustained pressure was refused and counted as a memory skip for a read the context already owned.
- `prismaquant/aura_cost.py`: `compute_aura_cost_streamed` logs the retried layers after a layer-major capture. Each retry is one serialized source read in the critical path, so it should be visible.
- `docs/ARCHITECTURE.md`: the contract in its own stamp, cross-referenced from the #394 layer-major paragraph.

## Contract change in the tests

[...]

PR #446, opened 2026-09-08
Give back the activation-scale environment execute borrows

### LINKED ISSUE #445: Full-suite flake: tessera_joint_aura.execute leaves PRISMAQUANT_PROD_ACT_SCALES=0 in the process
## What happens

The CPU full suite fails one test, always the same one:

```
FAILED tests/test_served_activation_contract.py::test_stock_nvfp4_cache_score_is_unchanged - assert False is True
1 failed, 7237 passed, 200 skipped, 3 xfailed
```

It has been read as random or as belonging to whichever PR happened to be running. It is neither. It is deterministic and order dependent.

## Cause

`prismaquant/tessera_joint_aura.py:672` writes the plan's activation-scale value straight into the process environment:

```python
os.environ["PRISMAQUANT_PROD_ACT_SCALES"] = config["execution"]["production_act_scales"]
```

`_load_plan` requires that value to be the string `"0"` for every admitted plan, so the write is always `"0"`. It is a bare assignment, not a `monkeypatch.setenv`, so nothing takes it back. Four tests call `bridge.execute` in-process with `require_cuda_hot_path` patched to a no-op, and each of them leaves `PRISMAQUANT_PROD_ACT_SCALES=0` behind for the rest of that worker.

The victim is `prismaquant/production_weight_cache.py`, in `_local_forward_render_score`:

```python
if (activation_max_abs is not None and float(activation_max_abs) > 0
        and _env_flag("PRISMAQUANT_PROD_ACT_SCALES", True)):
    x_quant_input = x.clamp(-float(activation_max_abs), float(activation_max_abs))
    clipped = True
```

With `0` present the clamp is skipped and `clipped` stays False, so the record reports `activation_clipped` False.

That accounts for the exact failure shape. The test's first assertion, `record["score"] == dynamic`, passes, because both calls take the same unclipped branch and agree. Only `assert record["activation_clipped"] is True` fails.

## Reproduced

Through PrismaBuild, on `e4ed1cd863` (the #431 merge), one interpreter, one box, one file of difference:

| receipt | host | rc | files | result |
|---|---|---|---|---|
| `b712a5674dbb` | dl380g10 | 1 | `test_tessera_joint_aura.py` then `test_served_activation_contract.py` | 1 failed, 58 passed |
| `75b09a0bc7b6` | dl380g10 | 0 | `test_served_activation_contract.py` alone | 22 passed |

Under xdist the pair only lands together on some workers, which is what made it look random.

## Fix

Scope the write to the call. `execute` is a process entry point, so setting the value for the render path it drives is right; leaving it behind is not, and nothing outside `execute` reads the key. Restoring on the way out keeps the campaign byte identical inside and leaves the process as it was found.

The regression test belongs at `execute`'s boundary, not at the render scorer, and asserts both halves: the plan's value is live during the call, and the prior value is back after it.

RobTand:
Closes #445.

## The defect

`tessera_joint_aura.execute` sets `PRISMAQUANT_PROD_ACT_SCALES` from the admitted plan so the render path it drives reads the campaign's value. As a process entry point that is right. The write was a bare assignment, so an in-process caller never got the variable back, and `_load_plan` requires every admitted plan to carry `"0"` — which is exactly the input that turns the render scorer's activation clip off:

```python
# production_weight_cache.py, _local_forward_render_score
if (activation_max_abs is not None and float(activation_max_abs) > 0
        and _env_flag("PRISMAQUANT_PROD_ACT_SCALES", True)):
    x_quant_input = x.clamp(-float(activation_max_abs), float(activation_max_abs))
    clipped = True
```

Four tests call `execute` in-process with `require_cuda_hot_path` patched out, so each disarmed the clip for everything that ran after it in the same xdist worker. The CPU full suite has been failing one test on that:

```
FAILED tests/test_served_activation_contract.py::test_stock_nvfp4_cache_score_is_unchanged - assert False is True
```

The failure shape is the confirmation. That test's first assertion, `record["score"] == dynamic`, still passes, because both calls take the same unclipped branch and agree. Only `activation_clipped` disagrees.

## The fix

Scope the write to the call. Inside `execute` nothing changes and the campaign is byte identical. Plenty of code outside `execute` reads the key -- `production_weight_cache` is exactly that code, which is why the leak bites -- but nothing needs this command's value to still be set after `execute` returns, so the process is left as it was found. A decorator rather than an inline `try/finally` because the alternative is re-indenting 180 lines for a two-line behaviour change.

The regression test sits at `execute`'s boundary, not at the render scorer, and asserts both halves — the plan's value is live during the call, read from inside a callback that runs after the write, and the prior value is back after the call raises. Restoring without setting would be just as wrong, so both are checked, with and without a pre-existing value.

## Receipts

All through PrismaBuild on `dl380g10`, one interpreter, `/home/rob/venvs/pq-cpu312/bin/python`, one shard, no xdist.

| receipt | tree | rc | files | result |
|---|---|---|---|---|
| `b712a5674dbb` | `e4ed1cd863` (main, before) | 1 | `test_tessera_joint_aura.py` then `test_served_activation_contract.py` | 1 failed, 58 passed |
| `75b09a0bc7b6` | `e4ed1cd863` (main, before) | 0 | `test_served_activation_contract.py` alone | 22 passed |
| `7374f991765e` | this branch | 0 | the same ordered pair that failed | 61 passed |

The first two are the cause: one interpreter, one box, one file of difference. The third is the same ordered pair going green here, with the two new cases included in the count.

## Scope

Not a defaults, stage-graph, format-menu, lane or ship-gate change, so `docs/ARCHITECTURE.md` is unaffected. `docs/design/runtime_flags.md` documents the flag's meaning, which is unchanged; only who gives it back is.

#419's CI failure was this, not its own changes. It should be refreshed on this base and re-run once this lands rather than declared broken.

### DIFF
--- a/prismaquant/tessera_joint_aura.py
+++ b/prismaquant/tessera_joint_aura.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import argparse
 from collections import defaultdict
+import functools
 from dataclasses import dataclass
 import hashlib
 import json
@@ -610,6 +611,41 @@ def _io_counters():
     return values
 
 
+ACTIVATION_SCALE_ENV = "PRISMAQUANT_PROD_ACT_SCALES"
+
+
+def _restores_activation_scale_env(function):
+    """Scope ``execute``'s activation-scale write to the call that makes it.
+
+    ``execute`` sets ``PRISMAQUANT_PROD_ACT_SCALES`` from the admitted plan so
+    the render path it drives reads the campaign's value.  As a process entry
+    point that is right; called in-process it leaves the value behind.  Every
+    admitted plan carries ``"0"`` (``_load_plan``), and that is the input
+    which turns the render scorer's activation clip OFF for everything that
+    runs afterwards (``production_weight_cache.py``, in
+    ``_local_forward_render_score``).  Nothing outside ``execute`` reads the
+    key, so restoring it on the way out leaves the campaign byte-identical
+    and leaves the process as it was found.
+    """
+    absent = object()
+
+    @functools.wraps(function)
+    def wrapper(*args, **kwargs):
+        import os
+
+        prior = os.environ.get(ACTIVATION_SCALE_ENV, absent)
+        try:
+            return function(*args, **kwargs)
+        finally:
+            if prior is absent:
+                os.environ.pop(ACTIVATION_SCALE_ENV, None)
+            else:
+                os.environ[ACTIVATION_SCALE_ENV] = prior
+
+    return wrapper
+
+
+@_restores_activation_scale_env
 def execute(command, config, *, plan_sha256, prepared=None, resume=False, source_transition=None):
     """Execute one admitted preparation or one dependent cost action."""
     if source_transition is not None:
@@ -636,7 +672,7 @@ def execute(command, config, *, plan_sha256, prepared=None, resume=False, source
     from .tessera_reader import load_declared_reader
 
     require_cuda_hot_path("tessera_joint_aura", "cuda")
-    os.environ["PRISMAQUANT_PROD_ACT_SCALES"] = config["execution"]["production_act_scales"]
+    os.environ[ACTIVATION_SCALE_ENV] = config["execution"]["production_act_scales"]
     torch.set_num_threads(1)
     torch.set_float32_matmul_precision("highest")
     torch.backends.cuda.matmul.allow_tf32 = False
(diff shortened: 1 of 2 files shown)


RobTand:
Bite check for the regression test, so it is not just a test that happens to pass.

Receipt \`02f135f1b018\`, host \`dl380g10\`, rc 1, on a tree carrying this PR's test **and** its helper, with only the one decorator line removed from \`execute\`. Both parameter cases fail, and on the right assertion:

```
assert during["value"] == "0", "the plan's value must be live inside execute"   # passes
assert os.environ.get(key) == prior, "and must not outlive the call"
E   AssertionError: and must not outlive the call
E   assert '0' == '1'
2 failed, 37 passed in 5.77s
```

The live-during-the-call half passes and the restored-after half is what fails, which is the defect and nothing else. An earlier attempt that removed the whole commit failed on \`AttributeError\` at the constant instead, which would have proved only that the test runs.

PR #453, opened 2026-09-09
Integrate selected-anchor pricing, verified capture reuse and bounded profiling

### LINKED ISSUE #390: autoscale.selected_anchor_resources uses chosen multipliers with no process baseline; native guard peak 2.61 GB against a 12.6 MB plan

### LINKED ISSUE #443: Selected-anchor full-call CUDA traces exceed qualification evidence limits

### LINKED ISSUE #447: Selected capture reuse cannot select the existing verified loader
Selected streaming anchor reuse cannot opt into the existing verified activation loader: the frozen campaign rejects `--capture-load-policy` unless producing a new bounded capture, and its selected-reuse prefetch omits the policy.

Evidence: [`tessera_campaign.py:3962`](https://github.com/RobTand/prismaquant/blob/8d8a293ffbcadf6a40ec91fde03d8f690eafd2e5/prismaquant/tessera_campaign.py#L3962) rejects the selected path; [line 4292](https://github.com/RobTand/prismaquant/blob/8d8a293ffbcadf6a40ec91fde03d8f690eafd2e5/prismaquant/tessera_campaign.py#L4292) does not forward a verified policy. The existing legacy prefetch hashes then reloads each selected payload. Actual expert row 0076 selected 49,528,032,480 bytes; its enclosing sampled prefetch interval read 99,056,222,208 physical bytes. Main-thread samples and both-host Netdata are retained in `experiments/measurements/selected-verified-capture-20260908/`.

Severity P2: missing opt-in path and attributable execution evidence; no wrong artifact or served-quality defect is established. I could fix this in context and did: a separate reader branch forwards the existing loader policy, accounts for its private buffer/source pages/decode/scratch in the existing resource plan, and binds a separate immutable execution receipt to the original canonical capture. Defaults and normal resume/source gates remain intact.

Regression-first CPU tests through PrismaBuild observed three expected failures before the fix. The implementation has 235 completed unique CPU passes and one existing CUDA-only skip, with an OOM and a broader resume-file timeout retained as failed/incomplete attempts. The review branch carries explicit source boundaries; these results name the original frozen tested source. Native after-change I/O, timing, energy and tensor-parity evidence is still pending the coordinator's GPU window. No speedup is claimed. The linked draft PR is the in-session disposition; native claims remain gated.

Draft implementation: https://github.com/RobTand/prismaquant/pull/448 (stacked on #442; bounded native follow-up depends on #444).

### LINKED ISSUE #449: Allow explicit Tessera pricing families by dense or routed structure
The GLM first-artifact campaign needs to price dense layers and routed experts with different explicit Tessera family lists while reusing one calibration capture. The existing campaign expands the same family set before per-unit structure can narrow it.

Add an opt-in, versioned family restriction to the existing campaign menu path. Bind the policy and authoritative unit structures into checkpoint/cost provenance; refuse incompatible seed rows before linking their wires and inconsistent fanout merges. Preserve existing behavior when unset and retain all serving and release gates.

Implementation and targeted CPU validation are ready on `feat/tessera-campaign-family-restriction`; this issue tracks that bounded delivery. Native fit, full-model quality and serving qualification remain separate measured gates.

### LINKED ISSUE #451: Decorator docstring overstates who reads PRISMAQUANT_PROD_ACT_SCALES

RobTand:
GLM selected-anchor pricing now supports an explicit family policy by unit structure, verified reuse of the complete calibration capture, measured process-baseline admission, and bounded CUDA profile windows for scalar and compatible batched anchors. Unset options preserve existing behavior; capture bytes, producer pin, serving gates and format defaults remain unchanged.

Integrates the reviewed heads of #426, #442, #444, #448, #450 and #452 while preserving their commits. The architecture records all contributing contract changes. Closes #390. Fixes #443. Fixes #447. Closes #449. Closes #451.

Validation through PrismaBuild: the combined run found one integration regression (215 passed, 5 skipped): the new verified-prefetch phase still referenced a variable replaced by the shared capture-entry calculation. Separate commit `067e4d8654` uses that shared calculation; the affected reader/admission contracts then passed (27 passed, one CUDA-only skip, action `15879e538ca9ccc27b9eff7bdf8306826a23f254f5ef1a7eae38db136279b954`). Actual terminal, CAS output and source bundle were checked. Touched modules compiled before the one-line correction, and the corrected module has a separate compile action `7d727ea7ad3b11599597c495dab68e563d5f437b3560273fedc80003384bbcc9`. Required full CI must pass at the final head before merge.

Individual native evidence and its source boundaries remain in the included reports. The frozen pricing campaign stays at its separately bound source identity. This PR does not claim a served quality comparison or full expert qualification.


RobTand:
Confirming the integration preserved my two delivery heads, since that is the property the combined branch is trading against per-PR CI for.

Ancestry holds. `42b87588f6` (#426, carrying #452's `773502b708`) and `fa45a1d30e` (#444) are all three ancestors of `debf3ca6b36f44886b8cae50562251ea5cd5e189`. Checked with `git merge-base --is-ancestor`, not read off the body.

Content holds too, which ancestry alone does not prove: a later merge in the chain could have reverted a hunk and left the commit reachable. For each file those heads touch, I compared the added lines of my change against the added lines at the combined head, both measured from the same merge base. Zero of my lines are missing at `debf3ca6`, across `prismaquant/autoscale.py`, `prismaquant/memory_management.py`, `prismaquant/tessera_campaign.py`, `prismaquant/tessera_joint_aura.py` and `tools/dispatch_tessera_campaign.py`. `prismaquant/tessera_joint_aura.py` at the combined head is byte-identical to the #452 branch, so the reviewed prose landed verbatim.

The differences that do exist are additive and belong to #448: `capture_load_policy` threading through `autoscale.selected_anchor_resources` and the new `capture_prefetch` phase beside the terms #426 derived. That is composition, not overwrite.

Evidence from the frozen heads, for the record, since those runs finished after the freeze and are not being re-run here:

- #444 at `fa45a1d30e`: `tests (py3.12, cpu)` SUCCESS, `import surface` SUCCESS. Its earlier red was only `test_stock_nvfp4_cache_score_is_unchanged` asserting `activation_clipped is True`, the process-wide `PRISMAQUANT_PROD_ACT_SCALES=0` that #446 closed. Nothing in that branch was implicated, and this is a third independent confirmation that #446 fixed the leak rather than moved it.
- #426 at `42b87588f6`: `import surface`, `linked issue` and `publish linked-issue policy` all SUCCESS; its suite was still running at freeze time and I did not restart it.

No action requested. Flagging only so the preservation claim in the body rests on a check someone other than the integrator ran.

RobTand:
Completing the evidence from my previous comment: #426 at `42b87588f6` finished after the freeze with `tests (py3.12, cpu)` SUCCESS, so all four of its checks are green. Both frozen delivery heads are now fully green on their own, and that head is the one carrying #452's reviewed prose.

That does not substitute for required CI at `debf3ca6b3` -- the combined source is what merges -- but it does mean neither contributed head arrives with an open question of its own.

PR #457, opened 2026-09-09
Add sealed GLM TR3 full-vocabulary teacher and native vLLM scorer

### LINKED ISSUE #456: Track experimental GLM TR3 full-panel teacher and native scoring instrument
The sealed TR3 final panel contains 25 windows of 2048 tokens, but its published dataset revision does not include the named teacher logits. The existing gold tools have a distinct 8×512 contract. The comparison needs an explicitly attributed BF16 teacher for the sealed panel and a native full-vocabulary candidate scorer that proves causal row alignment and runtime identity before scoring all 51,175 positions.

The implementation is already on `experiment/glm-tr3-full-vocab`; this issue tracks its required PR delivery. It is not a deferred defect ticket. Scope is the experimental instrument and its attributable evidence, without changing production defaults, fitting capture, allocation inputs or serving-lane gates.

Acceptance: authenticated source and deterministic panel/teacher binding; CUDA full-vocabulary FP64 KL; public vLLM hooks with native TP/causal-alignment qualification; preserved runtime/source receipts and fail-closed replay comparison; tests and reported native results. Any unsuccessful run stays explicitly failed evidence. A paired Tessera-versus-EXL3 superiority claim remains a separate measurement outcome.

RobTand:
Closes #456

Add the experimental GLM TR3 teacher and full-vocabulary scorer for the sealed 25×2048 final panel. The published panel lacks its named teacher arrays, so the builder creates an attributed BF16 reference from authenticated bytes through the existing resident streaming path. The scorer retains KL(reference||candidate) over all 154,880 vocabulary entries in FP64, all 51,175 causal position values, and document/domain/window summaries.

Native one-window qualification binds exact source, teacher, candidate, image, tokenizer and TP configuration. Public vLLM hooks verify full-vocabulary geometry, unique score ownership and causal alignment. Restart replay ignores only positive block-count dimension 0 for three identified native KV backends; all other fields remain exact. Review follow-up adds bounded differing JSON paths and explicit malformed-shape/envelope diagnostics. Changed source requires fresh qualification before reuse.

This is an experimental instrument and attributable evidence. The EXL3 baseline uses the pinned external recipe image with its own patches; it does not attest Tessera's unforked-vLLM shipping gate.

Validation:

- Teacher: PB `90a36b449ef47cf5fbeb967400a59a7468281d661b05ac5bc60f3dac77ce4d2e`, exit 0 and cleanup verified. All 25 arrays separately read and hash-verified against teacher manifest `1cc798a32a3457f996e859f778fe61fd987561b91490fe2953b698457ea747ae`.
- Native source `ddc9aac80cda44c519d947b8468a0577a3ec17cb`: qualification09 and full25 completed. Full-panel mean **0.02446462473923544**, 25 windows / 51,175 finite positions, both TP ranks observing 1+1024+1024 calls per window, maximum native prompt alignment error `5.9569720178842545e-6` below `1e-4`. Artifact SHA256 `aec1930207bd18e680727ab4fac09515f30462443946dc8a87b723405f8d6ee5`. Logs report head exit 0 and matching hashes; their forced-shutdown/IPC warnings remain recorded. Listed producer files match historical source. No fresh native qualification is claimed for the diagnostic edit or integration with current main.
- Targeted diagnostic repair: four new regressions fail before the change; **83 CPU tests pass without skips**, plus scorer compilation, under PB `cffb2862274320f5f085614e24514bb14e28206d8831b7e1bb3109821554757f`. CAS receipt, payload, terminal cleanup and exact tested bytes verified. Four CPUs, 6 GiB, native threads1, priority -10, existing x86 Python 3.12 environment. Generic pb-cpu initially failed collection because compressed_tensors was absent.
- Review covered teacher/panel authentication, array identity, resident streaming, manifest publication, source policy, and the unchanged fail-closed replay comparisons. After merging main, PB `fe39aec07597fa18e498cf0d6d73aecdaeacee2bb1fa8b24bb96453828fa7319` passed 87 tests across the scorer and capture-profile suites; one native CUDA profiler test skipped on CPU. Terminal cleanup, CAS receipt/payload and source snapshot verified against PR head `08e44b2a2f`; no source differences.

**The full-panel result is measured but not yet a stable quality comparator.** [Issue #468](https://github.com/RobTand/prismaquant/issues/468) remains open: repeated identical final-0000 requests in one engine produced means 0.0462136121, 0.0472286403, 0.0573906394 and 0.0569008298 despite passing prompt alignment. No cause, noise floor, paired Tessera win or production admission is established. Merging this instrument does not resolve that measurement issue.

Evidence: `experiments/measurements/glm-tr3-full-vocabulary-20260908/README.md`, `native-completion-review.json`, `native-runtime-diagnostics-cpu-audit.json`, and real replay fixtures under `tests/fixtures/glm_tr3_runtime`. The teacher evidence includes a bounded in-process trace and both-host telemetry; it establishes completion/provenance, not a GPU-bound performance result or speedup. Final-panel positions are correlated within four documents; no independent-token significance claim is made. Frozen measurements remain immutable.

Full CI exposed a Qwen compatibility failure after the floating dependency advanced to Transformers 5.17. It was reproduced in the shared production streaming helper and repaired separately by #478; that repair passed 7545 tests in [full CI](https://github.com/RobTand/prismaquant/actions/runs/34400269994), with 207 skips, 3 xfails and 192 passing subtests. The TR3 scorer targeted tests passed. Historical native qualification remains bound to its original source.


RobTand (commented):
`experiments/measure_glm_tr3_vllm.py:286-321` (`qualification_runtime_matches`),
`:387` (the refusal)

Scoped review. I operate this instrument and hold the primary artifacts for the
two runs it cites, so I checked those claims and the KV comparison rather than
all 8286 lines. Someone else should read the teacher builder and the panel
loader.

Both regression fixtures are the real observations, byte for byte, not
reconstructions:

    tests/fixtures/glm_tr3_runtime/qualified-runtime.json
      1bdb1b41125ce6ef373a4c87200ff6cbf2a440c1e5c5e8d04eac841228ece605
      = my preserved logs/qualification-07/hook-qualification.runtime-c0c923d0...json
    tests/fixtures/glm_tr3_runtime/restarted-runtime.json
      54f3c5a3e3b98d8af8e54a1fc7e2499ce8e5dd74beeeb9dc49f267b5f69e55e6
      = my preserved logs/failed-full25-08/full-vocabulary-kl.runtime-f092b2d4...json

The qualification numbers in the PR body match the published artifact exactly:
2047 positions, call shapes `[1,154880],[1024,154880],[1024,154880]` on both
ranks, alignment error `2.3839675122871995e-6`.

"Three identified native KV backends" is exactly right, and stronger than the
body claims. Walking both fixtures: 66 entries allocate a KV cache, 22 each
from `DeepseekV32IndexerBackend`, `FlashInferMLASparseSM120Backend` and
`KpoolTailBackend`; `GDNAttentionBackend` allocates none across its 68 modules.
All 66 differ, every one only in dimension 0, and every non-shape field and
every other dimension is identical. So the relaxation is total within dim 0 and
empty everywhere else, which is what a capacity axis looks like and what a
layout change would not.

The declared ranks match the fixtures: 3, 3 and 4 respectively. The guard is
the right shape too. An unlisted backend gets no exception, `cache is None`
skips, a wrong rank raises rather than normalising, `type(size) is not int`
rejects bool, and the `except` returns False, so every unknown path fails
closed.

One finding, and it is about the message rather than the logic.

`:387` refuses with `native qualification differs from this
candidate/runtime/teacher/topology` and no detail. In this case it was none of
those four. Finding that out took a hand diff of both observations, flattened
to 1591 leaf fields, and root did the same diff independently. The function has
already built both normalised structures one line earlier; reporting the first
few differing paths would turn two separate investigations into output the tool
produces. Suggested shape: collect paths where the normalised values differ and
name up to a handful, so a future refusal says
`worker_runtime[0].attention_runtime[3].allocated_kv_cache.shape[0]` instead of
naming four things that did not change.

Second, smaller. `capacity_axis_backends` is keyed by full dotted module path,
so a vLLM refactor that moves a backend module silently drops the exception and
restores exact comparison. That is fail-closed and correct, but the failure it
produces is precisely the one this fix exists to prevent, reported with the
same misattributing message. The detail above would contain it.

The pending-result line is accurate. Fresh qualification on `ddc9aac8` is
staged and preflight-green on my side, holding for the root GPU window; I will
send the result when it lands.

VERDICT: MERGE AFTER the refusal names the differing fields

PR #459, opened 2026-09-09
fix: preserve observer sampling and concurrent progress serialization

### LINKED ISSUE #458: Preserve later Netdata samples after an individual sampling failure
The selected expert pricing observer stopped its Netdata thread after one response omitted the required GPU power chart at 01:39:43 UTC on 2026-09-09. `CaptureObserver.monitor` caught the error outside its loop (`experiments/glm_full_capture_profile.py:54`), so later samples from both Sparks were lost while encoding continued. The final evidence gate correctly remained failed.

P2: lost measurement evidence. Fix in progress on `fix/glm-observer-sample-recovery`: record bounded per-host sampling failures, continue the other host and later samples, and preserve the existing incomplete-evidence failure. Writer failures still stop monitoring.

I can fix this and have done so in the same session. The real failure is retained under `first-proof-anchor-preparation-02/workspace/rows/row-0076/profile`; recovered server history is separate. PB regression `aa80a577b9d0` fails on the original behavior; PB `ac140d7e0fb0` passes 43 tests with 3 CUDA skips. This CPU regression does not measure GPU performance.

### LINKED ISSUE #460: Observer threads insert dict keys while the main thread runs json.dumps(indent=2), so a completed anchor can surface as an encoder failure
`experiments/glm_full_capture_profile.py:68-77`, `:88-89`, `:145`, `:315-321`.

`CaptureObserver` shares one `self.result` dict between the monitor threads and the main thread. The main thread serialises it with `json.dumps(self.result, indent=2)` at `:145` and `:319`. `json` uses the C encoder only when `indent is None`; with an indent it runs the Python encoder, which walks every dict as `.items()`. A key insertion from another thread during that walk raises `RuntimeError: dictionary changed size during iteration`.

The monitor threads do insert keys while running:

- `:88-89` inserts `bytes_written` and `samples` into `result[kind]` on each thread's first round.
- `:68` inserts `sample_failures` into `result['netdata']` on the first sampling failure of the run.
- `:71-73` inserts the host key, and `:75-77` then inserts `last_error`, because the record at `:71-73` is built without it.

Up to five insertions per run, clustered at the first Netdata failure per host.

Consequence. The write at `:319` sits in a `finally` and catches `OSError` only, so the `RuntimeError` propagates and replaces the anchor's outcome: a completed anchor surfaces as an encoder failure and is re-encoded on retry. `AnchorObserver`'s own docstring at `:176-180` names that as the outcome the class exists to prevent. `:145` has no guard at all.

Measurement. PB `84835f970abc296763942cf69a530eccc013ad8ab977d71adb70c1579aa28756`, dl380g10, `pq-cpu312` 3.12.11, same object and same mutator across both arms: 376 of 400 dumps raised at `indent=2`, 0 of 400 at `indent=None`. That establishes the mechanism and isolates it to the encoder choice. It is not a field rate. The dump is roughly a millisecond against an anchor spacing measured in tens of seconds on the first-proof row 0076, so the chance per run is small.

Fix. Create every key the monitor threads will touch in `__init__`, and build the failure record complete before the single assignment:

- `netdata=dict(..., bytes_written=0, samples=0, sample_failures={host: None for host in hosts})`, and the same two counters on `python_sampler`.
- `:71-73` includes `last_error` from the start, so `:75-77` only replaces values.

Replacing a value for an existing key never resizes a dict, so the encoder cannot see a size change. `result['errors'].append` stays as it is: `_iterencode_list` iterates a list, which does not refuse on growth.

Test. Drive `monitor` with a failing `sample_netdata` while a second thread dumps `observer.result` with `indent=2`, and assert no `RuntimeError`. Mutating the observer rather than the fixture is what proves the guard bites.

Found while reviewing #459, which introduces three of the five insertions and is otherwise correct. Not a blocker there: losing the monitor thread was worse.

RobTand:
A transient missing GPU-power chart stopped the selected pricing observer’s Netdata thread, losing all later samples from both hosts while encoding continued. Continue after individual sampling failures and retain bounded per-host failure details; the final incomplete-evidence gate still fails and writer errors still stop monitoring.

A separate fix initializes monitor-owned dictionary keys before the threads start, preventing progress JSON serialization from failing when the first monitor round publishes its counters or first host failure. Progress remains non-atomic; final evidence is produced after shutdown.

Closes #458. Closes #460.

Validation through PrismaBuild: the original sampling regression fails on the old behavior; the deterministic concurrent-serialization regression fails for both monitor kinds on the old dictionary initialization. Final targeted run `550b801d8faa` passes 45 tests with 3 CUDA-only skips across the capture, selected-anchor and Netdata modules (DL380 CPU, four workers, one native thread each). Root verified the actual CAS payload, tested source, exit 0 and cleanup. Commands and full action hashes are recorded in `docs/experiments/glm_observer_sample_recovery_20260909/README.md`.

### DIFF
--- a/docs/experiments/glm_observer_sample_recovery_20260909/README.md
+++ b/docs/experiments/glm_observer_sample_recovery_20260909/README.md
@@ -1,0 +1,21 @@
+# Continue observation after a failed Netdata sample
+
+The live selected expert action `73af05a17fa4` lost Netdata monitoring at
+01:39:43 UTC on September 9 after one response omitted its required power
+chart. The outer exception handler ended the thread, losing subsequent
+samples from both hosts while encoding continued. The original evidence is
+retained under the first-proof preparation row 0076; recovered server history
+is separately recorded and never substituted into that original stream.
+
+Catch individual sampling failures, retain bounded per-host failure counts and
+timestamps, and continue other hosts and later samples. Writer failures still
+end monitoring; any missing sample still fails the existing evidence gate.
+
+PB regression `aa80a577b9d0733dd3956481b3bbc482537adc896c037fba040f600cc708cea9`
+failed because only the first host was attempted. PB
+`ac140d7e0fb0e7a320b3fa52703884f3c970b59521adf68993343e7e50c0b273`
+passed the three observer/Netdata modules: 43 passed, 3 skipped, 7.35 seconds
+on DL380 CPU with four workers and native threads bounded to one. The skips
+are CUDA-only coverage. An earlier wrong-environment submission lacked
+compressed_tensors and did not collect; it is not a regression result.
+This changes observation recovery, not pricing, sampling values or admission.
--- a/docs/experiments/glm_observer_sample_recovery_20260909/README.md
+++ b/docs/experiments/glm_observer_sample_recovery_20260909/README.md
@@ -19,3 +19,23 @@ on DL380 CPU with four workers and native threads bounded to one. The skips
 are CUDA-only coverage. An earlier wrong-environment submission lacked
 compressed_tensors and did not collect; it is not a regression result.
 This changes observation recovery, not pricing, sampling values or admission.
+
+## Concurrent progress serialization (issue #460)
+
+A first monitor round added dictionary keys while the main thread could be
+serializing progress, raising `RuntimeError: dictionary changed size during
+iteration`. Initialize monitor counters and both host failure slots before
+starting threads; each failure record also has its complete key set before
+publication. Progress remains a non-atomic observation; the final snapshot
+follows thread shutdown and retains any errors.
+
+The deterministic regression pauses the actual JSON dictionary iterator before
+running the first monitor round, for both monitor kinds. PB
+`2093a627f4189ac9360beba0cef59b369783d8b9d3ea271ac13e99e478893b8f`
+failed both cases with the dictionary-size error. Final green PB
+`550b801d8faaea6d9895201fa0959007a9b3dc3bb3bda7f540a5019c6306f1cd`
+passed 45 tests with the same 3 CUDA-only skips in 7.40 seconds (DL380 CPU,
+four workers, one native thread each). The terminal record confirms exit 0
+and resource cleanup. The CAS result hash is
+`b966e20248cd4757ea42a8d4f8593f0e3944122d14993d0b1efa9d8b3feb5261`;
+the tested snapshot is `232e6c89ddef8faa177c604d60c19f4d01240458`.
--- a/experiments/glm_full_capture_profile.py
+++ b/experiments/glm_full_capture_profile.py
@@ -39,9 +39,11 @@ class CaptureObserver:
             cpu_affinity=sorted(os.sched_getaffinity(0)),
             kernel=os.uname().release,
             netdata=dict(hosts=['sparky', 'sparklina'], interval_seconds=5,
-                         byte_cap=3*1024**3),
+                         byte_cap=3*1024**3, bytes_written=0, samples=0,
+                         samples_scope='rounds attempted',
+                         sample_failures={'sparky': None, 'sparklina': None}),
             python_sampler=dict(scope='main_thread_only', interval_seconds=1,
-                                byte_cap=512*1024**2),
+                                byte_cap=512*1024**2, bytes_written=0, samples=0),
             profile_layers=sorted(self.profile_layers),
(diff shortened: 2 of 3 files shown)


RobTand (commented):
`experiments/glm_full_capture_profile.py:63-79`, `:89`. Read the fix against the row-0076 stream it comes from.

The change does what it says. Only `sample_netdata` is inside the try, so a writer failure still ends the thread; the `for host` loop continues on the failing host's `continue`, so the second host is still sampled in the same round; `except Exception` leaves `KeyboardInterrupt` and `SystemExit` to the outer `except BaseException`; and `time` is already imported at `:19`, so the recovery path cannot itself raise the `NameError` that would reproduce the bug. The failure record is bounded at one entry per host with a counter, and aliasing it into `result['errors']` is what keeps the shutdown gate firing. The test drives the real `monitor` loop with a failing first call and asserts both halves, the retained later samples and the refusal at `__exit__`, which is the right shape: recovery is not gap erasure.

One non-blocking note, and it is the field I was misled by an hour ago.

`:89` increments `samples` once per while-iteration, outside the per-host loop. Before this change that was safe, because a host failure ended the thread and `samples` times hosts equalled the row count for the whole life of the file. After it, a run that loses one host for twenty minutes keeps incrementing `samples` at full rate while writing half the rows.

That is a live reading hazard, not a theoretical one. On row-0076 I used exactly that arithmetic, `progress.json` reporting 277 samples against 555 rows in `netdata.jsonl` for two hosts, to conclude the series was complete over the span it covered. With this fix in, the same arithmetic silently stops holding, and the reader has to know to go find `sample_failures[host].failed_samples` and subtract. The information is recorded, so this is about which field a reader reaches for first.

Cheapest fix is a per-host written counter alongside `samples`, so coverage is readable without a subtraction. Naming `samples` as rounds attempted in the record would also do it. Either is a follow-up rather than a change to this PR, since the behaviour here is strictly better than losing the thread.

Receipts read as stated: regression `aa80a577b9d0` fails on the original behavior, `ac140d7e0fb0` passes 43 with 3 CUDA-only skips, and the earlier wrong-environment submission is correctly excluded rather than counted.

VERDICT: MERGE

RobTand (commented):
`experiments/glm_full_capture_profile.py:68-77`, `:318-321`, `:88-89`. Reviewed on top of the existing read, which I agree with: only `sample_netdata` sits inside the try, the writer path still ends the thread, and aliasing the record into `result['errors']` is what keeps `__exit__` refusing. The `samples`-as-rounds reading hazard is real and is the right follow-up.

One defect the existing read does not cover, in the same lines.

The recovery path inserts dict keys from the monitor thread: `sample_failures` into `result['netdata']` at `:68`, the host key at `:71`, and `last_error` into that host's record at `:75`, since `:71-73` builds it without that key. The main thread serialises the same object with `json.dumps(self.result, indent=2)`. `json` reaches the C encoder only when `indent is None`; with an indent it runs the Python encoder, which walks every dict as `.items()`. A key insertion during that walk raises `RuntimeError: dictionary changed size during iteration`.

PB `84835f970abc296763942cf69a530eccc013ad8ab977d71adb70c1579aa28756`, dl380g10, `pq-cpu312` 3.12.11: 376 of 400 dumps at `indent=2` raised, 0 of 400 at `indent=None`, same object and same mutator. The encoder choice is the whole difference.

Consequence at `:315-321`. That write sits in a `finally` and catches `OSError` only, so the `RuntimeError` replaces the anchor's outcome and a completed anchor surfaces as an encoder failure, then re-encodes on retry. `:176-180` names that as the thing this class must not do. `:145` has no guard at all.

Rate, stated plainly so this is not oversold. A run inserts at most five keys, all clustered at the first Netdata failure per host, and the dump is roughly a millisecond against an anchor spacing measured in tens of seconds on row 0076. The joint chance per run is small. The receipt shows the mechanism, not a field rate; I did not observe this on a live action.

Half of it predates the PR: `:88-89` inserts `bytes_written` and `samples` on each thread's first round. The PR turns a startup-only hazard into one that can fire whenever Netdata first misbehaves.

Cheapest fix keeps the shape. Create the keys in `__init__` (`bytes_written=0`, `samples=0`, `sample_failures={host: None for host in hosts}`) and build the failure record complete, `last_error` included, before the single assignment. Every later touch is then a value replacement, which never resizes a dict, and the encoder cannot see a size change. Follow-up, not a change to this PR: losing the thread was worse.

VERDICT: MERGE

I will file the issue and carry that follow-up unless you would rather hold it.

PR #461, opened 2026-09-09
Install the Tessera the pin names, on the fleet too

### LINKED ISSUE #455: PrismaBuild's PQ interpreter carries an answer-equivalent but byte-different Tessera, so only the byte-identity assertions cannot pass in the fleet
The PrismaBuild interpreter PQ tests run on, `/home/rob/venvs/pq-cpu312`, has a
Tessera installed that is not the pinned one. Every test deriving from the
pinned contract therefore cannot run in the fleet, and passes only in GitHub
CI, which `tests/test_ci_tessera_install.py` gates.

Measured, not inferred. `pbtest` on `main` at `80388287`, DL380, action key
`c1edac13109e`, priority -10, x86, 1 shard, 1 worker, 1 thread, no GPU:

    tests/test_tessera_serving_pin.py
    tests/test_tessera_production_admission.py
    5 failed, 14 passed, 1 skipped

`test_the_pin_binds_the_bytes_the_installed_runtime_actually_packages` names
the cause:

    installed: 719daa02da1564b56a141ca2702ae29d4fda553460978efbb6510ddcd1824927
    pinned:    a688f8de244f936ec3a63a782e20af7985733e7a6fb0b4b981b5fe4c44112212
    pinned commit 07ad344c3275bb2fa7ce2432f93d89945d66f4c2

`load_tessera_contract` returns `None` on an unpinned install rather than
reading whatever is there, which is correct and is why the failures are honest
rather than silent. But the consequence is that the fleet cannot verify this
part of the suite at all.

Why it matters beyond a red count. These are the tests that hold producer-side
claims about the serving runtime to the runtime's own table, which is
principle 14's whole mechanism. An environment where they cannot run is an
environment where contract drift would go unseen, and it is the environment
every agent's `pbtest` receipt comes from. A receipt reading "green" from a
fleet that skipped or could not run them is weaker evidence than it looks.

Found while adding one such test in #454, whose new case fails in the fleet for
this reason and this reason only; the other 76 in that file pass (action key
`e40e67d6d6ae`). That PR is not blocked on this, but it cannot carry a green
fleet receipt until this is fixed.

Suggested fix: provision `pq-cpu312` with the Tessera the pin names, from the
same resolver CI uses (`tools/resolve_tessera_dev_pin.py`), and re-provision it
whenever the pin is re-reviewed. A re-pin is already a reviewed change to the
pin JSON and the module constants together, so the fleet install belongs on
that same change rather than drifting behind it.

RobTand:
Closes #455.

## What was wrong

CI checks out `RobTand/tessera` at the commit `tools/resolve_tessera_dev_pin.py`
prints and runs `pip install --no-deps` on that directory. Nothing did the same
for the PrismaBuild interpreters. `/home/rob/venvs/pq-cpu312` carried Tessera
built from `cc354b1f17` ("lane-eligibility v9"), which is an **ancestor** of the
pin `07ad344c32`: answer-equivalent, byte-different. The venv was behind the
re-pin, not diverged from it.

Confirmed read-only before touching anything, action `fa0be662f11d` on dl380g10:

    installed contract  719daa02da15...
    pinned contract     a688f8de244f...
    pinned commit       07ad344c3275bb2fa7ce2432f93d89945d66f4c2

The stale install carried the pin's own static version `0.1.0`, so nothing that
compares versions could have caught it. The contract bytes are what tell the two
apart, so the bytes are what the tool checks.

## The fleet is repaired

Provisioning action `15fc2a91fea9`, dl380g10, before/after inventories, then
`pip install --no-deps --no-build-isolation --force-reinstall` of
`/mnt/shared/tessera-pins/07ad344c3275bb2fa7ce2432f93d89945d66f4c2`, an
immutable tree whose `runtime_contract.json` hashes to the reviewed
`a688f8de244f`. Only `tessera-quant` was reinstalled; `uv pip check` reports no
broken requirements. `pb-cpu` and the GB10 project environments are untouched.

The three modules the issue names, re-run on the same interpreter after the
install, x86, priority -10:

| file | before (issue) | after | action |
|---|---|---|---|
| `tests/test_tessera_contract_v4.py` | 1 failed, 8 passed | 9 passed | `a5ccaba0b1d1` |
| `tests/test_tessera_serving_pin.py` | 2 failed, 12 passed, 1 skipped | 15 passed | `ac77036f782b` |
| `tests/test_tessera_production_admission.py` | 3 failed, 2 passed | 5 passed | `ea5967a72a5f` |

29 passed, no failures, no skips.

## What this adds

`tools/provision_tessera_pin.py` is that missing step, written so a re-pin can
carry it. It reads the pin the way the resolver does, out of the module's own
literal, and imports neither `prismaquant` nor `tessera`: the first pulls
`compressed_tensors` and the second is the thing being installed, and the
interpreter this exists to repair may have neither. `--check-only` reports and
exits 1 when an interpreter is off the pin, so a fleet audit can read the status
rather than a log line.

`tests/test_tessera_pin_provisioner.py`, 6 tests, action `f1021b93fb94`;
`tests/test_ci_tessera_install.py` still 6 passed, action `4edd9a3cd98b`. Both
x86 on the now-correct interpreter.

## What this does not establish

That every PrismaQuant test is CPU-portable, or that the other fleet
interpreters are on the pin. This provisions and verifies one interpreter and
gives the next re-pin a tool instead of a memory. GB10 project environments were
deliberately not touched in this change.

## Second round: root's four, and a fifth the fixtures were hiding

RED `9fe2f18f`, tests only, 6 failed / 12 passed (`3fd2bbb5d18d`). GREEN
`45a2ed6d`, 18 passed (`b7619de49cf6`). Both PB x86, `pq-cpu312`.

The fifth is the one that would have made the tool useless. Tessera's
`pyproject.toml` excludes `tessera._dev*` from what it installs, so the pin's
76 source files become 71 in a wheel. The digest compared them whole, so a
correctly provisioned interpreter reported drift, installed, re-probed and
exited 1, every run. Every synthetic test passed because the fixture copied the
whole source directory into site-packages. The fixture is fixed first, in the
RED commit, and the exclusion is now read from the pinned tree's own build
config and applied on both sides.

Root's four:

- `--check-only` is read-only. It reads a source that already verifies and
  reports `none, --check-only and no verified pinned source` otherwise, naming
  the command that would publish one.
- a tree that fails its manifest is renamed to
  `.quarantine-<commit>-<utc>-<random>` and kept. Something wrote to a shared
  pins root and those bytes are the only record of what.
- a tree with no manifest whose bytes match a fresh `git archive` is attested
  in place. The shared root holds one of those and rewriting it churns a
  directory other boxes read.
- staging is `mkdtemp` under the pins root, unique and owned, and publication
  takes an `flock` per commit. The PID-named scratch it replaces could collide
  between hosts and was removed by name.

And the build runs on a copy under `--staging-root`, with the tree re-checked
against its manifest afterwards, so root's "prove it leaves the frozen input
unchanged" is proved and not arranged.

## The real install, and what it found on the shared root

`tools/accept_tessera_pin_provisioner.py` makes a throwaway venv, provisions it
for real and reads the artifacts. Seeded from a copy of the live shared tree so
the repair path runs on the real bytes: six checks, all pass
(`df2e22218e8d`); unseeded `9fbb480fd14f`. Receipts and figures in
`docs/measurements/prismaquant-tessera-pin-provisioner-2026-09-09.md`.

It found the hazard already realised.
`/mnt/shared/tessera-pins/07ad344c3275bb2fa7ce2432f93d89945d66f4c2` digests to
`adb6511e` over 1256 files; the commit is `926f2f03` over 1179. The 77 extra
are 71 under `build/` and 6 under `src/tessera_quant.egg-info/`, written by an
earlier `pip install` that used the frozen tree as its build directory. (An
earlier revision of this body named a `_pb_native_moe_measure` directory among
them, read off a top-level listing rather than the diff; that one is in the
commit.) Nothing served from it is wrong, and **it has not been repaired
here** -- root has ruled it retained as bounded evidence.

`--clone` now defaults to `/mnt/shared/tessera-source.git`, a bare mirror, so
any fleet worker can archive the pin; no worker but sparky has a Tessera
worktree.

## Third round: root's two prose corrections, and the ordering behind one

Head `f0c8fd2d`.

1. The `materialise` docstring certified one live path as "predates the
   manifest and is correct". The acceptance run then found 77 extras in it. It
   now states the generic case: a tree written before this tool, or by a run
   that died after extracting and before attesting.
2. It also said a reader sees the old tree or the new one. Only a first
   publication is old-or-new; replacing a wrong tree is two renames with the
   path absent between them. The docstring says that, says the window is short
   and not zero, and says a repair wants a quiescent reader. The module
   overview no longer says an unmanifested tree is always re-materialised.
3. Root's optional improvement, taken: the staged tree is hashed and attested
   **before** the old name is vacated, so a failure there leaves the old tree
   in place. `test_a_failure_attesting_the_replacement_leaves_the_old_tree_in_place`
   makes `_write_manifest` raise and asserts it.
4. The pip spy in `test_the_install_does_not_run_inside_the_frozen_tree`
   returned `CompletedProcess(argv, 0)`, whose `stdout` is `None`.
   `capture_output=True` with `text=True` always yields strings, so the fake
   was answering in a shape the real call cannot, and the last commit's own
   logging raised on it. The previous head was never tested.

| what | action key |
|---|---|
| GREEN, `86e60a2f`: 19 passed | `2fc2f307ca5e` |
| the same tree with `_write_manifest` moved back after the quarantine rename: 1 failed, 18 passed | `b961932c705e` |

The mutation row is there because a test that passes on the code it is meant
to catch is not a test. The one failure is the ordering test.

### DIFF
--- a/docs/measurements/prismaquant-tessera-pin-provisioner-2026-09-09.md
+++ b/docs/measurements/prismaquant-tessera-pin-provisioner-2026-09-09.md
@@ -1,0 +1,87 @@
+# Provisioning the Tessera pin: what a real install proves that a mock cannot
+
+Issue [PrismaQuant #455](https://github.com/RobTand/prismaquant/issues/455).
+Branch `claude/pq455-provision-tessera-pin`.
+
+`tools/provision_tessera_pin.py` installs into a fleet interpreter the Tessera
+the pin names, and then checks that it took. Its tests place a package
+directory into site-packages and ask what the tool decides, which is the right
+shape for a decision and blind to two things only a build backend does: leave
+the pinned tree's bytes alone, and install a subset of them.
+
+Both turned out to matter, and one of them had already happened.
+
+## The digest could never have matched a real install
+
+Tessera's `pyproject.toml` carries
+`[tool.setuptools.packages.find] exclude = ["tessera._dev*"]`. That subpackage
+is the repository's own tooling, and a wheel does not ship it. The pin's tree
+holds 76 files under `src/tessera`; a correct install of that tree holds 71.
+
+The tool compared the two whole. On a correctly provisioned interpreter it
+would have found drift, installed, re-probed, found the same drift and exited
+1, on every run, forever. Every synthetic test passed because the fixture
+copied the whole source directory into site-packages, so the two sides were
+equal by construction.
+
+The exclusion is now read from the pinned tree's own build configuration
+rather than known in the tool, and applied on both sides. Tessera's
+`tools/check_wheel.py` proves the same exclusion on the built artifact.
+
+## The shared pins tree is not the commit, and pip is why
+
+`/mnt/shared/tessera-pins/07ad344c3275bb2fa7ce2432f93d89945d66f4c2` digests to
+`adb6511e...` over 1256 files. A `git archive` of that commit digests to
+`926f2f03...` over 1179. The 77 extra files are `build/lib/tessera/*` plus a
+`_pb_native_moe_measure` directory: output of an earlier `pip install` run
+with the frozen tree as its build directory.
+
+Nothing served from that tree is wrong -- `build/lib` is not a package under
+`src/` and never reaches an install -- but the directory is a working copy
+wearing a commit's name, which is the state a pin exists to exclude. The tool
+now builds from a copy under `--staging-root` and re-checks the tree against
+its manifest afterwards, so the property is proved rather than arranged.
+
+**The live directory has not been repaired.** Its replacement is a rename, and
+root's campaign may hold that path open; the timing is root's.
+
+## What the acceptance runs
+
+`tools/accept_tessera_pin_provisioner.py` makes a throwaway venv, provisions
+it for real, and reads the artifacts rather than the log. Seeded from a copy
+of the live shared tree, so the repair path runs on the real bytes:
+
+| check | result |
+|---|---|
+| the install did not write into the published tree | pass, `926f2f03...`/1179 before and after |
+| the tool says the tree is intact | pass |
+| the first run installed the pin | pass, `installed the reviewed source` |
+| the second run is a no-op | pass, `none, already the reviewed source`, drift `[]` |
+| `--check-only` reports clean | pass |
+| the tree it replaced was kept | pass, `.quarantine-07ad344c...-20260909T050854Z-j_ilktv2` |
+
+`unshipped_packages` reads `["tessera._dev*"]` off the pinned tree, and the
+wheel it built (`tessera_quant-0.1.0-py3-none-any.whl`, 725628 bytes) installs
+to a package the digest matches. That equality is the thing no placed fixture
+could show.
+
+## Receipts
+
+| what | action key | where |
+|---|---|---|
+| RED, `9fe2f18f`: 6 failed, 12 passed | `3fd2bbb5d18d` | dl380g10, `pq-cpu312`, `--tag x86` |
+| GREEN: 18 passed | `b7619de49cf6` | same |
+| acceptance, unseeded | `9fbb480fd14f` | same |
+| acceptance, seeded from the live tree | `df2e22218e8d` | same |
+
+The RED is a committed intermediate, not a scratch tree: `9fe2f18f` changes
+tests only and `45a2ed6d` is the fix.
+
+## What this does not establish
+
+One interpreter, one commit, one architecture. The GB10 venv is not
+provisioned by this and should not be until root rules on the PQ #455 test
+environment contract. The exclusion is read from setuptools' `packages.find`;
+a build backend that excluded files another way would need reading another
+way. And a lock serialises publication per commit within one filesystem's
+`flock`, which is what the shared NFS mount offers and not a distributed lease.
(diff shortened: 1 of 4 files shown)


RobTand:
Pushed `e6c6b67ef`, found by using the tool rather than reading it.

Auditing the GB10 interpreter returned `installed_contract_sha256_before:
null` and nothing more. Null covers three situations that want three different
repairs: no such interpreter, no Tessera installed in it, or a Tessera
installed without its packaged contract. The probe distinguished them and
discarded the answer.

`installed_contract` now returns `(sha256, reason)` and the report carries
`installed_contract_absent_before` beside the null. Against
`/home/rob/venvs/pq-cu130/bin/python` on sparklina it reads
`ModuleNotFoundError: No module named 'tessera'`.

That is a second, separate observation: the GB10 venv has no Tessera package
at all, where `pq-cpu312` had a stale one. Nothing has noticed because
`tests/conftest.py` puts the checkout's `src` on the path, so Tessera tests
import from source there. It bites only if a module needing the packaged
contract is run on the sparklina tag. Not provisioned, since I do not know
whether that venv is meant to carry Tessera; flagging rather than changing it.

Receipts, PrismaBuild x86: `tests/test_tessera_pin_provisioner.py` 7 passed,
no skips, `12907e3dba89`. The new case builds a real `venv` that really lacks
Tessera instead of skipping on the running interpreter, which under
`pq-cpu312` would have skipped every time the suite ran. Check-only against
the GB10 venv is the pbrun above, exit 1 as designed.

PR #465, opened 2026-09-09
Let the next batch encode while the last one is written

### LINKED ISSUE #464: Tessera campaign stalls the next encode on the previous batch's publication, receipt read-back and journal write
The Tessera campaign publishes two files per unit on the thread that just
finished the encode: the rendered weight shard through
`ProductionWeightCache._store_rendered_weight_entry`, and the `.tessera` wire
blob beside it. It then reads the wire file back off disk to make the unit's
receipt (`_checkpoint_wire_record`, `tessera_campaign.py:1787`), and writes the
journal row. The next batch's encode does not start until all of that has
finished.

Measured on the R1088 endpoint (root, 2026-09-09): a sixteen-unit arm spends
about 1.13 to 1.16 s inside sixteen `torch.save` calls and about 0.40 s inside
sixteen `Path.write_bytes` calls, against roughly 54 us of fsync. Those are
serialization and copy durations, not isolated NFS latency. None of that work
needs the GPU, and all of it sits in front of the next encode.

Moving the file writes alone is not enough. The receipt read-back is a
read-after-write barrier on every quantum, and the per-batch
`flush_checkpoint` performs journal I/O on the same thread, so an asynchronous
write awaited at the same barrier changes nothing.

What is wanted:

- An explicit, opt-in, versioned publication path that reuses the existing PWC
  entry writer, the existing wire path and the existing checkpoint journal.
  No parallel cache.
- GPU encode and scoring stay on their owning thread. Staged host bytes have
  stable ownership and a charged, bounded budget that the resource planner
  accounts for.
- Physical bytes, then receipt, then journal row: a completed checkpoint or
  receipt may never precede the durable files it cites.
- Fail closed on a writer failure, drain at every consumer barrier and at
  termination, keep recovery deterministic, and refuse ambiguous partial
  state.
- The existing fsync and page-release contract and the source identity are
  honoured as they are; no guard is weakened and no receipt is relabelled.
- Default unchanged, pending a real measured A/B.

Ordering and bound tests establish ordering and bounds only. The
before-and-after whole-publication-cycle profile, with both-host power and
work per joule, is a separate measurement on the boxes.

RobTand:
Closes #464.

The campaign pauses after encoding each batch to publish files, read wire receipts and write its checkpoint. This adds a bounded FIFO writer so the next batch can encode while publication completes. A row becomes measured only after its receipt exists; checkpoint writes follow the receipts they cite. Reservations happen before staging, oversized jobs fail closed, and unwinding preserves completed rows and propagates writer failures. Existing cache and atomic file publication remain the implementation.

`--publication-overlap-bytes` is opt-in and defaults to zero. Its memory budget is included in campaign admission and forwarded by the dispatcher. The architecture document describes the contract.

Validation at head `83d265aa560f54e8430e7d8da4c0ea783fccb1b7`:

- Full CPU CI: 7,352 passed, 204 skipped, three xfailed and 192 subtests. Root checked the actual job log and head binding; this is not CUDA coverage.
- Focused PB checks cover ordering, budget, failure/unwind, checkpoint/resume and the actual CLI. Root inspected executed source bundles and failing regression mutations; details remain in the branch's earlier review record and root evidence.
- Native GB10 / torch 2.13+cu130 / immutable b1 producer: the fully resident 864-unit GLM row with 32 encoded units, R832/B8, completed two warm profiles plus an unprofiled ABBA comparison. All 192 wires and scores matched. Synchronous versus overlap: 237.0481 versus 231.4060 seconds total, **1.02438x throughput**. Work/J improved only 1.00520x, an inconclusive energy difference. Before/after CPU/CUDA traces and both-host Netdata accompany the measurement.
- The R1088 comparison completed four arms with exact wires/scores, then refused the next memory phase guard. Its single unprofiled pair showed 1.02469x throughput; it is partial evidence, not a completed ABBA comparison. No OOM; cleanup completed.

Native R832 PB action `55a9fbd847d34329793c1eb04e3b8a95b392a55ea298d28c1554d4d426b64056`, independent audit `3c2158f07380a41b59bf23136abd2c9ca129904aa970b048a828b9dc19c8fedd`. Root verified terminal exits, cleanup, actual outputs, receipt hashes and source snapshots. Full results and negative evidence are retained at [the experiment record](https://github.com/RobTand/prismaquant/blob/349481522a/docs/experiments/glm_publication_cycle_20260909/README.md), with shared artifact paths and complete keys.

These bounded measurements establish neither full-model throughput nor served quality. The observed gain is modest, so the synchronous default remains. The measurement harness is retained separately from this production change.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -5,28 +5,41 @@ follow, newest first, each recording its own branch and date.
 
 Re-stamped (2026-09-09, `claude/bounded-publication-overlap`) for the campaign's
 opt-in **bounded publication overlap** (§4.10). `--publication-overlap-bytes N`
-(default `0`, which is the historical synchronous path byte for byte) stages up
-to N bytes of already-encoded artifacts and writes them on one bounded writer
-thread, so the next anchor batch encodes while the previous batch's render and
-wire files are written. The device-to-host copy of each render stays on the
-thread that owns the device work; the writer then calls the same
+(default `0`, which is the historical synchronous path byte for byte) runs one
+bounded writer thread so the next anchor batch encodes while the previous
+batch's artifacts are persisted. Three links move onto that thread, in queue
+order: the render and wire files, the wire receipt that is read back off them,
+and the `write_unit` journal write that cites the receipt. The
+device-to-host copy of each render stays on the thread that owns the device
+work, and the writer then calls the same
 `production_weight_cache._store_rendered_weight_entry` and the same
 tmp-plus-`os.replace` wire write the synchronous path calls, so there is no
 second cache and no change to what a published file is. An anchor is added to
-`measured` and marked for the journal only after its own files exist, so the
-checkpoint invariant that every journalled anchor row carries a wire receipt
-read back off a published file is unchanged; a round, a deadline stop and the
-end of the loop are drain barriers. Each job is charged
-`tensor.nbytes + len(blob)` and a submit blocks over budget, so staging is
-bounded; the memory guard's `after_selected_anchor_batch` bracket therefore
-includes staged bytes, and the publisher's budget, peak charge and blocked
-seconds are stamped on cost provenance under `publication_overlap` so a reader
-can attribute them. A writer failure is closed: everything queued behind it is
-dropped unwritten and the campaign's batch handler re-raises rather than
-continuing. The option is excluded from checkpoint input identity, because it
-chooses which thread performs two writes whose arguments it does not touch.
-Gate: `tests/test_tessera_publication.py`. Its barrier tests establish ordering
-only; no speed, work-per-joule or I/O claim is made here.
+`measured` and marked for the journal only after its own receipt exists, so
+the checkpoint invariant that every journalled anchor row carries a wire
+receipt read back off a published file is unchanged. A round, a deadline stop
+and the end of the loop are drain barriers; the publisher's lifetime is a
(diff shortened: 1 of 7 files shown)


RobTand:
Extra coverage worth naming, because it was missing from the receipt table
above and it is the one path that could have broken quietly.

`_finish_anchor`'s completed-anchor page release now lives inside the
`_publish` closure. On the default path that closure is called inline on the
encode thread, so the release still runs exactly where it ran, but nothing in
the 12-file regression drives it: the branch is guarded by the production
cache's `release_completed_anchor_file_pages`, which only a selected-source
row sets.

`tests/test_tessera_stack_group_cli.py` does drive it, end to end through
`campaign.main` on a sampled `s:` stack group, and it wraps
`release_activation_cache_file_pages` to check each call's `expected_stat`
against the file. Green at head, and it **ran** rather than skipping: the
pinned producer is present on the x86 worker.

| file | action key | rc | result |
|---|---|---|---|
| `test_tessera_stack_group_cli.py` | `c10f1c446b336fcb690f5de195412e61f4bdb22ce707d58087cc9477493da02b` | 0 | 1 passed |
| `test_glm_campaign_streaming.py` | `5a373004bae9b998849aeb82f118958c5a897b7eeec94804561a48ca1f38b287` | 0 | 12 passed |
| `test_tessera_stack_driver_integration.py` | `faa04f4f92a25f35b6055496412ad9f4a53107840289cda7cde89c27013b1cdb` | 0 | 5 passed, 1 skipped |

PR #473, opened 2026-09-09
Load authenticated selected source weights and preserve compatible anchor work

### LINKED ISSUE #474: Avoid unrelated source materialization in selected capture-reuse pricing
Selected anchor rows reusing a complete calibration capture still build unrelated resident source weights. In the pre-change path, `_build_streaming_context` materializes nonbody tensors and `snapshot_selected_weights` installs the full source layer before copying selected Linears. For the GLM layer-4 shared gate/up group this reserves 57 GiB and incurs roughly 47 GB of sampled process reads.

Fix in progress in #473: a shared metadata dependency closure for resource planning and the existing loader/cache, with source-only forward guards and full source-file authentication retained. Native A/B/B/A qualification measured 265.02→236.40 s, 6757.74→6386.74 device joules, and exact parity of all 14 anchors, quality/bpp and wire files. This does not establish routed-expert speedup or saturated encoding.

The same PR fixes a regression discovered while preserving previous work: seed adoption could inherit measured error after scoring activations changed. The production-path red test reproduced it; the fix binds the source manifest and actual scoring-row identity before accepting a unit. `--seed-workspace` then allows deterministic reuse of completed/partial row journals through those gates.

Severity P2 for source over-materialization/performance, P1 for the reproduced inherited-score defect. Both are implemented in separate commits on the active branch. This issue supplies the repository's required closing reference for the concrete PR; it is not deferred work. Evidence and exact PB receipts: `docs/experiments/glm_selected_source_snapshot_20260909.md` in #473.

RobTand:
Selected capture-reuse rows materialized unrelated head, vision and full-layer weights before copying selected Linears. Add opt-in `--source-snapshot-policy selected-tensors-v1` using a shared dependency closure for admission and the existing streaming reader/cache. Full source-file authentication and complete expert/concat parents remain required. Source-only contexts require planned keys and refuse forwards; configuration commits atomically.

A complete GLM shared-expert gate/up group in interleaved A/B/B/A runs fell from 265.02 to 236.40 seconds (10.80% less time) and 6757.74 to 6386.74 GPU joules (5.49% less energy). All 14 anchors, quality/bpp and wire bytes matched the prior production group exactly. Its reservation falls from 57 to 42 GiB. This result covers source preparation for that group; encoding headroom remains and routed-expert speedup is unmeasured.

Separate commits fix a seed-reuse defect reproduced during migration: a stored score could survive changed scoring activations. Seed adoption now validates the manifest digest, unit envelope and actual scoring input identity before linking wires. `plan --seed-workspace` reuses matching completed/partial row journals through those gates, requiring the same census, capture, group membership and sampling. At-plan digests are provenance; execution validates the then-current compatible seed.

Validation through PrismaBuild: source suite 215 passed/1 CUDA-only skip; focused snapshot tests and red/green guard regressions; seed suite 113 passed/2 missing-tool skips, then both skipped projection tests passed after supplying the pinned producer source; unchanged-input seed adoption passed; workspace reuse 6 passed; integrated planner/docs checks 58 passed; final compile passed. Counts overlap. Native action `782b0a7b31cce8a79a4e9d38a70b6fa523b4231001c71f874c764091b0480615` exited zero with complete cleanup; CAS payloads and actual source were verified. Both-host Netdata, main-thread stacks, child I/O and CUDA traces support the measurement. Independent review findings are addressed.

Evidence, limits, negative oversized-trace attempt and receipt paths: [measurement report](docs/experiments/glm_selected_source_snapshot_20260909.md). Runtime defaults, format menu, calibration and producer recipe remain as documented; the selected-source mode is opt-in. The GLM pricing restart uses the qualified mode with the original capture and 42 existing seed journals.

Closes #474.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,18 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-09 · `perf/glm-selected-source-scope`. Stamps
+As of: 2026-09-09 · `perf/glm-selected-source-snapshot`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-09, `perf/glm-selected-source-snapshot`) for seed-score
+identity validation. `--seed-checkpoint` now requires a digest-bound source
+manifest and matching unit envelopes, then compares calibration, score currency,
+static-scale policy, actual per-unit scoring-row identity and input scale before
+linking that unit's wires or inheriting its measured error. Producer input and
+wire checks still apply. Menu and source-package changes can reuse compatible
+measurements; changed scoring activations cannot inherit old scores. Missing
+unit shards remain eligible for fresh pricing, preserving partial-checkpoint
+recovery. Gates: `tests/test_tessera_campaign_resume.py` and
+`tests/test_tessera_campaign_fanout.py`.
 
 Re-stamped (2026-09-09, `perf/glm-selected-source-scope`) for opt-in
 `--source-snapshot-policy selected-tensors-v1` in selected anchor rows that
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -8,8 +8,11 @@ Re-stamped (2026-09-09, `perf/glm-selected-source-snapshot`) for
 matching prior row's completed or partial checkpoint to the existing seed gates.
 The planner requires the same model, census, capture, group bundles, membership
 and sampling; it records the source plan digest and each available seed manifest
-identity. Unstarted rows price normally. Seed files retain their original owner;
-PB continues to own all admission, placement and retries. The runtime validates
+identity. These at-plan digests are provenance, not execution gates: the source
+may advance before adoption, and the runtime checks the then-current scoring
+inputs and wires. Group additions or changed bundles/sampling refuse the plan;
+a matching row without a checkpoint prices fresh. Seed files retain their original
+owner. PB continues to own all admission, placement and retries. The runtime validates
 the seed identity and wire bytes when adopting them. Gate:
 `tests/test_tessera_campaign_seed_workspace.py`.
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,6 +3,16 @@
 As of: 2026-09-09 · `perf/glm-selected-source-snapshot`. Stamps
 follow, newest first, each recording its own branch and date.
 
+Re-stamped (2026-09-09, `perf/glm-selected-source-snapshot`) for
+`dispatch_tessera_campaign.py plan --seed-workspace`. A new plan can offer each
+matching prior row's completed or partial checkpoint to the existing seed gates.
+The planner requires the same model, census, capture, group bundles, membership
+and sampling; it records the source plan digest and each available seed manifest
+identity. Unstarted rows price normally. Seed files retain their original owner;
+PB continues to own all admission, placement and retries. The runtime validates
+the seed identity and wire bytes when adopting them. Gate:
+`tests/test_tessera_campaign_seed_workspace.py`.
+
 Re-stamped (2026-09-09, `perf/glm-selected-source-snapshot`) for seed-score
 identity validation. `--seed-checkpoint` now requires a digest-bound source
 manifest and matching unit envelopes, then compares calibration, score currency,
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,7 +1,35 @@
 # PrismaQuant Architecture
 
-As of: 2026-09-09 · `claude/dispatch-process-baseline`. Stamps
+As of: 2026-09-09 · `perf/glm-selected-source-scope`. Stamps
 follow, newest first, each recording its own branch and date.
+
+Re-stamped (2026-09-09, `perf/glm-selected-source-scope`) for opt-in
+`--source-snapshot-policy selected-tensors-v1` in selected anchor rows that
+reuse a complete, SHA-bound capture. The default remains `whole-layer-v1`.
+The source-only mode builds the declared meta skeleton but materializes no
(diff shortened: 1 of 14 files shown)


RobTand (commented):
Read of the draft while GPU qualification runs, so a defect does not get
qualified. The closure and the admission arithmetic hold up. Two notes on
coupling, and two affirmations.

## The plan-to-execution identity is default-open

`cost_streaming.py`, `snapshot_selected_weights`:

```
snapshot_only = getattr(self.context, 'source_snapshot_only', False)
if snapshot_only:
    self.context.configure_selected_snapshot(names, self.profile)
    if (expected_source_keys is not None and ...):
        raise RuntimeError('snapshot source dependencies differ from the admitted plan')
```

Narrowing is gated on `snapshot_only`; the check that the narrowing equals what
admission reserved is gated on `expected_source_keys is not None`. The two are
independently settable, so a context built with `source_snapshot_only=True` and
no expected keys narrows unchecked. The production caller in
`tessera_campaign.py` always passes them under `selected-tensors-v1`, so this
is not live, but the in-tree test at `test_selected_snapshot_scope.py` already
exercises the unchecked shape. The body's claim is that one closure drives both
admission and loading; `if snapshot_only and expected_source_keys is None:
raise` would make the code say that too.

## The new `_inflight` read does not take the lock

`streaming_model.py:634`:

```
if getattr(self, '_snapshot_source_keys', None) is not None or self._inflight:
    raise RuntimeError('selected snapshot source is already configured or active')
```

Every other `_inflight` access in the file is under `_inflight_lock` except one
pre-existing read at `:793`. This one gates the `weight_shard` / `weight_ckpt`
rebuild two lines later, which prefetch threads read, so it is the load-bearing
half of "no prefetch before this transition."

Related ordering: `_snapshot_source_keys` is assigned last, after the two dicts
are already replaced. If `_estimate_layer_cache_bytes` raises in between, the
context is narrowed but not marked configured, and a second call re-derives the
closure over an already-narrowed `weight_ckpt` and fails with a message about a
missing dependency rather than about the partial state. Building into locals and
assigning together would close both.

## Two calls worth affirming

`_campaign_checkpoint_identity` pops `source_snapshot_policy` from the settings
hash. That is right: the policy must not move results, so two runs differing
only in it should share identity, which is exactly what a wire-byte parity
criterion needs.

`source_validation_bytes` keeps the full-layer raw-page allowance with a comment
saying narrowing tensor reads does not narrow whole-shard authentication. The
conservative direction, and explained rather than asserted.

No production speedup claim to check yet, and the A/B/B/A shape with separate
processes and caches plus both hosts' series is the right instrument.

VERDICT: MERGE AFTER GPU qualification and the two coupling notes

RobTand (commented):
Read `cca7e38f20` and `df829b1392`. One finding, one question, and one thing
worth naming that the commit messages undersell.

## The scoring fix does more than its message claims

`tessera_campaign.py`, `_adopt_seed_checkpoint`: before this commit the seed
manifest's `identity_sha256` was read and used, unverified, as the identity that
`_load_unit` binds every shard to. A manifest could therefore claim an identity
it did not have and get its shards accepted under it. Now
`canonical_json_sha256(seed_inputs) != seed_identity` refuses first. That is a
larger change than "bind adopted scores to actual calibration activations", and
it is the part I would keep in the ARCHITECTURE stamp.

Both comparison loops are fail-closed on absence:

```python
if (field not in seed_inputs or field not in expected_identity or
        seed_inputs[field] != expected_identity[field]):
```

Written as `seed.get(f) == current.get(f)` two absent fields would compare equal
and inherit a score. They do not. Same shape for `units.<name>.scoring_rows` and
`input_global_scale`. Easy to get wrong, got right.

Replacing the hand-rolled `pickle.load` plus digest check with `_load_unit` is
the right direction too: one loader rather than a second copy of its rules.

## Recorded and never read

`dispatch_tessera_campaign.py:592` stamps `manifest_sha256_at_plan` into the
plan row, and `:691` stamps the row's `identity_sha256` beside it. Neither is
read anywhere in the tree; `test_tessera_campaign_seed_workspace.py:51` asserts
only that `identity_sha256` is *recorded*. The argv carries a path, so between
plan and execution the seed manifest at that path can change and the row adopts
whatever is there rather than what was planned.

This is not a correctness hole. `_adopt_seed_checkpoint` refuses a manifest whose
currency, calibration, scale policy or per-unit `scoring_rows` differ from this
run, so a substituted manifest that passes is scoring-compatible by
construction. But a hash written and never compared is the shape this repo calls
a confession log rather than a gate.

If these fields are inputs to the revalidation step your message describes and
that step is still to come, that is a fine answer. Please say so in the docs
stamp, so the next reader does not take them for gates that already bite.

## One asymmetry worth a sentence

`_seed_for_selection` returns `None` for a missing `cost.anchors.json`, and the
row prices fresh. A missing *row*, or a changed selection, raises and fails the
whole plan. So adding one bundle to a spec turns a reusable seed workspace into
a hard stop. The message names the constraint clearly, so if same-shape is the
intent this is only worth a line in the stamp.

## On the A/B

Four arms at exact 14-anchor quality, bpp and wire parity, with 1.121x wall,
5.49% device energy and 1.058x anchors/J, both hosts' series and bounded CUDA
traces. Energy alongside wall is the reading that survives; that it is 5.49%
against 10.80% wall is itself the honest part, and saying the encoding phase is
essentially unchanged rather than folding it into the headline is the right call.

VERDICT: MERGE AFTER a docs line on the two recorded-but-unread fields

RobTand (commented):
`9753a5b7c5` closes the MERGE AFTER. Read the lines rather than taking the
summary:

> These at-plan digests are provenance, not execution gates: the source may
> advance before adoption, and the runtime checks the then-current scoring
> inputs and wires. Group additions or changed bundles/sampling refuse the plan;
> a matching row without a checkpoint prices fresh.

That is both points, and it names the asymmetry rather than leaving a reader to
find it. Nothing further from me.

VERDICT: MERGE

PR #475, opened 2026-09-09
Record bounded server-cache warm-up and GLM loading-gap evidence

### LINKED ISSUE #476: Record bounded server cache warming and GLM startup-gap measurements
The GLM pricing investigation needed a bounded way to warm an explicit source/capture file set on dl380g10 and record whether later reads actually cross NFS quickly. Recent low-power windows also needed separation into startup/loading and steady encoding.

Implemented in PR #475: a read-only ZFS cache-warming helper and an attributable results report. A 66.089 GiB warm-up completed; a subsequent 19.985 GiB NFS read took 5.801 seconds during live pricing. The report records cache-retention limits, profiles, actual NFS counters, and source/CAS verification without claiming an isolated speedup.

This tracks the completed implementation for the repository's linked-issue merge gate; it does not defer a fix. Validation and evidence are in PR #475.

RobTand:
Record the requested dl380g10 cache warm-up and distinguish loading pauses from steady expert encoding. A bounded helper reads an explicit existing file set through the server's ordinary ZFS cache; it does not change quantization inputs or add an application cache.

The warm-up read 66.089 GiB. A follow-up four-shard transfer delivered 19.985 GiB to Sparklina in 5.801 seconds during live pricing, with NFS counters confirming network reads. The report records the workload, concurrent load, profiles, CAS/source audits, cache-retention limits, and the failed batch-width measurement without presenting it as a speedup.

Validation: PB compile check; three completed PB read actions with real output and receipt/source verification. This is an explicit diagnostic utility and evidence report, not automatic production prewarming.

Fixes #476.

### DIFF
--- a/docs/experiments/glm_server_cache_and_idle_windows_20260909.md
+++ b/docs/experiments/glm_server_cache_and_idle_windows_20260909.md
@@ -1,0 +1,89 @@
+# GLM server cache and GPU gaps — 2026-09-09
+
+A bounded warm-up read 66.089 GiB of an existing queued group's source and
+calibration files through dl380g10's ZFS cache. A subsequent read of the same
+four source shards transferred 19.985 GiB to Sparklina in 5.801 s: **3.445 GiB/s**
+while normal pricing continued. Client NFS READ counters increased by
+21.470 GB, consistent with the 21.459 GB diagnostic plus concurrent traffic;
+this was not merely a client page-cache read. The four shards read locally on
+the server in 6.219 s. These are observed warm-data rates, not an isolated
+before/after speedup or a claim about full capture-prefetch time.
+
+The server uses a four-disk RAIDZ1 pool, an NVMe L2ARC and a ZFS ARC configured
+with a 240 GiB maximum. Before the warm-up, a 20.044 s observation found
+10.006 GB of L2ARC reads and approximately zero HDD reads. That observation
+identifies an NVMe-backed interval; the aggregate demand-cache hit percentage
+alone would hide data supplied by predictive prefetch. It does not establish
+that every slow read has the same cause.
+
+The warm-up selected the existing `row-0045` layer-10 expert group: 864 capture
+files (46.104 GiB) and four complete source shards (19.985 GiB). It read each
+file once on the storage server with four 8 MiB buffers. It created no new
+weight/cache format and changed no quantization input. It took 183.240 s;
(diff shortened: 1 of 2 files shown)


RobTand (commented):
`experiments/server_cache_prime.py:35` and `:41`

Additive only, two new files, following the existing `experiments/` + `docs/experiments/` convention. The helper is properly fail-closed: plan pinned by sha256, hostname asserted, per-file size and `mtime_ns` revalidated by `fstat` **inside** the open handle and again after the read, `open('x')` on the result. That closes the TOCTOU the checks exist for.

The report is the part worth saying out loud. It refuses the speedup framing its own numbers would support ("observed warm-data rates, not an isolated before/after"), records the batch-width arm that did not qualify rather than dropping it, and applies principle 15 correctly on both instruments, including that GB10 utilization is not a saturation measure. The 32.06% below 15 W figure is walked back to 0.084% once preparation is excluded from the window, which is the honest correction and not the flattering one.

Two findings.

**1. `:41` The output path is the only path not validated.** Every `row['local_path']` passes `is_relative_to('/storage_pool/shared')` and a `resolve()==path` symlink check at `:35-36`. `plan['output']` is taken verbatim and written twice, at `:44` (`.started.json`, non-exclusive, so it clobbers) and `:57` (`open('x')`). The plan hash makes this unexploitable in practice, so this is asymmetry rather than a hole: on a helper that runs as an admitted action on the storage server, the path it *writes* is less constrained than the paths it *reads*, and writing is the more consequential direction. The same two-line guard applied to `output` would remove the asymmetry.

**2. `:31` The 67 GiB cap has no stated derivation.** The warm-up read 66.089 GiB and the bound is 67, so the constant reads as "just above what we did". The report gives the fact a bound could be derived from, a 240 GiB ARC maximum, but the code does not connect them. Naming what 67 protects, or expressing it against the ARC size, would make it a bound rather than a remembered number.

Neither changes what this measures. Both are one-liners if you touch the file again.

One confirmation for the record: the report's note that the deployed PB process-I/O aggregate carries the exited-child omission is exactly right, and correctly scoped. #451 is closed on merged #453, but the fleet runs generation `97f1f37e38a0` (PR #445), four merges behind, so the defect is live in what this ran on. That gap is now **prismabuild#456** and needs your decision, a drain window or a publication that does not require an idle queue.

VERDICT: MERGE

PR #490, opened 2026-09-11
campaign: streaming residency knobs leave the checkpoint identity; reseal tool drops them

RobTand:
## What

- `prismaquant/tessera_campaign.py::_campaign_checkpoint_identity` pops `streaming_cache_slots`, `streaming_prefetch_workers` and `streaming_cache_headroom_gb` from the bound settings. They size the layer cache, its prefetch pool and the free-memory floor the row keeps (`cost_streaming.build_streamed_causal_lm`); no captured activation, Hessian or wire differs at any of their values. `streaming` and `streaming_capture_policy` stay bound (they choose what is captured).
- `tools/reseal_campaign_identity.py` learns the matching identity delta: a pins file may carry `drop_settings`; a pending row must bind every listed setting and a migrated row none (`classify_pins`); the values leave `identity.settings` and are recorded in the migration record (`dropped_settings`); `verify` refuses a dropped setting that is still bound; a settings-only migration with unchanged pins is allowed, identical pins without a drop list are still refused.

## Why

The live GLM census rows are priced at 104 GiB, of which 24 GiB is `declared_headroom_bytes` — the whole slack between the priced plan and the measured rows (sparky MemAvailable min 33.5 GB, lina 36.2 GB at width 16 → rows use ~83–85 GB). Width 32 (`--anchor-batch-size 32`, already identity-neutral) prices at 106 GiB under a 24 GiB floor and does not fit the gb10 reservation; a 12 GiB floor does. With the floor bound into the identity, lowering it needed a new identity for every completed row for a value that never touched a receipt. This PR makes the floor a resubmission flag, and gives the reseal migration (the #486 ruling: completed rows may be resealed under GPU proof) the ability to carry completed rows across the pin move this change itself causes.

## Tests

- `tests/test_tessera_campaign_fanout.py::test_streaming_residency_knobs_are_scheduling_not_identity` — the three knobs (and `anchor_batch_size`) change nothing; `streaming_capture_policy` and `streaming` still do.
- `tests/test_reseal_campaign_identity.py` — three new tests: dry-run lists `identity.settings.<knob>`, migrate removes the keys, records the values, reseals every shard, verifies and is idempotent; a row missing the knob or still binding it after the pin move is foreign / fails verify; a settings-only migration keeps the pins.
- PrismaBuild receipts (x86, `pq-cpu312`, priority −10): pending, will be appended below.

## Deployment

Not part of the frozen campaign source; the live rows keep running under d403cc5a31. The next reseal cycle (one migration) carries `drop_settings` for the three knobs together with the new `prismaquant_source_sha256`; the resubmission of the ready rows then declares `--streaming-cache-headroom-gb 12` with a width-32 arm measured first (memory peak, J/anchor, `exact_parity`).

https://claude.ai/code/session_01XsJsJGJtE4CYHE9QKyRyiz

### DIFF
--- a/prismaquant/tessera_campaign.py
+++ b/prismaquant/tessera_campaign.py
@@ -1610,6 +1610,13 @@ def _campaign_checkpoint_identity(*, weights, acts, hessians, menus, args,
     # metadata for retaining the same producer receipt. Neither changes the
     # receipt, so binding either would make scheduling wear an identity's clothes.
     #
+    # ``streaming_cache_slots``, ``streaming_prefetch_workers`` and
+    # ``streaming_cache_headroom_gb`` size the layer cache, its prefetch pool and
+    # the free-memory floor it keeps (`cost_streaming.build_streamed_causal_lm`); every
+    # captured activation, Hessian and wire is the same object at any of their
+    # values, so they are scheduling too.  ``streaming`` itself and
+    # ``streaming_capture_policy`` stay bound: they choose what is captured.
+    #
     # ``units``, ``calibration_census`` and ``census_out`` are locations too,
     # and each one's load-bearing content is already bound by value somewhere
     # in this identity: the selection by the ``units`` map below (which holds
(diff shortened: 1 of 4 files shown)


RobTand:
PrismaBuild receipts (x86 dl380g10, `/home/rob/venvs/pq-cpu312/bin/python`, priority −10, 2 shards × 4 workers, tree c93e3ecc01), submitted from dl380g10 because of PB #500:

- `?` on ?: 57 passed, 2 skipped, 56 warnings in 26.56s — test_reseal_campaign_identity.py, test_tessera_campaign_family_restriction.py, test_tessera_bound_identity.py
- `?` on ?: 101 passed, 5 skipped, 56 warnings in 155.84s (0:02:35) — test_tessera_campaign_fanout.py, test_selected_verified_capture.py, test_tessera_campaign.py

2/2 shards green; the 7 skips are the pre-existing CPU-only skips. Receipt: `dl380g10:/home/rob/tmp/glm-perf-w1e/pbtest-identity-knobs-x86.{json,log}` (copy under `/home/rob/tmp/glm-perf-20260910/w1e/` on sparky).

Issue #560, opened 2026-09-13
Three defects in the fixed-resource admission consumer: a discarded attestation, a dropped placement obligation, and an unattested v1 path


RobTand:
## Three defects in the fixed-resource admission consumer, found while reviewing D37

A read-only review of `origin/main` @ `98da5093` turned up three problems that are
independent of the D37 contract decision and can each be fixed on the PrismaQuant
side alone. Full note: `/home/rob/tmp/agents/goal2/design-note-fixed-resource-admission.md`.

### 1. A fixed-resource refusal discards a native-row attestation that passed

`admit_runtime_provenance` (`runtime_provenance.py:809-817`) calls `admit_native_rows`
and `admit_fixed_resources` inside one `try`, and the loader sets a single flag
(`measured_runtime_prices.py:466-469`). `admit_fixed_resources` cannot pass today
for any table, so the 49 rows `admit_native_rows` admitted for the first time ever
in PQ #557 are thrown away with it.

Split `producer_admitted` into `native_rows_admitted` and `fixed_resources_admitted`.
Two gates that answer different questions should not share one answer.

### 2. `recomputed_placement_obligation_bytes` is demanded and then dropped

The gate refuses a table whose placement obligation is not recomputable
(`runtime_provenance.py:696-701`), and then nothing consumes it. `RESOURCE_FIELDS`
(`measured_runtime_prices.py:31`) has no off-step slot and `fixed_device`
(`allocator.py:3901-3905`) sums only the four fixed terms.

The obligation is `max(scalar_budget_bytes, non_step_transient_peak_bytes)`
(`full_engine_resource_report.py:124`, `:926-938`). When the off-step peak is the
larger of the two, the allocator prunes against the smaller number — which is
exactly the failure `runtime_provenance.py:690-695` describes in prose, on unified
memory where it is an OOM and not a spill.

Add the off-step slot to `RESOURCE_FIELDS` and carry the obligation into
`fixed_device`, so the number the gate demands is the number the DP prunes on.

### 3. A v1-schema table reaches the DP with no admission at all

`parse_measured_runtime_table` keys v2 off the schema string (`measured_runtime_prices.py:371-381`).
A v1 table carries no `runtime_provenance`, so neither gate is called
(`:466-469`), and `build_runtime_resources` raises only when
`runtime_provenance is not None and not producer_admitted` (`:476`).

So the path the whole admission apparatus exists to close is open by
construction: emit the same measured table under the v1 schema and it reaches the
[...]

Issue #559, opened 2026-09-13
Price the fp4 route class: the prefill curve measured only fp8 and bf16


RobTand:
## The gap

PQ #557 (merged `98da5093`) priced 49 native operator cells for Qwen3-0.6B
layer 0 and concluded that prefill at that layer is a **route-class step, not a
rate axis**. That conclusion was measured over exactly **two** route classes:

| family | rows | route symbol |
|---|---|---|
| `TESSERA_E4M3_K1_R*` (16 rates) | 42 | `torch._scaled_mm`, fp8 x fp8 |
| `TESSERA_BF16_K1_R1792` | 7 | `torch.mm` |

The third route class the pinned Tessera contract publishes on sm_121 —
**`TESSERA_NVFP4`, fp4 x fp4 through the same `torch._scaled_mm`** — was never
priced. It is absent for a scoping reason, not a capability reason: the table
was built from the formats seven *already-served artifacts* assign, and no
served artifact assigned an fp4 format.

That leaves the fast end of the prefill axis untested. On Blackwell an fp4 x fp4
`_scaled_mm` is expected to beat fp8 x fp8, so the missing arm is the one most
likely to move prefill, and the published finding "no prefill ordering resolves"
is a statement about **rate within one route class**, not about the format axis.

## What is actually reachable today

`TESSERA_E2M1_K2_R896` is servable and attested right now. Verified in code:

* `src/tessera/serving/runtime_contract.json` (`contract_version` 24) carries a
  `TESSERA_E2M1_K2` row, grid `E2M1x2`, reader range `[896, 896]`, attested at
  896, and `platforms.sm_121.executes["TESSERA_E2M1_K2"] = "e2m1_group16_ue4m3_static"`.
* `serving/scheme.py:230-240` maps that family to route `TESSERA_NVFP4` with
  `gemm_symbol = "torch._scaled_mm"`; `serving/nvfp4_route.py:236` executes it
  and `:242` stamps the symbol — the exact analogue of the `expected_route.symbol`
  the E4M3 cells recorded.
* The static input scale the NVFP4 bench requires is plumbed end to end:
  `bench_native_operator.py:406-407` demands a finite positive
  `input_global_scale`, `pq_frontier_native_cells.py:298` supplies it, and
  `tessera_formats.py:1520` sets `measured_as_served=True` on Tessera W4A4 rungs.
* The PQ harness hardcodes no format list: `pq_frontier_native_cells.py:385`
[...]

PR #561, opened 2026-09-13
Two gates two flags, an emitter that reports both, and a binding that names its route class (#565)

### LINKED ISSUE #565: Two admission gates, the obligation the gate demands, an emitter that reports both, and a route field that names its class
## What this asks for

`#560` lists three defects in the fixed-resource admission consumer. Two of them
are fixed on `claude/560-fixed-resource-admission-consumer`; the third (a
v1-schema table reaching the DP with no admission at all) is **not**, because
deciding whether that is intended back-compatibility or a hole needs evidence
the branch could not establish. So #560 stays open and this issue scopes exactly
what the branch delivers.

### 1. Two gates, two flags (#560 defect 1)

`admit_native_rows` and `admit_fixed_resources` attest different objects and
shared one `producer_admitted` flag inside one `try`. `admit_fixed_resources`
cannot pass for any v2 table today (D37), so the 49 rows `admit_native_rows`
admitted for the first time in PQ #557 were thrown away with it. Split the flag;
raise on the native gate, return the fixed-resource refusal, and spend that
refusal at the consumer that reads `fixed_resources`.

### 2. Prune on the placement obligation (#560 defect 2)

The gate refuses a table whose
`max(scalar_budget_bytes, non_step_transient_peak_bytes)` is not recomputable
and then nothing consumes the number, so the DP's device filter compares the
budget against the per-step composition alone. When the off-step peak is the
larger, that prunes against the smaller number. Carry the obligation into the
solver's placement bytes. Absent means *not priced*, never zero.

### 3. The emitter must not certify an admission it does not have

Consequence of (1), and a regression the hosted CI caught on #561:
`admit_runtime_provenance` no longer raises on the fixed-resource refusal, it
**returns** it, and `native_receipt_table.main` never reads the returned
`fixed_resources_refusal`. The CLI exits 0 and prints
`{"status": "admitted", "refusal": null}` for a table whose fixed resources are
still refused. Surface both flags and the refusal text, and decide deliberately
what the exit code means now that two gates can disagree.

### 4. `binding.operator_route` collapses two route classes

`native_receipt_table.py:179` and `native_moe_panel.py:546` set a row's
`binding.operator_route` to the **GEMM symbol alone**. `TESSERA_FP8` and
`TESSERA_NVFP4` both execute `torch._scaled_mm`, on differently packed operands
under different activation contracts, so fp8 and fp4 rows are indistinguishable
in the one field a downstream consumer compares. Found while pricing the fp4
route class (PQ #559 / #563). Make the field identify the route, not the symbol,
and state what the change of meaning invalidates.

## Done when

Each fix has its own commit behind a regression demonstrated first,
`docs/ARCHITECTURE.md` is updated in the same commit as any contract or gate
change with its provenance block re-stamped, and hosted CI is green on the
pushed head. #560 stays open for its defect 3.

https://claude.ai/code/session_01GmYxSxhmzbBcBoFTjbLi1G

RobTand:
Closes #565.
Part of #560.

Four fixes, each its own commit behind a regression demonstrated first. #560
stays open for its defect 3 (the unattested v1 path), which this branch does
not touch; #565 scopes exactly what it delivers.

## Commit 1 — `435e038e` split `producer_admitted` (#560 defect 1)

`admit_native_rows` and `admit_fixed_resources` attest different objects and
now carry different answers.

- `admit_runtime_provenance` raises only on the relation and native-row gates,
  and **returns** the fixed-resource refusal text.
- The loader sets `native_rows_admitted`, `fixed_resources_admitted` and
  `fixed_resources_refusal` in place of one `producer_admitted`.
- `build_runtime_resources` refuses on **`native_rows_admitted`** — it reads
  `row.resources` for priced candidates and never `table.fixed_resources`.
- The fixed charge gets its own consumer-side gate,
  `measured_runtime_prices.admitted_fixed_resources`, which carries the refusal
  text; `allocator.py` reads `fixed_resources` through it at all three sites.

No table becomes admissible that was not: the same refusal fires, at the
consumer it is about, and a v2 table still cannot price a device budget. What
changes is that the 49-row native attestation from PQ #557 survives it.

## Commit 2 — `e68a610d` prune on the placement obligation (#560 defect 2)

The gate refuses a table whose
`max(scalar_budget_bytes, non_step_transient_peak_bytes)` is not recomputable
and nothing consumed the number, so the DP's device filter compared the budget
against the per-step composition alone. When the off-step peak is the larger,
that prunes against the smaller number.

- `RuntimeResources` gains an optional `non_step_transient_peak_bytes`. It is
  one whole-engine obligation, so a per-unit row that declares one is refused,
  exactly as `kv_bytes` already is.
- The gate checks the declared value against the recomputed one.
- `allocator_solver._placement_bytes` prunes and reports
  `max(step_total, off_step)`.

**Absent means "not priced", never zero.** The field is optional on the wire,
every table emitted before it existed re-emits byte-identically, and the
arithmetic is unchanged wherever the off-step peak is absent or the per-step
composition already covers it.

## Commit 3 — `44f9f824` the emitter reports both gates (#565 item 3)

Hosted CI was red on `e68a610d` and the failure was a real regression in this
PR. Commit 1 stopped `admit_runtime_provenance` raising on the fixed-resource
refusal — it returns the text now — and `native_receipt_table.main` read only
the exception. So the emitter printed `{"status": "admitted", "refusal": null}`
and exited 0 for a table whose fixed charge is refused: it certified an
admission it did not have.

`native_receipt_table.admission_report` reads both flags and the refusal. The
report's `admission` block carries `native_rows`, `fixed_resources` and a
`status`, and the top-level `refusal` is `null` only when `status` is
`admitted`. **What the exit code means**, now that two gates can disagree:

| exit | what the table is |
|---|---|
| `0` | both gates admitted. Unreachable while D37 stands. |
| `3` | the rows are admitted and priced; the fixed charge is refused, and `admission.refusal` is the gate's own reason. |
| `2` | the loader refused the table outright: no row is priced, and the fixed-resource gate was never reached. |

Nonzero in both refusing cases, because an emitter must never certify an
admission it does not have. Distinguishable, because a caller holding 49 priced
rows should not read that the same way it reads "this table prices nothing" —
re-collapsing the two answers at the CLI would undo the split one layer up.
The module docstring, which still said `build_runtime_resources` refuses a v2
table the loader did not admit, is corrected in the same commit.

## Commit 4 — `cbd5d31d` the binding names its route class (#565 item 4)

`binding.operator_route` carried `declared_route["symbol"]` alone.
`TESSERA_FP8` and `TESSERA_NVFP4` both execute `torch._scaled_mm`, on
differently packed operands under different activation contracts, so fp8 and
fp4 rows were indistinguishable in the one field a downstream consumer
compares. Found while pricing the fp4 route class (#559 / #563). Both panel
kinds had it — `native_moe_panel` froze the same projection into the panel's
own `runtime_binding`.

`native_operator_panel.operator_route_identity` is the single spelling of that
field: the declared route itself, canonically ordered by the same
`json.dumps(..., sort_keys=True, separators=(",", ":"))` that `identity_sha256`
hashes. Four callers: the dense emitter, the dense expectation in
`admit_native_rows`, the frozen MoE binding, and
`experiments/pq_runtime_relation.py`.

**What it invalidates.** The field's meaning changed and nothing accepts the
old spelling. A v2 table emitted before this commit carries a bare symbol, its
freshly derived expectation is now the route identity, and `admit_native_rows`
refuses it at `native row operator binding`. That is one file today — the
49-row Qwen3-0.6B table D37 cites — and it is re-emitted from its own receipts
rather than migrated. No measured number moves; the table's digest does. D37
now records this, and stamps its 49-row admission as measured on the old
spelling. Accepting both spellings was rejected: it re-admits the ambiguity the
change exists to end.

## Merge with main

`origin/main` merged twice: `8f3ea8d4` (#562, D38) and `a73ab9ce` (#563, D39).
The only conflict was the §12 debt register, where #562 added D38 in the same
table this branch amends D37 in; both rows are kept, sorted, and **nobody is
renumbered**. This branch adds no new debt row of its own.

**D39 is edited, deliberately, and not renumbered.** #563's row said the route
field collapse was "separately observed and not fixed"; commit 4 fixes it, so
leaving that sentence standing would leave a false claim in the register. The
row keeps its number, its author's framing and its scope; what changed is the
observation sentence (now "fixed since, in #565") and its two
`native_receipt_table.py` line citations, which commit 4 moved.

## Receipts

PrismaBuild, `pq-cpu312`, `--tag dl380g10 --priority -10`.

**Regressions, each red before its fix:**

- `test_the_emitter_never_reports_an_admission_it_does_not_have` —
  `assert 'admitted' != 'admitted'` (3 failed / 13 passed, `redA2.json`).
- `test_two_route_classes_that_share_one_gemm_symbol_are_two_bindings` —
  `assert 'torch._scaled_mm' != 'torch._scaled_mm'` (2 failed / 109 passed,
  `redB.json`), demonstrated with the field still defined as the symbol alone.

**Green at head `a73ab9ce`:** 2 shards, 532 passed, 0 failed, 0 skipped, over
14 files — `test_native_receipt_table`, `test_native_operator_panel`,
`test_native_moe_panel`, `test_runtime_provenance`,
`test_runtime_admission_split`, `test_runtime_fixed_resource_admission`,
`test_full_engine_resource_report`, `test_container_runtime_identity`,
`test_measured_runtime_prices`, `test_allocator_measured_runtime_cli`,
`test_allocator_runtime_frontier`, `test_runtime_candidate_preservation`,
`test_architecture_doc`, `test_docs_staleness`.
JSON: `/home/rob/tmp/agents/goal2/w561fix/greenC.json`.
(`greenB.json`, 388 passed over 12 files, is the same set at `cbd5d31d`,
before the #563 merge.)

**Correction — the venv pin did not hide the commit-3 regression.** That
diagnosis was recorded on this PR and is wrong, and commit 3's message carries
it. `pq-cpu312` reproduces the failure exactly: `redA2.json` above is that
venv, red on the lie, before the fix. Whatever let the regression through a
green PB run, it was not a contract disagreement about this test. The general
point survives — `pq-cpu312` is pinned to an older Tessera contract and its
failure set is not hosted CI's — but it does not explain this failure, and I
have not established what does. The remaining candidate is test **selection**
(a targeted PB run certifies only the files it is given, and this one names
files explicitly); I did not find the `e68a610d` PB record to confirm it, so
that is a hypothesis, not a receipt.

The acceptance bar for this PR is the hosted `tests (py3.12, cpu)` conclusion
on the pushed head, which runs the whole suite and needs no selection
judgement.

## Not done

- #560 defect 3 (a v1-schema table reaching the DP unattested): unchanged. The
  issue asks for intent to be established before closing it, and the history
  does not establish it. `measured_runtime_prices.v1` predates the admission
  apparatus (`10a469a3`); v2 was added on top in `607ef367`, whose message
  records no decision about the older schema, and no comment at the v1/v2
  branch says which it is. Open question for whoever takes it: is any live
  artifact still emitted at `measured_runtime_prices.v1`, or is v1 now
  reachable only by hand?
- Re-emitting the 49-row Qwen3-0.6B table under the new binding spelling. It
  needs its receipts and the `--full-engine-report` it was emitted against, and
  it changes no measured number — but until it is done, that file on
  `/mnt/shared` is refused by `admit_native_rows`.
- No served or GPU measurement: this is a producer/consumer contract change,
  and no v2 table can price a device budget either before or after it (D37).

https://claude.ai/code/session_01GmYxSxhmzbBcBoFTjbLi1G


RobTand:
CI is RED on \`e68a610d\` and the failure is a real defect in this PR, not a stale test.

\`\`\`
FAILED tests/test_native_receipt_table.py::test_emitted_table_is_admitted_by_admit_native_rows_and_refused_only_at_fixed_resources
  assert 0 == 2
  {"admission": {"refusal": null, "status": "admitted"}, "rows": 1, ...}
\`\`\`

The test asserts the emitter CLI exits 2. After the split it exits 0 — and, worse, it
reports \`"status": "admitted"\` with \`"refusal": null\` for a table whose fixed
resources are still refused. Splitting the flags was right; losing the refusal at the
CLI surface is not. The emitter now certifies an admission it does not have.

\`admit_runtime_provenance\` no longer raises on the fixed-resource refusal, it returns
it, and this caller never reads the returned \`fixed_resources_refusal\`. The fix is to
surface both flags and the refusal text in that JSON and to decide deliberately what
the CLI exit code means once the two gates can disagree. Then update the test to the
new contract, rather than the other way round.

Worth noting why the PrismaBuild arms missed it: they ran on \`pq-cpu312\`, which is
pinned to an older Tessera contract, so a failure set from that venv is not the same
object as hosted CI. The house rule is to compare failure SETS against a same-venv
\`origin/main\` baseline, and a green count there does not substitute for CI.

Not merging. Fix forward on this branch.

https://claude.ai/code/session_01GmYxSxhmzbBcBoFTjbLi1G

Issue #604, opened 2026-09-14
Retain bounded PWC candidate renders across repeated COST probes


RobTand:
A joint COST candidate window needs to reuse the same rendered weights across four probes. The existing `ProductionWeightCache.resident_window` admits at most one prefetch quantum (`max_workers` keys), so larger candidate rosters repeatedly load and release files.

Extend the existing PWC owner with an opt-in retained lifetime. Preflight only the selected cache keys before loading; count all resident backing storages, including unrelated aliases, under an aggregate cap; bound each call to the existing prefetch pool by `max_workers` and a concurrent serialized-buffer cap. Keep loaded tensors in PWC through repeated passes, preserve file stat/SHA receipts and resident-only lookup, then release selected disk-backed owners on success or failure. Do not add another cache or change the legacy window default.

Acceptance: focused CPU lifecycle and integrity tests pass through PrismaBuild; architecture documentation records the contract. This issue covers only the shared PWC API; the combined COST planner and probe loop are separate work.

RobTand:
Closing: the shared PWC retained-lifetime API this issue asks for is on main.

- Landed as bd30b87b7c ("Retain bounded PWC candidate windows across probes") through PR #612, merged 2026-09-14T07:30:31Z as aab3255fd1. PR #606 carried the same commit and was closed as superseded by #612.
- API: `ProductionWeightCache._retained_window_preflight`, `plan_retained_window`, `retained_key_costs`, `retained_window` (`prismaquant/production_weight_cache.py:574-659` on main).
- Architecture contract: `docs/ARCHITECTURE.md:148-149` ("opt-in retained PWC candidate windows").
- Acceptance re-run on aab3255fd1 through PrismaBuild: `tests/test_pwc_resident_windows.py` 37 passed, 0 failed (PB `238b4501163d`, x86 pq-cpu312, CPU, priority -10, executed on dl380g10 in 9s).

The combined COST planner and probe loop were out of scope here and landed separately (#612 retained COST windows, #614 read-schedule test).

PR #606, opened 2026-09-14
Retain bounded PWC renders across repeated candidate probes

RobTand:
A candidate set larger than `max_workers` currently needs several `resident_window` lifetimes, so each repeated COST probe reloads the same rendered weights. This adds an opt-in retained lifetime to `ProductionWeightCache`: selected files are preflighted before any load, then the existing prefetch pool loads bounded quanta while PWC keeps all selected tensors resident through repeated passes. The context releases selected disk-backed owners on exit, including failure.

`retained_key_costs` prices only selected keys for the combined planner. Admission counts all existing backing storages and incoming archive storage under the resident cap, plus a separate cap on each concurrent serialized load quantum. Checked file-page advice now follows each completed quantum, once after a shared file's last read, so later loads need not accumulate avoidable page-cache residency. The optional main-thread `before_load_quantum` callback reports current PWC resident bytes, remaining incoming archive storage, and the next serialized-buffer size. The caller can check actual host memory and refuse the next load if page advice did not reclaim enough. Existing file stat/SHA receipts, CB validation, resident-only lookup, aliases, and LRU refusal remain on the actual reads. The legacy single-quantum window is unchanged.

Validation: PrismaBuild CPU action `689f43213f3a45296463dd517afc38f8767360cd6f610730afd3f1c7c847775f` passed 65 focused cache and architecture tests on dl380g10; compile action `cafef352bf96ec8c38d11677241a4be222d8638396aa0b1f4740ad4d0916dcf0` passed. No GPU fit or throughput claim is made here; the combined panel measures that separately.

Refs #604.


RobTand:
Superseded by #612 (merged as aab3255fd1da769b7f705d0011ab6ca35a91e9be), which carries both commits of this PR: 1c8a869c48 (same patch as 9d8430bd2a) and bd30b87b7c (same PWC retained-window code as ef788eb8d8), plus the prepared-render checksum refusal. Closed by main-campaign during the release-queue takeover; branch kept.

Issue #237, opened 2026-09-05
[P2] Connect joint AURA and prefill-aware selection for overlapping Tessera families


RobTand:
Rob requested a quality/prefill control for overlapping Tessera weight rates across activation contracts (for example W4A4/W4A8/W4A16), explicitly accounting for downstream propagation through AURA. This is a capability and validation gap; no new current served regression is claimed.

## Current source evidence

At audited PrismaQuant `d7fc82b8`:

- `prismaquant/aura_cost.py:8–23,1426–1451,2630–2714` projects rendered **weight** deltas through KL/Gauss–Newton adjoints. Candidate activation quantization is absent. The limitation is explicitly documented in `docs/ARCHITECTURE.md:4743–4750`.
- `prismaquant/production_weight_cache.py:1653–1666` measures the joint local residual `Xhat @ What.T - X @ W.T`, but reduces it to scalar MSE. Tessera campaign uses this currency; `allocator_candidates.py:1319–1354` applies scalar sensitivity. It is not directional joint AURA.
- `aqua_activation_cost.py:244–263` is a separate diagonal activation approximation. It does not retain the joint weight/activation cross terms and cannot simply be relabeled as the KL-adjoint currency.
- `allocator_solver.py:78–123` carries bytes and predicted loss, without runtime in its DP state. `allocator_candidates.py:1924–1971` / `tessera_menu.py:1306–1366` prune across bytes/loss and keep one lowest-loss row per byte bin. A faster same-byte higher-loss activation route can disappear here.
- Existing `--slo-prefill-p95-ttft-ms` and decode controls (`allocator.py:2075–2113`) filter proposed assignments later (`serve_constraints.py:85–119`, `allocator.py:4784`). They cannot recover those unproposed alternatives. `speed_quality_frontier.py` has the right constraint policies but only declared speed hints, without calibrated Tessera serving timings or connection to this search.

## Bounded design and acceptance

1. Reuse the streamed AURA cotangents and existing production weight/activation cache. Project each route's full local output residual before squaring: `a[k] = <G_Y[k], Xhat What.T - X W.T>`; price `0.5 mean(a[k]^2)`. The residual includes `X dW.T + dX W.T + dX dW.T`. Bind source, calibration, actual rendered bytes, activation contract/scales, arithmetic, and probes. Prevent a second scalar-Fisher or AQUA application.
[...]

PR #667, opened 2026-09-17
Price a whole routed MoE owner per rank: per-rank vectors, rank budgets in the search, GLM-288 geometry

### LINKED ISSUE #658: No routed-stack row path for GLM's 288-expert MoE: most of the prefill bill cannot be priced
Found while preparing the GLM prefill-cost measurement path for the
AURA+AQUA+prefill allocation (`astra-takeover-20260915/joint-aura-resume/`,
`PREFILL-MEASUREMENT-READINESS.md` and `PREFILL-MEASUREMENT-RUNCARD.json`).
Read-only inspection of `main` `2fa95995bde188f80fd88c6dc24819694c42b16f`; no
measurement was taken.

## The gap

The campaign's prefill constraint has to price one complete routed stack per
layer. Nothing in the tree can emit a table row for one, at any geometry:

`solid` `prismaquant/native_moe_panel.py:38`

```python
if shape["experts"] != 32 or not 1 <= shape["top_k"] <= shape["experts"]:
    raise ValueError("native MoE panel supports one complete 32-expert stack")
```

GLM-5.3-Flash has `n_routed_experts = 288`, `n_shared_experts = 1`,
`num_experts_per_tok = 8`, `moe_intermediate_size = 2048`
(`/mnt/shared/models/GLM-5.3-Flash-BF16/config.json`).

`solid` `prismaquant/native_moe_panel.py:76` requires the stack unit name to
match `model\.layers\.[0-9]+\.feed_forward\.experts`; GLM's units are
`model.language_model.layers.N.mlp.experts` (`source-tensors.json`,
`ddfe1a89786c547aef317101600bda466d9fe50cdc86540d0abc8254229bec70`).

Even a frozen routed panel has no row path: `prismaquant/native_receipt_table.py`
binds `tessera.native_dense_panel.v1` only (`:3`, `:74`, `:221`) and
`prismaquant/native_operator_panel.py:21` declares the only `owner_kind` it can
carry (`single_dense`).

## Impact

The roster is 36,423 units over 45 layers, of which layers 3-44 are sparse
(`mlp_layer_types` in the model config); those routed experts are most of the
model. The allocation would price its prefill constraint on the dense units
only and read a partial prefill bill as if it were whole-model. Design §8.1
forbids the workaround of summing isolated expert matmuls and calling it a
whole-MoE price.

Second, independent prerequisite for the same measurement: the documented
first positive v2 resource adapter is TP1-only, and TP2 is the served topology
(`A4-TP2-PREFLIGHT-20260916.md`; `nvfp4_moe_route.py:374-381` cuts the
intermediate axis). A rank sum or rank maximum must not be written into today's
v2 fields, and `allocation/prefill-contract-audit.md` records that this needs a
versioned resource-vector plus solver/feasibility extension. Fixed resources
remain refused for every v2 table under debt D37 (#560, #572).

## Options

1. Tessera publishes a GLM-geometry routed panel/receipt contract, and PQ gains
   the routed row path in `native_receipt_table` (preferred: PQ must not
   declare a producer contract the producer has not agreed to).
2. PQ widens `native_moe_panel`'s geometry and naming to the 288-expert
   `…mlp.experts` stack by explicit decision, accepting that the panel schema is
   shared with the Tessera producer.
3. Keep the routed axis unpriced and say so wherever the frontier is read. This
   is what happens today; it is listed only so the trade is on the record.

## Not measured

No panel was prepared, no receipt was requested, no GPU was used, and no timing
or memory number is claimed. This is a source-and-artifact finding plus the
prepared run-card that would execute option 1 or 2.

## Why this was filed rather than fixed here

The fix is not mine to make: it changes a PQ↔Tessera producer contract (panel
geometry, unit naming, receipt schema) and moves which measurements the
campaign can price, which is a decision Rob prices. Widening the validator alone
would admit a geometry the Tessera producer has never been asked to run, so it
was deliberately left unchanged.

## Active campaign disposition — 2026-09-16

Producer-led implementation is authorized and underway in Tessera and PrismaQuant. Geometry/routing fixes and CPU regressions are partial progress only. Remaining acceptance includes a runnable native A4/A8/A16 whole-routed operator measurement path under TP2, routed receipt-table emission, integration with the existing atomic packed-group allocator, and correctly represented per-rank resources. Whole-MoE latency must be measured; isolated expert latency sums are not a substitute. No further product decision is pending for this scope. The issue stays open until implementation, measurement and integration satisfy that scope.

RobTand:
## What this prices, and what it refuses

A tensor-parallel routed owner has no scalar answer to "how many device bytes
does this row cost": every rank holds its own share, so a rank sum and a rank
maximum are both numbers no device ever held. This closes the **minimum complete
producer -> table -> existing allocator path** for the campaign's own geometry
(GLM-5.3-Flash: 288 routed experts, top-8 `noaux_tc` with the live FP32
correction bias, routed scale 2.5, SwiGLU clamp 10.0, served TP2), and keeps the
32-expert LFM panel byte-identical in behavior.

Consumer and emitter (`runtime_provenance`, `native_receipt_table`,
`measured_runtime_prices`):

* `routed_owner_rank_resources` reads the rank **roster**, recomputes each
  rank's own resource digest, checks every other rank's against the copy that
  rank gathered before it was allowed to time, and builds
  `prismaquant.runtime_rank_resources.v1` -- one record per rank plus one
  whole-owner timing pair (`slowest_rank_median_of_one_whole_owner_apply`, with
  every rank's median kept beside it; never a sum of leaf timings, never a world
  mean).
* The module's **wire extent is charged once** (one canonical container, sharded
  locally), from the frozen panel's own member wire records -- not once per rank
  that views it. Regression: two ranks referencing one 100 B container price
  100 B, not 200 B.
* The timed region is accepted on the producer's **count** of the runtime's own
  `tensor_model_parallel_all_reduce` at
  `vllm.model_executor.layers.fused_moe.runner.moe_runner:_maybe_reduce_final_output`
  (`collective_calls_per_phase`: once per priced phase above a world of one,
  never at a world of one), with `includes_output_collective` read against those
  counts. A receipt that priced the quant method's partial sum without the
  runner's reduction is refused by name -- which is what the pre-fix TP2 arm
  did.

Solver and allocator (`allocator_solver`, `allocator`, `serve_constraints`):

* `solve_runtime_frontier` carries three coordinates **per rank** (summed
  residency, per-rank scratch peak, per-rank activation peak) in its dominance
  vector and filters each rank against its own budget **inside the fold**, so an
  alternative a rank's budget would have accepted cannot be pruned before that
  budget is read. Test: a faster candidate that breaches rank 1 and a slower one
  that fits -- the feasible one survives the frontier.
* Per-rank budgets arrive as one versioned object
  (`prismaquant.runtime_rank_device_bounds.v1`) beside the fixed whole-engine
  charge per rank, whose provenance is either
  `recomputed_full_engine_partition` (an axis that admits) or
  `pending_measurement` (prices the rank dimensions, admits nothing). A declared
  zero charge is refused rather than read as a charge. Scalar v2 rows keep their
  meaning byte for byte; a scalar row under a TP>1 context, and a ranked row
  meeting a scalar device budget, are refused by name.
* A ranked row publishes **no device total** (`device_memory_total_published:
  false`, per-rank totals instead), and the allocator refuses a scalar
  `--serve-device-budget-bytes` for such a table.

## Geometry

The task's main objective is not the LFM fixture, so this port carries the
versioned GLM geometry (`native_moe_panel`, cherry-picked from
`/home/rob/tmp/pq-glm-routed-owner-20260916` @ `4255d5a749` with its own
`tests/test_native_moe_glm_geometry.py`): geometry is an explicit versioned
object, the TP cut is declared (rank-local intermediate 1024 at TP2 where TP1 is
2048), and LFM keeps its exact field set and refusal text.

## Validation (CPU, PrismaBuild)

`pbtest --shards 4`, all green, 497 tests across
`tests/test_runtime_rank_resources.py`, `tests/test_native_receipt_table_routed.py`,
`tests/test_native_receipt_table.py`, `tests/test_runtime_provenance.py`,
`tests/test_measured_runtime_prices.py`, `tests/test_runtime_admission_split.py`,
`tests/test_allocator_runtime_frontier.py`,
`tests/test_allocator_measured_runtime_cli.py`, `tests/test_prefill_frontier.py`,
`tests/test_native_moe_panel.py`, `tests/test_native_moe_glm_geometry.py`,
`tests/test_native_operator_panel.py`, `tests/test_architecture_doc.py`,
`tests/test_docs_staleness.py`.

End-to-end regressions in `tests/test_native_receipt_table_routed.py`:

* GLM-288 TP2: two rank receipts -> one atomic member-assignment row with
  rank-local member shapes (1024x4096 / 4096x1024), one canonical container
  charge, slowest-rank median, then the **loader's own intake gate** re-deriving
  the whole vector from the roster. That gate is the one `allocator.main`
  reaches through `load_measured_runtime_table`.
* Refusals: a peer receipt that declares the wrong rank, a peer whose bytes
  moved, a peer receipt from another runtime, a rank vector the receipts do not
  support, a world the context does not price, a scalar device budget for a
  ranked table, and a pending per-rank charge (prices the dimensions, admits
  nothing).
* The allocator CLI prices a ranked table end to end (aggregate/expand path,
  round-tripped assignment, no device total).
* No GPU ran. Every receipt here is a synthetic CPU fixture and says so.

## What is not claimed

* **#658 and #237 are not closed by this PR.** Debt D37 is unchanged in
  substance: a ranked row still prices no device budget. The ranked device axis
  refuses today for two named reasons -- the fixed whole-engine charge per rank
  has no admitted value (pending measurement), and the runtime-global workspace
  has no versioned cross-row composition rule. That refusal is now specific
  rather than unconditional, and the versioned interface plus solver support
  exist and are tested, but nothing here admits one.
* No GPU evidence, no real TP2 receipt pair, no measured row. The producer's own
  two-rank gloo run is its evidence, not this PR's.
* `allocator.main` over a GLM *cost model* (the whole-owner serving unit and its
  864 member cost rows) is not assembled: the routed row is exercised through
  the loader's own intake gate, and the allocator CLI leg is exercised with the
  ranked row spelling. Composing a GLM cost fixture into a full allocator run is
  the next step, and it needs no new mechanism here.

## Dependency and alignment

Aligned to the producer's current contract (PR #534 then its local head
`37c9a2b`: `a66f7b3` moved the priced region onto the runner's own reduction):
`runtime.collective` is read as the declaration, `latency_scope`'s counted
calls as the evidence, and the callsite string is pinned on this side too.
Pairs with `/root/native_tp2_measurement_runtime`.

---

Refs #658. Refs #237. Neither is closed by this PR: debt D37 stands, and the ranked
device axis admits only once the per-rank partition capture below exists.

**Dependency (updated).** The producer's PR #534 is merged into Tessera master
`87233d14702e9ae71454cf3d5149e18be45b7004` (head `9d10e351`), which carries the
functional GLM-288 A4/A8/A16 TP2 core, the observed output collective and the rank
evidence; this PR consumes that counted-call contract. No GPU qualification is claimed
by that merge or by this PR. See the review comment for the follow-up engineering
(workspace composition rule, recomputed per-rank fixed charge) and the concrete
producer request for `prismaquant.full_engine_rank_partition.v1`.

Closes #658


RobTand:
Adds review follow-ups: the frontier's reported dimensions, the runtime-global workspace rule, and a per-rank fixed charge that is recomputed rather than declared.

### Bounds admission (review: "recomputed" was unsupported at intake)

`RankDeviceBounds` no longer accepts the admitted spelling from a payload at all. The one construction path is `RankDeviceBounds.recomputed(...)`, which compares every value with a `runtime_provenance.RankFixedCharge`: the verdict of reading a **sealed per-rank partition** (`prismaquant.full_engine_rank_partition.v1`) and cross-checking its per-term sums against the **sealed full-engine report this consumer recomputes itself**. A partition allocates the fixed charge across ranks; it may not create it.

Refusals covered: a declared charge with no recomputation behind it; a charge that is not the recomputed one; a world mismatch; a forged report digest; a foreign rule string; a term the report cannot recompute. The synthetic capture observes no fixed resident/activation/KV term, so the per-rank partition refuses by name rather than charging zero -- the mechanism is complete and the remaining input is a capture that observes those terms.

### Workspace composition (was a named refusal)

`RANK_WORKSPACE_RULE` charges one frozen workspace identity **once per rank** no matter how many priced rows view it, adds distinct identities, and refuses two records that disagree about the bytes behind one identity. A device total includes it instead of refusing a term that now has a rule.

### Diagnostics

`diag["dimensions"]` is built from the axis lists rather than slicing the combined vector, which labelled rank 0's residency as the scalar device term whenever the decode axis was absent (one axis too many, and the wrong one).

### Validation

PrismaBuild CPU, `pbtest --shards 2`, 478 tests green across `test_runtime_rank_resources.py`, `test_native_receipt_table_routed.py`, `test_runtime_provenance.py`, `test_measured_runtime_prices.py`, `test_runtime_admission_split.py`, `test_allocator_runtime_frontier.py`, `test_allocator_measured_runtime_cli.py`, `test_native_receipt_table.py`, `test_prefill_frontier.py`, `test_native_moe_panel.py`, `test_native_moe_glm_geometry.py`, `test_architecture_doc.py`, `test_docs_staleness.py`. The full suite ran green before this follow-up: 8/8 shards, 9,433 passed / 208 skipped / 3 xfailed.

### Producer dependency and request

Dependency updated: PR #534 is **merged** into Tessera master `87233d14702e9ae71454cf3d5149e18be45b7004` (head `9d10e351`), which carries the functional GLM-288 A4/A8/A16 TP2 core, the observed output collective and the rank evidence. This PR consumes the counted-call contract and pins the callsite `vllm.model_executor.layers.fused_moe.runner.moe_runner:_maybe_reduce_final_output`; no GPU qualification is claimed by that merge or by this PR.

Concrete request to the native TP2 measurement worker (the source of the pending per-rank charge): publish, per TP2 run, a `prismaquant.full_engine_rank_partition.v1` document beside the full-engine report -- `world_size`, the report's `{path, sha256}`, the report's `partition.capture_sha256`, `rule: sealed_per_rank_terms_summing_to_the_recomputed_whole_engine_terms`, and one `{rank, terms}` row per rank whose `fixed_resident`, `fixed_activation`, `fixed_scratch` and `fixed_kv` are that rank's own observed bytes. Additionally the report's own observations must make those four terms recomputable (they are `None` today), so the two documents can be cross-checked. That is the whole blocker for the ranked device axis; no further consumer-side rule is missing.

### Still not done in this PR

* `allocator.main` over a **GLM cost model** (the whole-owner serving unit plus its 864 member cost rows). The routed row is exercised through the loader's own intake gate, and the allocator CLI leg with the ranked row spelling; the GLM cost fixture is the next step and is reviewable separately.
* No GPU evidence, no real TP2 receipt pair, no measured row.

Refs #658. Refs #237. Neither is closed by this PR: debt D37 stands, and the ranked device axis admits only once the capture above exists.

RobTand:
**Review follow-ups accepted; recorded here so they are durable on the PR** (assigned after `daac8a0e16`, not yet implemented -- this branch is handed over at that head):

1. **Workspace must be composed INSIDE the search.** `compose_rank_totals` adds the workspace after the frontier is built, so a fast option with a huge workspace can still prune the slower option that fits. Recipe agreed: keep a per-rank `identity -> bytes` map in the solver state, union it during the fold, reject one identity at two sizes, and compare dominance **only within an identical workspace identity set** (numeric axes as today) -- the simple equality bucket, reusing the existing frontier bounds. Regression to add: fast-but-over-workspace versus slower-that-fits, where the feasible one survives.
2. **`recompute_rank_fixed_charge` currently validates only the per-term SUM** against the whole-engine report, so an arbitrarily redistributed (rehashed) partition passes: it binds no per-rank observation. `prismaquant.full_engine_resource_report` is scalar today and has no rank contract, so the fix is NOT to treat a sum as physical evidence. Each rank must reference **its own actual capture/report**, whose identity names rank, world and runtime, and whose four fixed terms this consumer recomputes; the whole-engine cross-check is optional, never a replacement. Use the existing full-engine capture instrumentation and extend the contract there; measurement is then the only thing pending, until that capture runs.
3. **`RankDeviceBounds.recomputed` must not duck-type.** `getattr(recomputation, ...)` accepts a `SimpleNamespace` with arbitrary values; require the actual `runtime_provenance.RankFixedCharge` type (or an equivalently minimal trusted construction) and refuse anything else by name.
4. **A legitimate CLI path must exist.** There is no `rank_devices` wiring in `allocator.py` today, so the admitted path is unreachable in production. The GLM 864-member cost-model -> `allocator.main` end-to-end test will force that integration: the allocator's DP unit for a whole owner is the aggregated serving group (`allocator_candidates._PACKED_GROUP_MARKER = ".__packed_serving__."`), whose name differs from the native panel's unit, so the row's key and the group unit must be reconciled in the same change. No permanent-pending design: the ranked device axis is refusing today only because the per-rank capture does not exist yet.
5. New public report contract updates `docs/ARCHITECTURE.md` in the same commit.

**Producer requirement (to relay to `/root/glm_owner_input_producer`).** Per TP2 run, publish `prismaquant.full_engine_rank_partition.v1` beside the full-engine report: `world_size`; one `{rank, terms}` row per rank whose `fixed_resident`, `fixed_activation`, `fixed_scratch` and `fixed_kv` are that rank's OWN observed bytes; and, per rank, a reference to that rank's own capture/report whose identity names rank, world and runtime (so the terms are recomputable from that rank's evidence, not inferred from a world total). The full-engine capture's observations must make those four terms recomputable for that rank -- they are `None` today (`OWED_OBSERVATIONS` in `full_engine_resource_report.py`: `kv_observations`, `owner_views`, ...), and extending that instrumentation is producer work, not a consumer refusal.

PR #696, opened 2026-09-17
The Tessera menu refusal names its own cause; the D37 pair is one measurement

RobTand:
Closes #572. Leaves #570 open with its decision recorded and its blocker named.

The two issues are one debt seen from two sides, and the measurement below is
what makes that an identity rather than a phrase.

## 1. The first refusal today is neither of them

Measured through PrismaBuild on `dl380g10`, CPU, `--priority -10`, against the
only real `measured_runtime_prices.v2` table
(`.../frontier-qwen3-0.6b-20260913/scoped-sweep-20260913/qwen3-0.6b-layer0-all.json`,
sha256 `d841ed57f555…`, 49 rows) under `--measured-runtime-fixed-scope shape-only`.
A scoped arm (PB `520ef95b1c68`) and an unscoped arm (PB `0c06c65fa611`) stop at
the same line:

```
[alloc] ERROR: measured runtime: native producer admission refused: native panel:
this panel priced an activation residual through PrismaQuant's own activation
quantiser and carries no attestation that the runtime executes the same rounding
rule.
```

`require_panel_activation_attestation`, `prismaquant/native_operator_panel.py:424`
(refusal `:441-449`), reached from `consume_native_receipt:389` and surfaced by
`admit_native_rows` at `prismaquant/runtime_provenance.py:1258`.

42 of 49 bound panels refuse and 0 of 49 carry the stamp. The seven that pass are
the `TESSERA_BF16_K1` rows, whose format does not quantise its input, so the split
is the format's (PB `60ca5b6fddf6`). This is the re-freeze D37 already owes,
arriving one gate ahead of the charge boundary.

## 2. The 2026-09-17 `operator_route` finding was a stale file

Two tables share a name. `table/qwen3-0.6b-layer0-all.json` (sha256 `9bbb283a2ea4…`)
carries `binding.operator_route: "torch.mm"`, the symbol-only spelling #565
replaced. `scoped-sweep-20260913/qwen3-0.6b-layer0-all.json` (sha256 `d841ed57f555…`)
carries the route-class spelling and clears `native row operator binding` on
current `main`. A probe that reported that check as today's first refusal was
reading the older file. Verdict: **stale recorded table, not a main-side
regression.**

## 3. #572's three hypotheses, answered

Read from the packaged `runtime_contract.json` the pin names (sha256
`db9ca4c0c457…`, PB `33c8064e7b64`), not from any reviewed copy:

| family | attested rung |
|---|---|
| `TESSERA_E2M1_K2` | `R896` |
[...]

PR #695, opened 2026-09-17
Require an owner class exactly where a composition term reads one

RobTand:
Closes #697. Consumer half of RobTand/tessera#478. **Land it with the Tessera half** (RobTand/tessera PR for `flash/478-owner-classification-20260917`): the two sides recompute the same partition, so a producer that splits its rows this way and a consumer that does not will disagree on every report.

## What was wrong

`_classify` asked the ownership question first and refused a row with no owner category before it ever asked what the row's lifetime was.

No composition term charges an allocation proven live during no declared engine step. `_recompute_terms` selects over the membership rows, `_charged` reads that same list, and the off-step peak is a sweep over lifetimes. So nothing in the consumer reads such a row's owner class — and every term went null for a field nothing consumes. That is a refusal on missing information, not on a contradiction.

## The rule

An owner class is an invariance claim, and only a term that charges a row needs one.

- A row proven live during no declared step is classified off-step on its lifetime alone, and carries `owner_class: null` when no checkpoint census saw it. The report reader accepts a null class on off-step rows only.
- Every row a term charges still needs exactly one supported class. A `shared` boundary tensor is unclassified exactly as before, and an unowned row that is live while a declared step runs still nulls every term.

The off-step proof never used ownership: it is complete step coverage plus the row's own liveness, both read from the observations.

## Before and after, on the real capture

Recomputed from the a5 raw ledger (`/mnt/shared/tessera-runs/receipts/399-qwen3-0.6b-20260913/report-a5/ledger.json`, 21,556 rows, complete step coverage) through both classifications, through PrismaBuild action key `71418237b8d7be1b53fd6e58caa2dac11957820fca3c10fa00c775222f2f5d39`:

| | before | after |
|---|---:|---:|
| unclassified rows | 21,104 | 1,276 |
| unclassified bytes | 80,279,393,776 | 731,857,273 |
| off-step rows | 0 | 19,828 |
| off-step bytes | 0 | 79,547,536,503 |
| uncharged rows | 224 | 224 |
| terms expressible | none | none |

Every term and every domain on that capture is unchanged. This removes a blocker; it closes no domain.

## Tests

`tests/test_full_engine_resource_report.py` + `tests/test_runtime_fixed_resource_admission.py`: **207 passed** through PrismaBuild on sparklina, action key `7ff7f63352bf024598ea0c5b3ded3179f46c8941e9e41087dc2c6eb9b423538f`.

[...]

PR #698, opened 2026-09-17
A PWC resident quantum is as wide as its admitted bytes, and prices each file once (#693)

### LINKED ISSUE #693: Joint-AURA prepare is single-threaded: 9 of 10 threads do zero work, 3.01 TB read = 5.1x the model
## Summary

`tessera_joint_aura prepare` runs single-threaded. Nine of its ten threads do
literally no work, and the one that does is the only thing driving a 3 TB read.

## Measured (sparky, PID 232253, PB action `8b53c37c09e9…`, GLM-5.3-Flash-BF16)

Steady state over 4h58m: **3.01 TB read = 5.1x the entire 592 GB model**, at
168 MB/s, 49% of one core, GPU 13.5 W of a 140 W envelope (9%).

`py-spy record`, 300 s at 100 Hz, 215,079 samples (receipt:
`/home/rob/dq-runs/salvage/joint-aura-perf-20260917/prepare-300s.speedscope.json`):

| thread | busy | idle | % busy |
|---|---|---|---|
| MainThread | 23,757 | 6,242 | **79.2%** |
| layerread_0..3 | **0** | 119,996 | **0.0%** |
| prefetch_0 | **0** | 29,999 | **0.0%** |
| ThreadPoolExecutor-33287_0..3 | **1** | 35,084 | **0.0%** |

**Total: 23,758 busy of 215,079 samples (11%).** The pools are constructed and
never submitted to. `PRISMAQUANT_LAYER_READ_THREADS=4` and
`--streaming-prefetch-workers 1` allocate threads that never receive a task.

MainThread phase share (busy samples only):

| share | phase |
|---|---|
| 42.2% | other |
| 15.9% | `load_verified_activation_cache_entry` (perturbed_x_cache.py:638-641) |
| 14.5% | `verify_anchor_render` (tessera_joint_aura.py:805) |
| 7.7% | `_EndRecData` (zipfile __init__.py:315) — EOCD back-scan per `torch.load` |
| 7.2% | `verify_cached_unit` (cached_unit.py:366) |
| 5.0% | `_cb_cache_tensor_identity` (production_weight_cache.py:3659) |
| 4.5% | `digest_host_tensor` (cached_unit.py:54) |
| 2.2% | `verify_plane_region` |
| 0.9% | `sha256` |

79.2% busy against 49% of one core means MainThread alternates CPU work with
blocking reads and **nothing overlaps**.

## Contributing: the read path is cold and the prewarm is self-evicting

Cold read scaling, sparky -> dl380 over NFS/RDMA (10.100.98.3, `proto=rdma`,
`nconnect=16`, `rsize=1M`): 1 stream 28.6 MB/s, 2 -> 40.5, 4 -> 86.9,
8 -> 134.2, 16 -> 204.7 MB/s. **Warm re-read: 12,142 MB/s.** The fabric is not
the limit; cold seeks on dl380's 4-disk raidz1 are.

dl380 ARC `c_max` 240 GiB / `size` 175.7 GiB; L2ARC 1.64 TiB with **both cache
devices 100% full**, `l2_write_bytes` 12,091 GiB to fill 1.5 TB (8x turnover),
hit rate 51.4%. The job's prewarm manifest declares **469,008 entries /
6,840 GB** — ~3.6x the whole ARC+L2ARC tier, so the prewarm evicts itself.
Tracked separately as prismabuild#499.

## Proposed fix, in evidence order

1. **Feed the pools that already exist.** Overlap the next unit's load with the
   current unit's verify. `prefetch_capture`
   (tessera_calibration_cache.py:1240-1300) is a plain sequential
   `for name in names:` on the calling thread.
2. **Right-size the prewarm** to the measured cache tier so the working set is
   warm (70x on the read leg). prismabuild#499.
3. **Kill the per-`torch.load` zipfile EOCD back-scan** — 7.7% for nothing.
4. **Cache the identity digests.** `digest_host_tensor` + `_cb_cache_tensor_identity`
   + `verify_cached_unit` + `sha256` = ~17.6% combined. These are pure functions
   of bytes that never change; they can be computed once (Rob has authorized
   doing this a priori on dl380) instead of on every run.
   Note: packed-expert targets are `weight[expert_id]` views of a split packed
   tensor (routed_experts.py `profile_declared_packed_expert_projections`), so a
   sidecar must replay the split geometry, not hash raw file offsets.
5. Hygiene, not throughput: `digest_host_tensor` (cached_unit.py:45) does
   `value.view(torch.uint8).numpy().tobytes()` — a full extra copy per tensor.

## Acceptance

Profile before AND after (principle 15); the delta is the claim. Before-state is
the receipt above. Target: busy-sample share well above 11%, with the read leg
overlapped rather than serialized.

RobTand:
Closes #693.

## Headline: the measured win is 1.09-1.19x, not the ~3x the issue expected

That gap is the finding, not a shortfall to explain away. The ~3x estimate
assumed the loads were serial and would become parallel. They were already
parallel (see refutation 2 below), so nothing about I/O concurrency changed
here. What this PR removes is a serial MainThread cost - the archive scan - and
a quantum-width rule that made each pass re-pay it. On the A/B bench that is
worth 1.094x, 1.185x and 1.096x across three box-load conditions, with the
deterministic counts halving (16 quanta -> 8, 120 archive reads -> 40). The
production effect is larger than the bench because the bench runs warm, but it
is a lower bound, not a projection, and no end-to-end campaign number is
claimed.

**On the dl380g10 pool view.** The retargeted acceptance criteria asked for one.
There is no arm to show. The pool view would compare read concurrency between
the arms, and read concurrency is identical: both arms issue the same 40 render
`open()`s from the same 4-thread pool, which is exactly what refutation 2
establishes. The bench was also deliberately run on an ARC-resident working set
so the two arms differ only in scheduling, not in what the pool must fetch. A
pool trace here would show two indistinguishable traces and would not be
evidence of anything. The box-level instrument that does have signal - CPU,
iowait and GPU power on both boxes - is reported below.

## Read this first: two of the issue's measured claims are refuted

Both come from the analysis scripts in
`/home/rob/dq-runs/salvage/joint-aura-perf-20260917/`, not from the code. The
receipts that refute them are the live prepare's own telemetry and a corrected
re-read of the same `prepare-300s.speedscope.json`.

**1. The windows were not one key wide, and bytes were not what closed them.**
The running prepare prints its own window receipts. Every five-render unit:

```json
{"unit": "…layers.19.mlp.experts.0.down_proj", "keys": [4 keys], "loaded": 4,
 "resident_bytes": 67108864, "budget_bytes": 536870912,
 "load_buffer_capacity_bytes": 67119001, "load_buffer_budget_bytes": 536870912}
{"unit": "…layers.19.mlp.experts.0.down_proj", "keys": [1 key],  "loaded": 1,
 "resident_bytes": 16777216, "budget_bytes": 536870912}
```

A quantum held **67,108,864 of 536,870,912 bytes — 12.5%**. The renders are
16,777,216 bytes each, not 512 MB, so `nbytes + incoming > max_resident_bytes`
never tripped. What closed the quantum was `len(window) == max_workers` with
`file_hash_workers = 4`. A nine-render `shared_experts` unit split 4 + 4 + 1.

Raising the byte budget alone would have changed nothing.

**2. `_load_file_tensor` does not run on MainThread.** All 7,656 of its samples
are on the four `ThreadPoolExecutor` threads, every one reached through
`_load_one (production_weight_cache.py:1304)`; MainThread has zero. The
"5,177 self samples on MainThread" came from `parse.py` counting `self_t[s[-1]]`
across every thread and reporting the total as MainThread's. The
"pool threads: 0 busy samples" came from `parse2.py` classifying any stack
containing `_worker` as idle — which is every pool sample, running or not. Per
thread, by leaf frame:

| thread | samples | leaf-idle | busy |
|---|---|---|---|
| MainThread | 29,999 | 6,253 (20.8%) | 79.2% |
| ThreadPoolExecutor ×4 | 8,527–9,220 | 66.8–69.0% | ~33% each |
| layerread_0..3, prefetch_0 | 29,999 each | 100% | 0% |

Corrected total busy is **35,083 of 215,079 (16.3%)**, not 11%. The "9 of 10
threads idle" headline holds only for `layerread_0..3` and `prefetch_0`, which
are the source-prefetch pool and are out of scope here.

Everything else in the issue stands, including the 0.377-of-a-core MainThread
figure and the GPU at 13.5 W of 140 W.

## What this changes

**A quantum is as wide as the two byte budgets its caller admits.**
`plan_resident_windows` no longer closes a quantum at `max_workers` keys. It
closes on the resident cap and on a new explicit `max_load_buffer_bytes`
serialized cap, which `resident_window` already enforced after planning and now
also splits on. `max_workers` stays what it always was underneath: the loader
concurrency, which `prefetch` bounds on its own pool whatever key count it
receives.

**The derivation, which is the guard's own.** `prepare_cache`'s
`reserve_allocation('before_joint_qualification_unit:…')` already prices this
unit's residency at `max_render_bytes` and its serialized buffers at
`policy['max_load_buffer_bytes']`, as two separate terms. No term it charges
scales with a quantum's key count: the capture payload is per unit, and the wire
reader is one lookahead, hence its fixed `2 * max_wire_read_bytes`. A full-width
quantum therefore spends bytes this unit was already admitted — or refused —
for, and the planner still splits before it can exceed either budget. Nothing
was raised; `prepare_cache` stopped passing `min(max_render_bytes,
policy['max_load_buffer_bytes'])`, which charged the serialized cap against
residency.

**The same seam on the `run` path.** `joint_statistics_replay.resident_candidates`
is how `run` reaches every render, and it planned against the same minimum, with
a comment calling it the conservative choice. It is not one any more. It now
passes both budgets, and the guard right below it is reserved on the same two
terms, so raising residency from the minimum to `max_render_resident_bytes`
makes that reservation larger or equal, never smaller.

**One archive scan per file per window lifetime.** Preflight reaches the same key
three times before anything is read — the caller's plan, the window's re-plan,
and the window's file table — and each reach opened the Torch archive and read
its central directory. On the live prepare that was **11.0% of MainThread wall
time**: 3,305 of 29,999 samples in `torch_archive_storage_bytes`, every one under
`_window_file` inside `plan_resident_windows`. The total is a pure function of
the file's bytes, so `_window_file` now remembers it under the identity this
cache already trusts for that purpose — the `cache_file_stat_signature` tuple
every window read re-checks. A file whose signature moved is a miss and is
rescanned. The memo is one small entry per distinct backing path, and it is
dropped when a window closes or the cache is compacted.

## Measured, before and after

Interleaved A/B, both arms in one process against one file set, so the only
difference is the code under test. Eight units × five 16.78 MB renders on the
same NFS export the campaign reads, `release_file_pages=True` as in production.
`experiments/pwc_window_width_ab_693.py`, run on sparklina through PrismaBuild.

**Deterministic counts, per pass over 8 units / 40 keys:**

| | quanta | archive-directory reads | render-file `open()` |
|---|---|---|---|
| before | 16 | 120 (3 per key) | 40 |
| after | **8** | **40 (1 per key)** | 40 |

Same files opened the same number of times: this removes metadata round trips
and serialization, not data.

**In-process profile (py-spy at 100 Hz, PB `--profile sample`, 30 repeats per
arm, one profiled action per arm):**

| | samples | MainThread share | pool-thread share | `torch_archive_storage_bytes` |
|---|---|---|---|---|
| before (`ff21d71d61c0`) | 2,583 | **28.7%** | 71.1% | **2.1%** |
| after (`ae6af2118f4e`) | 2,552 | **24.5%** | 75.5% | **1.0%** |

The work moved off the serial thread onto the pool, and the archive scan halved.
It halved rather than dropped to a third because this counter also sees the
in-memory `io.BytesIO` scan inside `_load_file_tensor`, which is one per key in
both arms: 3 + 1 → 1 + 1.

**Wall time, interleaved A/B/A/B:**

| run | repeats/arm | before median | after median | ratio |
|---|---|---|---|---|
| light box load | 5 | 0.4520 s | 0.4132 s | 1.094× |
| per-arm profiled | 30 | 0.5130 s | 0.4330 s | 1.185× |
| heavier box load | 6 | 0.8273 s | 0.7548 s | 1.096× |

**Bit identity, shown not asserted.** Each arm SHA-256s every tensor exactly as
the consumer receives it and reports one roster digest. Both arms, every repeat:
`c1e7e68143e7cf041e2788671a64b75cda3bc509d32fb32b84e9567cc4a061eb`. PB's own
byte counters agree: 18.9 GB read before, 19.2 GB after, for 30 identical passes.

**Box-level view, which no in-process tool can see.**
- sparky, over the live prepare's last six hours: GPU **13.16–13.61 W** of the
  140 W envelope (9.4–9.7%), min 12 W, max 19 W. This is a CPU and NFS-path
  change and the GPU is the instrument that says so — it was never the limit.
  `gpu_utilization` is non-diagnostic on GB10 and `utilization.memory` reads a
  hard 0, so power against the envelope is the reading.
- sparklina during the bench: CPU user peaked 44.3% in the legacy arm's bucket,
  `iowait` ≤ 0.64% throughout, so neither arm was stalled on the client.

**What is NOT claimed.** The bench reads a working set the dl380g10 ARC holds,
so its absolute wall times are warmer than the campaign's cold path. The ~1.1×
measured here is a **lower bound** on the production effect, not a projection of
it: on the live run the archive scan costs 11.0% of MainThread rather than the
2.1% it costs warm, and the one-key quantum reads at single-stream cold speed
rather than warm. No end-to-end campaign speedup is claimed, because none was
measured — the running prepare is a sealed request and was not restarted.

## Tests, through PrismaBuild at `--priority -10`

Before, `origin/main` plus these gates (3 shards, keys `0c3cffe69568`,
`8c40b455cf58`, `1c00dda425c1`): **7 failed, 85 passed**. The COST-path gate
fails with three 2-key quanta where the policy admits one 6-key quantum; the
prepare-path gates fail on the 4 + 1 split and on the three-per-key scan count.

Those failures are not all the same kind, and the difference matters for
"mutate the driver, not the fixture". Four of the seven fail **behaviourally**
on main, with an assertion that names the wrong schedule:
`test_window_width_follows_the_admitted_bytes_not_the_loader_count`
(`AssertionError`: two quanta where one is admitted),
`test_candidate_windows_hold_what_the_operator_policy_admits` on the COST path
(`AssertionError`: three 2-key quanta vs one 6-key), and both
`test_joint_qualification_windows` resume gates. Those are the ones that prove
the check bites. The remaining two,
`test_serialized_buffers_have_a_separate_aggregate_limit` and
`test_window_preflight_scans_each_archive_once_per_lifetime`, fail with
`TypeError: ... unexpected keyword argument 'max_load_buffer_bytes'`, which
proves only that the signature changed. They are reported for completeness, not
as behavioural evidence.

After, this branch, the impacted suite (5 shards, keys `cfa6bd113d23`,
`73667ddf13de`, `42e7acfce757`, `7eadc43af49b`, `a2630e20ee43`): **450 passed,
5 skipped, 0 failed** over `test_pwc_resident_windows`,
`test_pwc_file_load_receipts`, `test_production_weight_cache`,
`test_joint_retained_window_plan`, `test_joint_retained_streamed`,
`test_joint_retained_statistics_replay`, `test_joint_qualification_windows`,
`test_joint_cost_read_schedule`, `test_joint_cost_streamed_schedule`,
`test_joint_head_progress_678`, `test_joint_replay_frontier_607`,
`test_glm_cost_v2_data_manifest`, `test_glm_source_derivative`,
`test_tessera_joint_aura`, `test_tessera_joint_encoder_reuse`,
`test_capture_memory_guard`, `test_streaming_production_cache`,
`test_production_cache_render_identity`, `test_joint_operator_windows`,
`test_joint_operator_windows_profile`, `test_joint_aura_streamed`,
`test_joint_aura_packed`.

Three existing gates asserted the rule this removes and are re-pointed at the
property rather than the CPU count. Each still refuses, and none loads what it
refused:
- `test_serialized_buffers_have_a_separate_aggregate_limit`: the separate
  serialized cap is now shown by the planner splitting and the one-quantum window
  refusing, plus a single oversized buffer still refusing outright.
- `test_retained_window_keeps_more_keys_than_workers_for_repeated_passes`: the
  contrast between a retained lifetime and a resident window is its budget, not
  its worker count.
- `test_joint_qualification_windows` ×2: with `file_load_workers=1` the walk
  planned one key per window; it now plans one window per unit, and the gate
  asserts the unit roster as well as the count.

`docs/ARCHITECTURE.md` is re-stamped in each commit that changes behaviour, per
principle 13. No pipeline default, stage graph, format, lane, pin, plugin
contract, ship gate or published byte changes.

## Left open, deliberately

- **Overlapping load with verify.** `prefetch` is 24.8% of MainThread on the live
  run, all of it MainThread blocked on `pool.map` while the pool loads and then
  idle while MainThread verifies. Overlapping quantum N+1's load with quantum N's
  verify needs two quanta resident at once and changes what the guard prices, so
  it is a separate change with its own reservation.
- **Items 3–5 of the issue's comment** — prewarm sizing (prismabuild#499), the
  per-unit identity sidecar, and `digest_host_tensor`'s extra copy — are
  untouched here.
- The running prepare (PB action `8b53c37c09e9…`) was not restarted. It is the
  before-state.


RobTand:
### Re-validated after merging `origin/main` (twice)

`origin/main` moved under this branch while it was open — `f344e91bc9` (#696) and
then `dbd07ffe14` (#695). Both conflicted only on the `docs/ARCHITECTURE.md`
stamp header; both resolutions keep one `As of:` line and every stamp, newest
first. Nothing else conflicted, and `git diff origin/main...HEAD` is still
exactly the 8 files / 611 insertions this PR introduces — the merges pulled
main's work in, they did not move any of this PR's hunks.

Impacted suite re-run on the merged tree (`b1c99b3dec`) through PrismaBuild,
`--priority -10`, `--tag dl380g10`, `--python pq-cpu312-tessera-4c384e60`
(pin `4c384e6049dca3eeaf503bb2c9cd1cd2778978d1`, 87 files verified on every
shard), measured demand `mem_gb=2 cpu=4` against an observed `rss≈415 MB`:

| shard | action | receipt | result |
|---|---|---|---|
| 0 | `68bfae43d8dc` | `697f9ac11fa9…` | 97 passed |
| 1 | `a722abc0d0fc` | `7475be33a7cb…` | 61 passed, 2 skipped |
| 2 | `c26f340816cd` | `e61487a02e3e…` | 129 passed |
| 3 | `58a8005f8b7e` | `cb98248c98e9…` | 127 passed, 3 skipped |
| 4 | `6ffdd73c5c5b` | `2d61736d2640…` | 45 passed |

**459 passed, 5 skipped, 0 failed, 5/5 shards green.** 23 files: the 22 impacted
by this change plus `tests/test_tessera_menu_attested_count.py`, which the #696
merge touched.

RobTand:
### Principle 13's mechanical gates, and one stale doc found and flagged

`tests/test_docs_staleness.py` and `tests/test_architecture_doc.py` were not in
the impacted set, which was an omission: this PR edits `docs/ARCHITECTURE.md`
three times and then resolves it across two `origin/main` merges. Both run green
on the final head through PrismaBuild, `--priority -10`, `--tag dl380g10`:

| test | action | receipt | result |
|---|---|---|---|
| `test_docs_staleness.py` | `36d2bd29927c` | `e107acbe4ac8…` | 6 passed |
| `test_architecture_doc.py` | `557080a9c777` | `6570a9073390…` | 13 passed |

Sweeping the rest of `docs/` for the retired rule turned up one real stale page.
`docs/measurements/pwc-resident-windows-2026-09-08.md` is the dated #381 record
and still says `plan_resident_windows(keys, *, max_resident_bytes, max_workers)`
and "Each quantum has at most `max_workers` entries" — both now false. It is a
measurement record, so it is not rewritten; a dated note at the top says which
two sentences #693 retired and points at `docs/ARCHITECTURE.md` for the current
rule. The only other `min(max_render` / `len(window) == max_workers` occurrence
anywhere under `docs/` is inside this PR's own stamp, where it is describing the
rule being retired.

PR #700, opened 2026-09-17
Charge an anchor-batch slot the encoder working set it holds (#639)

### LINKED ISSUE #639: selected_anchor_resources charges ~0.19 GiB per anchor-batch slot; encoder measures ~0.43 GiB per slot
`selected_anchor_resources` charges about 0.19 GiB per anchor-batch slot; the encoder allocates about 0.43 GiB per slot.

## Measurement

GLM-5.3 E2M1 census, row-0055, layer 20. 192 distinct units, each encoded once warm and 4 times timed at widths 8, 16 and 32 (interleaved 8,16,32,32,16,8). Encoder `a4c92094`, PQ tree 9753a5b7c5, sparky (GB10). PB `fab0ab383f16`, rc 0. Every unit-encode was byte-identical to the stored census wire (960/960).

| width | `torch.cuda.max_memory_allocated` GiB | planned `resident_anchors` GiB | planned demand (GB) |
|---|---|---|---|
| 8 | 3.535 | 98.890 | 101 |
| 16 | 6.970 | 100.390 | 103 |
| 32 | 13.853 | 103.390 | 106 |

- Measured slope: (13.853 - 3.535) / 24 = 0.43 GiB per slot.
- Planned slope: 1.5 GiB per 8 slots = 0.1875 GiB per slot, from `encoder_memo_bytes` plus `compatible_batch_weight_bytes` (`prismaquant/autoscale.py:518` and `:539` on main `e217a7291f`).
- At width 32 that is about +10.3 GiB actual against +4.5 GiB planned. At width 16 a re-plan raises the cap by 2 GiB against about 3.5 GiB of real growth.

Results: `/mnt/shared/tessera-measurements/tessera486-fused-lut/batch-width-ab-20260915/summary.json` (`q4_demand_and_guard`) and `run1/result.json`.

## Scope and limits

- The harness encodes one layer's units outside the campaign, so its absolute peak is not the census row's peak. The claim is the slope only.
- The per-slot term `:530-539` already flags as a gap: the producer's working set inside `encode_linear` is not traceable from PrismaQuant and is charged at a fixed bound. This measurement says the bound is low by about 2.3x.
- Today's census runs at width 8, where the missing slope is about 2 GiB. Its rows have completed under the guard; their guard slack was not read here.
- `CaptureMemoryGuard` refuses an argv-only width change before encode (`tessera_campaign.py:4522-4525`), so an under-planned row fails closed rather than OOMs.

## Expected

Charge the per-slot term from a measured or attested encoder working set, not the fixed bound, and assert the guard peak against the plan at two widths in the native test, so a slope under-charge fails the test (follows #390).

RobTand:
Closes #639.

## What the gap was

**(a) A genuine under-count**, not two quantities measured apart and not a
stale constant.

`selected_anchor_resources` charged 0.1875 GiB an anchor-batch slot on the
GLM-5.3 E2M1 census shapes. The recorded three-width sweep measured
0.4302 GiB a slot: row-0055 layer 20, 192 units at widths 8/16/32,
`torch.cuda.max_memory_allocated` 3.535 / 6.970 / 13.853 GiB, producer
`a4c92094`, sparky, 960/960 encodes byte-identical
(`/mnt/shared/tessera-measurements/tessera486-fused-lut/batch-width-ab-20260915/`).

The old per-slot charge was two terms: `encoder_memo_bytes`, which is derived,
and `compatible_batch_weight_bytes` at sixteen bytes a weight element. Its own
comment says what those sixteen were: eight for the PrismaQuant-side copies,
traced to their call sites, and eight as a stated-but-underivable stand-in for
the producer's working set inside `encode_linear`. Nothing could read the
second eight against any line, and it was low by about 6x.

## What the producer actually holds

`tessera.encode.encode_units` builds one `_encode_unit_steps` generator per
unit and `_drive_in_step` advances them in lock step, so every unit of the
batch holds its working set at the same time. The producer states the scaling
itself, in `encode_units`' docstring: *"Memory is B times one unit's working
set (work, the LDLQ base / ldlq_target / recon / targets in fp32) plus the
joined call's own buffers."*

Read against the pinned producer
(`prismaquant/tessera_runtime/tessera_serving_runtime_pin.json`, commit
`4c384e60`), per unit and per slot:

| Term | Lines | Bytes an element |
|---|---|---|
| `compatible_batch_weight_bytes` | PrismaQuant-side: BF16 device copy in, FP32 `reconstruct_unit`, BF16 cast out | 8 |
| `encoder_working_set_bytes` | `encode.py` `:3377` `scale`, `:3384` the `"scale"` trellis weighting, `:3395` `base`, `:3396` `ldlq_target`, `:3397` `recon`, `:3398` `targets`, `:3401` `residual` | 28 |
| `encoder_code_plane_bytes` | `encode.py` `:2987-2990` `anchors`/`completion_bits`/`codes` int64 `[steps, cols]` and `body_bits` uint8, retained on the returned `EncodedUnit` `:3527-3532` | 25 |

The batching path is byte-identical between the pin and the census producer
`c92826fa4`: the only two diff hunks in `encode.py` are inside `_fit_lut`.

## Before and after

Per anchor-batch slot on the measured shapes (`gate/up [2048, 4096]`,
`down [4096, 2048]`, 2^23 elements each):

| | GiB a slot |
|---|---|
| Plan before | 0.1875 |
| **Measured concurrent peak** | **0.4302** |
| Plan after | 0.5391 |

Exact slopes, from the per-window bytes in `run1/result.json`:
460,933,632 (8 to 16) and 461,917,824 (16 to 32) bytes a slot.

## What the charge does not attribute

The code-plane term is the one **bound** here rather than an allocation:
`steps = rows // arity` (`encode.py:2976`), a plan built from argv does not
know the grid, `arity` is at least 1, and a k-tuple grid halves `steps` faster
than `:2986` widens `body_bits` from uint8 to int32 -- so 25 bytes an element
bounds every grid. It is not what the measured grid pays:
`TESSERA_E2M1_K2_R896` is a `tuple_grid(E2M1, 2)`
(`tessera_formats._build_grid`), so the census pays half.

On the measured shapes and grid, the enumerated terms that are concurrent with
the producer's peak come to about **0.395 GiB a slot** -- 28 + 12.5 at
`arity` 2, plus the BF16 weight in and the memo -- against a measured
**0.4302 GiB**. About four and a half bytes an element of the producer's peak
therefore has **no line named for it**. `encode_units`' docstring ends "plus
the joined call's own buffers", and the `_TCQPlan` state over
`[rows, B*ldl_block]` and the LUT plane's candidate tables in
`_pack_scales_lut`/`_fit_lut` are not enumerated. The shipped 0.5391 GiB
covers the measurement, but it covers that residual through the `arity`-1
bound and through PrismaQuant-side copies the batch path builds only after the
producer releases its working set -- not through a term that means it. The
comment says this at the term. Enumerate the joined call before tightening
either direction.

**Operational consequence, stated rather than hidden:** a streaming row's
demand rises by about 2.8 GiB at width 8. Row-0055's 101 GB becomes about
104 GB against a `box_memory_gb` of 104, and rows that already asked 102 GB
will be declined until they are re-planned. That is the number the
measurement says the box needs.

## Tests

`tests/test_tessera_selected_source.py`:

* `test_anchor_slot_charge_covers_the_encoder_peak_a_measured_sweep_read` is
  new. It drives the real `selected_anchor_resources` at widths 8, 16 and 32
  on the sweep's own shapes, asserts the two slopes agree, asserts the slope
  equals the per-slot sum derived independently in the test, and asserts that
  sum covers the recorded 0.4302 GiB. Mutating the driver's width is what
  moves the assertion; there is no stored plan to edit.
* `test_selected_plan_terms_follow_the_allocations_they_bound` gains the three
  per-slot terms.

Both FAIL before this change (`assert 201359360 == 578846720`) and pass after.

PrismaBuild, `--priority -10`, `--tag dl380g10`,
`/home/rob/venvs/pq-cpu312-tessera-4c384e60`:

* before-state, `tests/test_tessera_selected_source.py`: 2 failed, 21 passed,
  1 skipped.
* after-state, the twelve suites that read the plan or the doc gates
  (`5386feaa4859`): 208 passed, 1 skipped, exit 0, receipt published.
* after-state, every suite that reads `autoscale`, the plan or the dispatcher
  (39 files): see the run recorded in the review comment.

### Not delivered, and why

* **The issue's "assert the guard peak against the plan at two widths in the
  native test" is not delivered.** On the streaming fixture the row's one-time
  first-use cost is 165.6 MB against a `resident_anchors` plan of 10.3 MB, so
  a per-slot delta on that fixture is inside the noise of its first encode
  step and a two-width assertion there would pass on anything. The CPU
  arithmetic test above is what fails before and passes after. A native
  two-width assertion needs a fixture whose per-slot delta exceeds its
  first-use cost; that is a separate piece of work on sparklina.
* **`experiments/glm_native_wire_screen_plan.py` is unchanged.** Its GPU
  subset names a fixed tuple of `resident_anchors` keys and so omits the two
  new device terms. Its own `gpu_peak <= physical_peak` assertion still holds,
  and it is a sealed 2026-09-08 screen at `DERIVED_UNMEASURED`; adding keys
  there would change a completed screen's arithmetic for no result of its own.
* The peak-reset question the triage left open is closed: the per-window
  records show a width-8 pass reading 3.80 GB immediately after a width-32
  pass read 14.87 GB, and `timed-0` and `timed-5` agree to the byte. Peak
  statistics were reset per window.

## Docs

`docs/ARCHITECTURE.md` re-stamped in the same commit, top provenance block,
and the #390 stamp's "two terms are stated as gaps" sentence corrected to one.
No pipeline default, format menu, stage graph, serving lane or ship gate
changes.


RobTand:
## PrismaBuild receipts

All three at `--priority -10`, `--tag dl380g10`, interpreter
`/home/rob/venvs/pq-cpu312-tessera-4c384e60`.

| Run | Scope | Result |
|---|---|---|
| before-state, `tests/test_tessera_selected_source.py` | the fix reverted, the tests kept | **2 failed**, 21 passed, 1 skipped -- `assert 201359360 == 578846720` |
| `5386feaa4859` | the 12 suites that read the plan or the doc gates | 208 passed, 1 skipped, rc 0, receipt published |
| `24a4a6db9d3e` | every suite that reads `autoscale`, the plan or the dispatcher (39 files) | **660 passed, 8 skipped**, rc 0, 748.2 s on dl380g10, receipt published |

Not collected: the 8 skips ran under `-q` without `-rs`, so their reasons are
not in the log. The worker is CPU-only x86, where the GPU-gated branches of
these suites skip by construction.

## The measurement's peak-reset question, closed

The triage comment listed "whether the historical generator resets CUDA peak
statistics between widths" as not established. `run1/result.json`'s per-window
records settle it:

| phase | width | `cuda_max_memory_allocated_bytes` |
|---|---|---|
| warmup | 32 | 14,874,173,952 |
| warmup | 16 | 7,487,675,392 |
| warmup | 8 | 3,800,206,336 |
| timed-0 | 8 | 3,796,019,712 |
| timed-1 | 16 | 7,483,488,768 |
| timed-2 | 32 | 14,874,173,952 |
| timed-3 | 32 | 14,874,173,952 |
| timed-4 | 16 | 7,483,488,768 |
| timed-5 | 8 | 3,796,019,712 |

A width-8 window reads 3.80 GB immediately after a width-32 window read
14.87 GB, so peaks are reset per window; and the repeats agree to the byte.
Slopes: 460,933,632 bytes a slot from 8 to 16, 461,917,824 from 16 to 32.
Intercept at width 8 is 103,301,632 bytes, so essentially the whole peak is
per-slot.

Issue #567, opened 2026-09-13
The fp4 render and the fp4 execution are different objects: decide which side is authoritative before any fp4 number is published


RobTand:
`Part of #559.` Principle 8's rendering-identity-without-execution-identity confound, on the Tessera fp4 lane.

## Measured 2026-09-13

Same harness, same box (sparklina `GPU-b1eceeea`, sm_121, TP1, batch 1, eager, resident), `atol = rtol = 0.015625`. The panel runs two gates: `numerics` compares the GEMM **output**; `qdq_numerics` compares the **activation representation** (`experiments/bench_native_operator.py` ~`:492` computes `represented_native_input(layer, x)`, ~`:505` compares it against the panel's frozen `reference_qdq`).

| cell | phase | `qdq_numerics` max_abs_error | status |
|---|---|---:|---|
| `TESSERA_E4M3_K1_R750` | prefill | **0.0** | passed |
| `TESSERA_E4M3_K1_R750` | decode | **0.0** | passed |
| `TESSERA_E4M3_K1_R1006` | prefill | **0.0** | passed |
| `TESSERA_E2M1_K2_R896` (three MLP units) | both | **0.0** | passed |
| `TESSERA_E2M1_K2_R896` (`o_proj`) | prefill | 0.095703125 | **FAILED** |
| `TESSERA_E2M1_K2_R896` (`v_proj`) | prefill | 0.1435546875 | **FAILED** |
| `TESSERA_E2M1_K2_R896` (`v_proj`) | decode | 0.0078125 | passed |

The **output** gate passed for every fp4 cell at both phases. The kernel's GEMM agrees with its own reference. What diverges is the A-side, and it diverges per unit, not format-wide.

## Why fp8 is the bar and not the tolerance

`represented_native_input` (~`:269-294`) quantizes each family differently, and only fp4's grid is not a function of the activation:

- `TESSERA_BF16` — a cast, no quantizer. Its 0.0 proves nothing.
- `TESSERA_FP8` — `native_fp8_quant(x)` then `codes.float() * scales`. The scale is **derived from x**, so reference and execution compute the same function of the same tensor. **Its exact 0.0 is the bar.**
- `TESSERA_NVFP4` — reads a **static** `layer.trellis_input_global_scale`, then `native_fp4_quant(x, g)` with group-16 blocked scales and `E2M1_VALUES`. Not a function of x.

[...]

PR #704, opened 2026-09-17
Refuse a compressed-tensors card whose served routes disagree with its price

### LINKED ISSUE #631: Principle 14 serve-side leg on the compressed-tensors lane: no route telemetry to compare against
Split from #575. PR #630 closes the serve-side leg on the Tessera lane: the `route.trace` shipcard slot compares the artifact's priced per-contract module histogram with every rank's `tessera.route_trace/1` telemetry.

The compressed-tensors lane has no equivalent. Stock vLLM emits no route telemetry, and `validate_native_export` does not compare priced activation contracts with served routes. This leg stays open there, as recorded in https://github.com/RobTand/prismaquant/issues/575#issuecomment-5674183088.

## Acceptance

- A serve-side source of route truth for the compressed-tensors lane, attested rather than asserted per principle 14. For example: a post-load sweep of the resolved quant method and kernel per module, emitted by a probe that does not vendor vLLM.
- A shipcard slot that refuses on disagreement with the priced `activation_contracts` histogram, with a missing record read as not verified.
- A bite test.

RobTand:
Closes #631. Principle 14's serve-side leg on the **compressed-tensors** lane. #630 closed the same leg on the Tessera lane, on the plugin's own telemetry. Stock vLLM emits none, so this lane's served side is **read off the running engine** instead of received from it.

## The route-truth source, and why it is attestation

`validate_native_export --route-sweep-out <path>` takes the sweep from **the same loaded engine its eager smoke just generated on**, through vLLM's own supported hook `LLM.apply_model` (one result per worker, so this is the TP>1 shape too). Nothing is vendored and nothing is monkeypatched: the values are vLLM's own attributes on vLLM's own objects, and they cross into the gate as a JSON file.

Per module the sweep records:

- the resolved `quant_method` and `scheme` classes, and the scheme object's own attributes (`use_a16`, `group_size`, `is_static_input_scheme`) — the numbers the served activation descriptor is read from;
- the **kernel** the scheme selected (`FlashInferCutlassNvFp4LinearKernel`);
- `isinstance(quant_method, QuantizeMethodBase)` — the predicate vLLM's own post-load sweep filters on;
- `dispatches` — a forward-hook count over the generate the smoke already runs.

**The isinstance trap, checked rather than inherited.** vLLM finalizes weights with `isinstance(quant_method, QuantizeMethodBase)` (`model_loader/utils.process_weights_after_loading`), which is nominal, not structural: a method that is not a subclass is skipped in silence and dies on the first forward. A sweep that filtered on the same predicate would under-report and the gate would compare against a truncated histogram. So this sweep **filters on nothing** and carries the predicate as a field; a `false` is a refusal in the gate, not an absence in the record. Verified on the real serve: the walk found 373 modules and recorded 142 — 112 quantized Linears, 28 `Attention` KV-cache methods, `lm_head` and `model.embed_tokens` — and the gate independently proves the 112 by requiring every one of the artifact's 252 `config_groups` targets to reach a served module.

**Why this is not a producer-side assertion dressed as a measurement.** The priced side is the artifact's own `config.json`, which is *literally the input vLLM's compressed-tensors dispatcher resolved against*. The served side is what the runtime built from it, taken from the runtime. A disagreement means the runtime did not build what our price implies — ignored, fell back, or resolved differently on this platform — which is precisely what no producer-side derivation can establish. The lane spec's `served_activation_quantization.executes` block is the producer leg and remains derived from the exporter's scheme table; this is the other leg, and it is about one artifact on one box.

**The residual assertion, named.** vLLM publishes no machine-readable contract table, so `compressed_route_sweep_gate.SCHEME_ACTIVATION` has to say what `CompressedTensorsW4A4Fp4` means. It is kept as thin as possible: the table says *which observed attributes to read*, and the numbers come from the scheme object. A class the table does not name is **NOT VERIFIED** — never a pass, never a guess — and the refusal prints the class and its module. The roster was read from the pinned image (`vllm/vllm-openai@sha256:61fc8a89...`, vLLM 0.28.0); only `CompressedTensorsW4A4Fp4` has been observed end to end.

## The slot

`lane_specs/compressed_tensors.json` declares a required `route.sweep` gate, verified by `shipcard._verify_route_sweep_record` in `LANE_SLOT_VERIFIERS`. `python -m prismaquant.shipcard_cli fill-route-sweep <shipcard> --sweep <rank0.json> [--sweep ...] --expected-ranks N`:

- **0** the sweeps agree with the price; the slot is filled.
- **1** REFUSED. Each conflict is named: a priced target that reached no served module, a served module the artifact never priced, an ignored module the runtime quantized, a served contract that differs from the priced one, a priced module that dispatched **zero** forwards, a `quantize_method_base: false`, or two ranks with different contract maps.
- **3** NOT VERIFIED. Missing, unreadable, empty, other-schema, non-`enforce_eager`, spec-decoded, `apply_model`-less, or carrying a scheme class the gate has never observed. The slot stays unfilled and the card stays unpublishable.

No operator flag overrides a refusal. `verify` replays the comparison from the carried sweeps and config text and refuses carried config text that is not the artifact's, so a hand-set `passed` closes nothing.

**The exporter now stamps `lane`.** It passed none, so `lane_gate_slots` answered `()` for every native card and the lane's own declarations opened nothing. Harmless while the lane declared only base slots; not harmless now — a slot no card opens is a gate that cannot bite. Cards written before this change carry no lane and keep verifying against the base set they were opened with.

## Granularity

Per module, per rank. Unfused leaves (`q_proj`, `k_proj`, `v_proj`, `gate_proj`, `up_proj`) are reconciled through the `packed_modules_mapping` the sweep **read off the live model class**, not a roster in this repo, so a 252-target checkpoint resolves against 112 served modules exactly.

## What this leg does NOT see

- **The activation representation.** A scheme class is a name, not the quantizer rule the kernel applied, so #567's shape still passes. Same hole the Tessera trace has.
- **Kernel nativeness.** The kernel class is recorded and in the verdict's histogram, because 2026-08-17 was a body riding an older-architecture schedule. It is **not** a refusal condition: vLLM publishes no table of which kernel is native on which target, and a gate that refused on a kernel name would assert exactly what principle 14 forbids.
- **Packed MoE.** No small packed-MoE compressed-tensors artifact was available to sweep, so no MoE method class is in the decoder and an MoE artifact reads **NOT VERIFIED** until one is observed. That is the recorded gap, deliberately, rather than a table written from reading source.
- **Compiled or graph-captured forwards.** Forward hooks do not run under CUDA-graph replay, so only the eager arm writes a sweep and a non-eager one is NOT VERIFIED.

## Measured

`tests/fixtures/compressed_route_sweep_0p6b/sweep_rank0.json` is a **real sweep**, written by this path on sparklina (NVIDIA GB10, sm_121) inside `vllm/vllm-openai@sha256:61fc8a896b0a...` (vLLM 0.28.0) against `/home/rob/dq-runs/fc45-0p6b-nvfp4/exported` (Qwen3-0.6B, NVFP4, 252 targets, 2 ignores). Verdict: 112 modules, all `CompressedTensorsW4A4Fp4(use_a16=False, group_size=16)` on `FlashInferCutlassNvFp4LinearKernel`, 16 dispatches each; `A4:float/g16:dynamic: 112`; every priced target covered. `fill-route-sweep` closed a real card from it, exit 0.

Two operational notes. The vLLM container ran **directly**, per Rob's universal vLLM exemption from PrismaBuild (2026-09-07); the `require_pool` hook has no vLLM carve-out, so the invocation lives in a named script (`vllm_validate_native.sh`) rather than being dressed up as something else. And a second sweep was attempted on `/home/rob/models/Qwen3-0.6B-nvint-smoke`, a four-group artifact that would have observed `CompressedTensorsW8A8Mxfp8` and `CompressedTensorsWNA16` in one run: vLLM refused to load it (`No compressed-tensors compatible scheme was found ... num_bits=3, type='int', format='pack-quantized'`), so no sweep was written and those decoder entries remain unobserved — exercised by unit tests, not by a serve. A failed load closes nothing, which is the intended shape.

## The run prints how to close the slot it opens

`_write_shipcard` now opens `route.sweep`, and `run-pipeline.sh` ended by printing `shipcard_cli show`, which names the OPEN slots but not what closes them. A slot a run opens and prints no way to close is how an operator learns to reach for `--force-unverified`, which `shipcard.py`'s own docstring names as the failure mode. The native arm now also prints the lane's gate report (`python -m prismaquant.lane_shipcard gates --lane compressed-tensors`), read from the lane spec rather than respelled in the shell, so a lane that adds a gate prints its recipe with no edit to the driver. `test_the_run_prints_how_to_close_the_slot_it_opens` pins it.

## Validation (PrismaBuild, x86 CPU lane)

`pbtest.py --python /home/rob/venvs/pq-cpu312-tessera-4c384e60/bin/python --tag dl380g10 --priority -10`

- **This branch.** 19 test files — the two new ones plus every existing test touching the shipcard, lane specs, publication, the native-export validator, the compressed lane's executes derivation, the Tessera route gates and the two docs tests. **8/8 shards green, 525 passed, 0 failed, 0 skipped.** Keys: `06b9fb82232e` `5f8f43c5b1db` `22f5f6b1b11a` `d6d9204a42d7` `95ca1dc7d1e3` `505d3219d09d` `40295a72ea77` `0c5180e966b9`.
- **Bite demo, byte-identical test file on both sides** (`tests/test_bite_631_route_sweep.py`, md5 `e36dd10f234c378d067458dba060e5ac`; imports only APIs that exist before this change). On `origin/main` `05f0b40bbb` with the fixture copied in: **3 failed, 0 passed** (key `94825d15d73b`) — the lane owes no such slot, a fresh card does not owe it, and `fill-route-sweep` exits 2 because the subcommand does not exist. On this branch: **3 passed**, inside shard `5f8f43c5b1db`.
- **Driver mutation.** `compressed_route_sweep_gate._descriptor` was made contract-blind (it kept `quantized` and dropped `num_bits`, `type`, `group_size`, `dynamic`). `tests/test_compressed_route_sweep_gate.py` then gives **3 failed, 33 passed** (key `aa2539a20df8`), including `test_refuses_when_the_price_itself_moved_under_the_observation` — with a contract-blind descriptor, a price that moved from group 16 to group 32 under an unchanged serve passes, and the test catches it. The mutation was made in a throwaway worktree and is not in this branch.
- **Discarded first submission.** The first `pbtest` named `pq-cpu312-tessera-7dbbacbd`, which the dev-pin resolver refused before pytest (main pins `4c384e60`). It is not counted.
- **One existing assertion changed on purpose.** `test_shipcard.py::test_export_writes_a_card_with_build_facts_and_empty_slots` asserted the exporter's card opens exactly `REQUIRED_SLOTS`. That encoded "the native lane declares no slot of its own", which just stopped being true. It now asserts the union *and* the lane's declared slot tuple, so dropping either the lane stamp or the declaration still fails it.

## Docs (principle 13, same commit)

`docs/ARCHITECTURE.md` re-stamped: new provenance entry, §7.1's stage table gains the sweep row, §7.1's slot documentation gains `route.sweep`, §9.1 says this lane now uses its in-process load as its window onto what the runtime resolved. `CLAUDE.md` principle 14 no longer says this leg is open on the compressed-tensors lane.


RobTand:
Follow-up filed: #706 — `route.sweep` reads NOT VERIFIED on every packed-MoE artifact on this lane, because no MoE method class has been observed on a sweep. Refusing rather than passing is the intended default, but it is closable with one sweep of any packed-MoE compressed-tensors artifact; none was on disk with weights. Issue names what closes it and what was checked and ruled out.

PR #728, opened 2026-09-18
Run stage: compare the prepared render proof with the installed source, and carry the fix to the sealed GLM-5.3-Flash prepare

RobTand:
Closes #727.

Integration branch = campaign branch `flash/run-render-proof-meta-skeleton-20260918` (cut from the prepare commit `208d340d27`, so the sealed package reconstructs byte-exact) + `origin/main` (one stamp conflict in `docs/ARCHITECTURE.md`, both stamps kept). The campaign run is submitted from the campaign branch, not from this merge; the receipt binds package bytes, and `prismaquant/` is identical on both branches except for main's own unrelated changes.

## The defect

`compute_aura_cost_streamed` compared every prepared render proof's `dtype` and `logical_bytes` with `linears[name].weight` up front, before any layer was installed. On the streamed path that parameter is the meta skeleton, which `_from_config` builds in float32 for a wrapper config whose `dtype` is None (GLM-5.3-Flash; measured on the x86 lane, PB `cfa6adf0132c`, transformers 5.16.1 as in the campaign image). Prepare verified every cell against the installed bf16 tensor, so the run refused its own prepare on the first pair (PB `4bafaaeb…`, after a 50 min intake).

## The fix (`prismaquant/aura_cost.py`)

The dtype and byte identity of a source is a fact about the installed tensor: the loader decides it (`_read_layer_to_device` casts to the recorded model dtype, fp8 dequantization hard-codes bf16), so it is not a safetensors-header fact either. The comparison moves to `_require_installed_render_sources(layer)`, called in the reverse loop right after `_refresh_packed_layer_views(layer)`, against the installed tensor; it also refuses a parameter that is still meta. The up-front loop keeps coverage, key, digest-format, shape and verified-cell checks and now only requires the dtype and byte fields to be well-typed. Skeleton dtype and the prepare path are unchanged.

Red first: `tests/test_joint_retained_streamed.py` gained a `skeleton` hook (meta parameter of a chosen dtype until `install`). RED on `208d340d27`: PB `e4c59e50f0fb` (2 failed: "prepared render tensor proof differs for model.layers.0.proj@FP8_E4M3"). GREEN with the fix: `c34885c2af9c` / `0a63bf1de9ad` (9 passed).

## The transition (`prismaquant/joint_aura_run_transition.py`, `joint_aura_transitions.py`)

[...]

Issue #707, opened 2026-09-18
PrismaBuild residency-map reader for the stage tier


RobTand:
PrismaBuild's tiered caching (RobTand/prismabuild#583) stages the files a
consumer's data manifest names onto an SSD stage tier and writes a residency
map. The launcher injects the map's path into the consumer container as
`PRISMABUILD_RESIDENCY_MAP`. PrismaQuant owns the consumer half: without a
reader, the stage is, in the design's words, "a copy nobody reads, and the
before/after will say exactly that."

Rob has gated the GLM-5.3-Flash campaign's `run` stage on this landing.

## What this issue covers

1. **One resolver.** `prismaquant/residency_map.py`: when
   `PRISMABUILD_RESIDENCY_MAP` is set and names a valid map, a declared
   manifest path that has an entry is read from the entry's `stage_path`
   after a per-entry identity check. A mismatch or a missing stage file
   refuses that entry, falls back to the pool path and is recorded. Without
   the variable the resolver is inert and the bytes are identical to today.
2. **Closed-loop accounting** in the joint run's `results.json`:
   `residency: {map_path, map_sha256, entries, hits, misses, fallbacks,
   bytes_from_stage, bytes_from_pool}`.
3. **`--residency stage`** on `tools/dispatch_tessera_campaign.py`'s GPU
   submissions, passed through to `pbrun.py` and recorded in the demand
   record. `verify_joint_submission_demand` is unchanged.

## Schema

PrismaBuild owns it. Built to the draft:

```json
{"schema": "prismaquant.prismabuild.residency_map.v1",
 "tier_id": "...", "stage_root": "/stage/prewarm",
 "leads": ["<mover key>"],
 "entries": {"<manifest path>": {"stage_path": "...", "bytes": 0,
                                 "sha256": "<64 hex>"}}}
```

Entries are keyed by the data manifest's own path strings
(`prismaquant.prismabuild.data_manifest.v2`, absolute and normalized under
`mount_prefix`). Entries exist only for complete, verified copies. The map is
atomically replaced as movers progress, so the reader re-reads it on identity
change rather than loading it once.

## Not covered

- The measurement. `/stage` is not exported to the Sparks yet.
- Byte ranges. The draft is whole-file; a rev-2-shaped map (`entries` as a
  list of `{index, path, offset, bytes, resident}`) is refused whole with a
  reason naming the shape.

Issue #632, opened 2026-09-15
Tessera rungs dropped at TP2 because max_world_size is 1


RobTand:
## Problem

When the allocator runs at tensor-parallel size 2 in the attested menu, it drops every Tessera rung. The cause is the reviewed Tessera pin (`7dbbacbd`, contract v24): its closed-world `tensor_parallel` block lists `TESSERA_E2M1_K2`, `TESSERA_E4M3_K1` and `TESSERA_BF16_K1` at `max_world_size: 1`.

The refusal travels through three places:

- `prismaquant/tessera_menu.py:499-507` (`tessera_tp_world_attested`) refuses `tp > declared` with `tp2_unattested`.
- `prismaquant/tessera_menu.py:1233-1236` (`tessera_tp_legal`) calls that check whenever `require_attested_world` is set.
- Both callers set it in attested mode:
  - `prismaquant/allocator_candidates.py:768-772`, the per-Linear applicability check;
  - `prismaquant/tessera_menu.py:1471-1475`, the `expand_tessera_menu` build.

As a result, a GLM TP2 allocation gets no E2M1_K2, E4M3_K1 or BF16_K1 rungs.

## What changed upstream

Tessera #517 (master `4c384e6049`, contract v29) raises `max_world_size` to 2 for all three units. Each raise is backed by a `world_size_receipt` entry in a new `world_size_receipts` table, and the contract also adds `kv_head_replication`.

## Fix

- Repin PrismaQuant to `4c384e6049` using the procedure in `prismaquant/tessera_runtime/README.md`.
- Confirm that the reader and preflight accept the v29 `tensor_parallel` keys.
- Keep refusing any `tp` above the attested world.
- Provision matching PrismaBuild venvs for the new pin.

The repin also accepts contract versions v25 through v28. Those change admission and need their own review in the PR:

- v25: the `activation_quantizers` table.
- v26: the E2M1_K2 `row` axis becomes sharded.
- v28: two routed-MoE E2M1_K2 q896 cells whose smoke is `not_recorded`.

PR #736, opened 2026-09-18
Shard reads go through the residency resolver, ranges included

### LINKED ISSUE #732: Residency map covers wire reads only: layer_streaming's safetensors shard reads bypass the resolver and stay pool-fed
## What happened

Run `ad8803aa` (GLM-5.3-Flash, `--residency stage`) has been in boundary capture since 16:24:17Z with the GPU at 13 W (9% of envelope) and four reader threads in D state on page faults. The files it has mapped are all under `/mnt/shared/models/GLM-5.3-Flash-BF16/` and `/mnt/shared/tessera-measurements/...` — the pool — while the composed residency map names 55 of those model shards on `/stage/prewarm`, staged for exactly this consumer.

`residency_resolver()` is consulted in one place, `tessera_joint_aura.py:102` (`resolver.staged_read(wire, ...)`), i.e. for render wires. `layer_streaming` / `streaming_model` open the BF16 shards directly, so every phase that reads model weights (boundary capture, any layer install) is pool-fed regardless of what the movers staged, and the bytes the plan moved for those entries are moved for nothing. Zero `[residency] fallback` lines were printed, which is consistent: the resolver was never asked.

## What is wanted

Route the shard opens through the same resolver (`staged_read` / stage path substitution at open, with the size + sha check the wire path already does, and the same `record_stage_read` / `record_pool_read` accounting), so the residency report states what fraction of *all* bytes came off the stage. Until then the report overstates stage coverage: it counts only wire bytes.

Measured 16:34Z: sparky RDMA rx 0.8–1.0 GB/s from dl380g10, pool read 266 MB/s at the server, GPU 13 W.

RobTand:
Closes #732.

PrismaBuild stages the byte ranges a run's data manifest declares and hands the run a residency map (`PRISMABUILD_RESIDENCY_MAP`), but no BF16 source-shard read ever asked for it. `layer_streaming._source_safe_open` called `safetensors.safe_open` on the declared pool path, and two bare `safe_open` calls sat outside that seam. On the live GLM-5.3-Flash run every mapped model file in `/proc/<pid>/maps` was a pool path. This is blockers 3 and 4 of PQ #732; #733 fixed 1 and 2.

**A path rewrite could not have fixed it.** PrismaBuild's mover writes each staged range as a file of its own starting at byte 0 of the range (`tools/fleet/stage_move.py:80-96`, `os.lseek` at `:199`), so 34 of the run's 55 staged shard entries are byte ranges and `ResidencyResolver.staged_read` refuses every one by contract (`residency_map.py:366-371`). Byte 0 of a range file is tensor payload, not safetensors' header-length prefix.

## What this adds

**Resolver.** `staged_range(declared, start, end, *, declared_size=None)` returns the single entry whose `[offset, offset+bytes)` covers `[start, end)`, under the same pre-open fences `staged_read` applies: the entry has to fit inside the declared file, the staged copy has to be a regular file of exactly the entry's length, and a refusal falls back and is recorded through `_record_fallback`. It reads a per-declared-path interval index built lazily and rebuilt whenever `_adopt` or `_forget` installs a new `_entries`, in both the manifest's spelling and the realpath one. `stages(declared)` is the cheap question a reader asks before it wraps anything. **`staged_read`'s contract is untouched** — its two whole-file consumers depend on it, and a test asserts it still refuses a byte range.

**Reader.** `prismaquant/residency_shard_reader.py`. `staged_shard_opener(declared, pool_open)` returns `pool_open` itself when the resolver is inert or the map does not name the file; otherwise it returns a closure building `StagedShardReader`, a drop-in for `safe_open` at the seam including through `source_authentication.safe_open(opener, path, **kwargs)`, which hands the opener `/proc/self/fd/<fd>` rather than the declared path (the declared path travels in the closure). Behaviour:

- the header is read from the file it is handed, once, lazily on the first payload read, so a `keys()`/`metadata()`/`get_slice()`-only open pays nothing;
- each tensor's payload comes from the staged range covering its whole `[data_start+begin, data_start+end)` span, read with a `preadv` loop into a writable buffer and rebuilt as `frombuffer(...).view(dtype).reshape(shape)`, then `.to(device)` when `device=` was passed, so the direct-CUDA path keeps its semantics;
- a span straddling two ranges, covered by none, or of length zero reads the pool — never partial bytes, never a concatenation nothing fenced;
- `keys()`, `metadata()` and `get_slice()` delegate to the pool handle;
- each staged range is opened once per reader and its stat signature is re-checked after every read; a failure drops that range and reads the pool.

The dtype table is `safetensors.torch._TYPES`, read rather than re-declared; without it nothing is served from the stage.

**The two bare calls** (`layer_streaming.py` per-expert pack, `streaming_model._estimate_layer_cache_bytes`) now go through `_source_safe_open`.

**Accounting.** `residency_report()` gains `range_hits` and `range_misses`. `range_hits` is counted inside `hits`, so `hits` stays "reads the stage served" and `range_hits` says how many came from a range. A span no entry covers is a silent miss, not a printed fallback: a half-staged shard is the ordinary mid-flight state and would otherwise print a line per tensor.

## What it admits on, and what it costs — stated plainly

The fences here are **structural, not a digest**. A shard read takes one tensor's span out of a multi-gigabyte staged range whose only published `sha256` covers the whole range, so unlike `production_weight_cache._read_file_tensor` and `tessera_joint_aura._read_wire_bytes` nothing re-hashes what it reads. Recorded in `docs/ARCHITECTURE.md` §12 D43 rather than left implicit:

1. Under `source_authentication` the payload handed to the caller may come from the stage while the `prismaquant.selected_source_authentication.v1` receipt's SHA-256 is of the declared pool file. Its structured fields stay literally true and no gate reads it, but its `authentication` prose no longer describes where the payload bytes came from. The honest fix is an additive per-file staged-read count on that receipt; it is not in this PR.
2. Hashing each staged range once on first touch would close that gap at the price of a second full read of every staged extent before its first tensor can be served — a hot-path change owing a principle-15 before/after profile.
3. In *fresh* authentication mode the first payload read hashes the whole pool file anyway, so a redirect there adds a read. The joint run adopts `prepare/source-identity.json` (`tessera_joint_aura._prepare_source_owner:1676-1678`) and runs in the adopted mode where no pool payload read is issued, but nothing in the reader knows which mode it is in.
4. **Page release (D43(c)).** `_advise_consumed_safetensors_pages` advises the *declared* file, whose payload pages a staged read never faults in, while the staged range's pages land in page cache through `preadv` and are advised by nothing — the same gap D43(c) already recorded for activation captures, now applying to shards.
5. **The read's memory shape changed (D43(c2)), unmeasured.** `safe_open` materializes from the shard's mmap; this reader `preadv`s into an anonymous buffer and then copies to the device, so a staged tensor costs one extra copy plus a transient anonymous allocation the size of the in-flight tensor, per reader thread (`PRISMAQUANT_LAYER_READ_THREADS`, 4 on the live run). On 121 GB of unified memory beside a 13.8 GB layer it fits, but it surfaces as a peak `MemAvailable` dip rather than a wall-clock change, and nothing here measures it.

## Tests — all through PrismaBuild on sparklina, `--priority -10`

`tests/test_residency_shard_reader.py` builds a real safetensors shard (F32, BF16, F16, F8_E4M3, I8, U8, BOOL and a zero-element tensor), stages ranges exactly as the mover writes them, and compares every served tensor bit for bit against `safe_open`'s.

**RED, twice.**

| key | host | rc | result |
|---|---|---|---|
| `9bdd161b500d` | sparklina | 1 | against origin/main's source: `1 error` — `ImportError: cannot import name 'residency_shard_reader'` |
| `be297058600b` | sparklina | 1 | reader and resolver present, seam **unrouted** (`layer_streaming`/`streaming_model` at origin/main): **8 failed**, 10 passed — every test that asserts a staged read through the seam |

**GREEN** (new file plus the neighbours), 204 passed / 4 skipped / 0 failed:

| key | host | rc | shard | result |
|---|---|---|---|---|
| `778e22e0a2b0` | sparklina | 0 | `test_residency_shard_reader`, `test_streaming_source_pages`, `test_joint_retained_streamed` | 51 passed |
| `db567700c2eb` | sparklina | 0 | `test_prismabuild_residency_map`, `test_streaming_production_cache`, `test_streaming_text_only_wrapper_config` | 51 passed, 2 skipped |
| `7e631428c2ae` | sparklina | 0 | `test_layer_streaming`, `test_glm_campaign_streaming`, `test_streaming_buffer_precision` | 12 passed, 1 skipped |
| `f4bb494b988d` | sparklina | 0 | `test_layer_streaming_fp8`, `test_selected_source_authentication`, `test_collector_source_release` | 39 passed, 1 skipped |
| `5d56e08c0d5d` | sparklina | 0 | `test_layer_streaming_mxfp4`, `test_docs_staleness` | 26 passed |
| `e58633d47507` | sparklina | 0 | `test_dsv4_layer_streaming_rename`, `test_architecture_doc` | 25 passed |

The second commit's D43 edit re-ran the docs gates: `266fba4f2022` (`test_docs_staleness`, 6 passed) and `a1d691f655c2` (`test_architecture_doc`, 13 passed), both sparklina, rc 0.

No GPU work, no measurement claim: this PR changes where bytes are read from, and the before/after read-rate delta is D43's, on a real joint pass with a real map.

`docs/ARCHITECTURE.md` is re-stamped in the same commit as the code for the resolver's new public methods, and its D43 debt row is rewritten.

https://claude.ai/code/session_015ZUrgvkeGTwzCCBhYKZDcK


RobTand:
# End-to-end proof: a shard read is served off the PrismaBuild stage tier

PB action `336350a2aaa4760466074052fe4c1eb25bacc583c67c48733619b346b37470c4`,
worker **sparklina**, 21 s, rc 0, CAS receipt
`2bd587ce01886ac7781c8ac8475de990f1b74a7e4fd6d51ae1172ddb639007de`.
Branch `flash/732-resolver-routed-shard-reads-20260918` (PQ #736).
Harness: `stage_fed_demonstration.py` in that worktree. Read-only; wrote nothing.

Not a unit test. It uses the **live** residency map composed for the running
GLM-5.3-Flash joint action `ad8803aa` (96534 entries), the real staged ranges on
`/stage/prewarm`, and the real seam `layer_streaming._source_safe_open`. It
measures bytes per NFS mount from `/proc/self/mountstats`, so a redirect that
silently fell back to the pool cannot pass.

## Target

`/mnt/shared/models/GLM-5.3-Flash-BF16/model-00001-of-00120.safetensors`, named
by four map ranges:

| range | size | on stage |
|---|---|---|
| `[0 .. 2538602496)` | 2.36 GiB | yes |
| `[2537553920 .. 3117416448)` | 0.54 GiB | **no** |
| `[3116367872 .. 3696230400)` | 0.54 GiB | **no** |
| `[3695181824 .. 5358221856)` | 1.55 GiB | yes |

Half-staged, and the staged extents are `.pbrange` files whose byte 0 is tensor
payload rather than the safetensors header prefix. That is exactly what the
whole-file `staged_read` contract refuses, so this exercises #736's range path.

## Result

Six largest covered tensors, 2484.0 MiB of payload:

| mount | client bytes | from the server |
|---|---|---|
| `/stage/prewarm` | 2484.0 MiB | 2484.9 MiB |
| `/mnt/shared` | 55.6 MiB | **0.0 MiB** |

wall 3.71 s.

Resolver report: `bytes_from_stage` 2,604,662,784 · `bytes_from_pool` 0 ·
`range_hits` 6 · `hits` 6 · `misses` 0 · `range_misses` 0 · `fallback_count` 0 ·
`tier_id` `prismabuild-stage:dl380g10` · map sha256
`76aedd40ee2aed566c0575edb24e05028138b1ba7360446cc143db92e6190638`.

All six tensors compared byte-for-byte against `safetensors.safe_open` on the
pool file: **6 of 6 identical**.

## What this proves, and what it does not

Proves: the resolver, the range reader and the seam serve real staged bytes,
correctly, and the bytes demonstrably come off the stage mount rather than the
pool.

Does **not** prove: the container path. This ran outside a container, so PQ #733's
map forwarding and `/stage/prewarm` bind are still unit-tested only. Nor does it
claim any speedup; it establishes source and correctness, not rate.

## Incidental findings

1. **Two of the four ranges for this shard are not on the stage.** The map names
   ranges whose movers never ran, so shards are half-staged. This is PB #632
   (19 of 46 movers staged, 1 of 46 egresses ran) seen from the read side.
2. **Consecutive ranges overlap** by the manifest's 1 MiB rounding
   (`[0..2538602496)` and `[2537553920..3117416448)`), which `staged_range`'s
   single-covering-entry rule handles without ambiguity.
3. A first attempt was refused with `no data manifest digest is bound to this
   process`. That is the guard working: a pass must declare which read set it is
   entitled to. The joint run binds it at `tessera_joint_aura.py:1894` from
   `--data-manifest-sha256`; the harness now does the same.

RobTand:
## CI caught three regressions this PR introduced; fixed in `b3ab57ae4f`

The PB receipts in the report were per-file and did not cover the whole suite. The full CI run did, and separated cleanly against main's own baseline:

**Main already fails 3** at `e7c1c71f` (the #733 merge), unrelated to this PR:
`test_joint_cost_streamed_schedule.py` x2 (`read() takes 1 positional argument but 2 were given`) and `test_campaign_image_content.py` x1 (`resolves to no commit`). Filed separately.

**This PR added exactly 3**, all one cause:
```
AttributeError: module 'prismaquant.streaming_model' has no attribute 'safe_open'
```
in `test_glm_original_graph_profile.py` and `test_selected_source_authentication_ab.py` x2.

### Why, and the fix

Routing `_estimate_layer_cache_bytes` through `_source_safe_open` removed `safe_open` from `streaming_model`'s namespace. Two experiment recorders patch that name as a **separate interception point** from `layer_streaming.safe_open`: `glm_original_graph_source.reader_binding` binds it for metadata-only reads (`metadata_only=True`), and `selected_source_authentication_ab` observes all three modules.

Merely restoring the import would have been worse than the failure: the patch target would exist while the call no longer went through it, so the metadata interception would have gone **silently dead**.

The routing also bought nothing. `_estimate_layer_cache_bytes` reads only `get_slice().get_shape()` and `.get_dtype()` — safetensors header fields, not one payload byte — and `StagedShardReader.get_slice` delegates to the pool handle anyway. So `streaming_model.py` is reverted to main and the seam that matters, `layer_streaming._source_safe_open`, is untouched and still covers every payload read including the per-expert pack at `:1338`.

Consequence for the closed source transition: the three `streaming_model.py` hunks in `source_rewrites_732.py` are no longer needed.

Revalidated on sparklina through PrismaBuild, 4/4 shards green: `test_glm_original_graph_profile` 25 passed, `test_selected_source_authentication_ab` 14, `test_residency_shard_reader` 18, `test_layer_streaming` 5.

Also adds `tools/stage_fed_demonstration.py`, the harness behind the end-to-end receipt above, so the proof is rerunnable rather than a one-off.

RobTand:
CI red here is **entirely inherited from `main`**, not from this PR. The `tests (py3.12, cpu)` job on run 35379993749 fails on exactly three tests, and they are the same three that fail on `main` at `e7c1c71f96`:

```
FAILED tests/test_joint_cost_streamed_schedule.py::test_real_streamed_cost_reports_before_reads_and_after_durable_windows
FAILED tests/test_joint_cost_streamed_schedule.py::test_partial_and_complete_resume_keep_original_windows
FAILED tests/test_campaign_image_content.py::test_resolved_image_is_executed_after_content_check
```

Nothing this PR touches appears in that set — its own three regressions (the `streaming_model.safe_open` `AttributeError`s) were removed by reverting that file in `b3ab57ae4f`, and PrismaBuild ran 4/4 shards green on the branch.

**#742 fixes all three** (issue #739): both are test doubles that drifted behind real production changes — `_load_file_tensor` gained a load-bearing `key` argument in #707, and the container launcher gained a `git rev-parse` in #728 that the test's catch-all `subprocess.run` fake was answering with Docker inspection JSON.

**Merge order: #742 first**, then update this branch; the job goes green without another change here.

Issue #753, opened 2026-09-18
Fleet x86 test env pq-cpu312 carries tessera-quant as a local-directory install, so the pbtest pin gate refuses every shard


RobTand:
What breaks: any `pbtest.py --python /home/rob/venvs/pq-cpu312/bin/python` run refuses before pytest with `tools/resolve_tessera_dev_pin.py: expected commit=4c384e6049dca3eeaf503bb2c9cd1cd2778978d1; distribution=tessera-quant installed commit=<unknown>; require a non-editable Git install at the reviewed commit (local-directory installs do not record a Git commit)`. Reproduced 2026-09-18 from a pq-perf worktree (PB refusals on dl380g10, `--priority -10`).

Filer could not fix it here and chose to file: re-provisioning the shared fleet interpreter (tools/provision_tessera_pin.py) while the pool is non-idle risks other users' runs; per the agent skill that needs an idle fleet or a separately provisioned env. Workaround used: an admitted `pbrun --here` on sparky with /home/rob/venvs/pq-cu130-tessera-4c384e60 (196 passed, 4 skipped, action d7f88ec3514e).

Severity: P2 (provenance/observability -- the gate is working as designed; the env is behind it). Tripped over while verifying flash/725-737-738-aura-perf-20260918; filed same session.

RobTand:
Closed as superseded during the 2026-09-19 unactionable sweep: the PQ shard path now runs on git-installed pin venvs (pq-cpu3essera-79ddd4c60, and pq-cu130-tessera-cc739a55 for the #766 pin) which record their commits and pass the pbtest dev-pin gate — verified in every PB shard tonight. Twin of PrismaBuild #658, closed on the same evidence (561e7cc648 recorded the refusal and the idle-window fix). — GLM-5.3 coordinator

RobTand:
Correction to the closure comment above: the venv names are **pq-cu130-tessera-79ddd4c60** and **pq-cu130-tessera-cc739a55** (the string was garbled in transcription). The evidence is unchanged.

Issue #738, opened 2026-09-18
Joint AURA boundary capture cannot fan out: one action pins one box for its whole life while the second GB10 idles, though the capture is 512 independent calibration partitions


RobTand:
## The observation

The GLM-5.3-Flash joint run has occupied sparky alone for its entire life while sparklina sat at 0% GPU and ~1.5 load. That is not a scheduling accident: the run is a **single PrismaBuild action** that claims one box and holds it, and nothing in the joint path can be split.

```
$ grep -nE "def .*(shard|partition|fanout|campaign|parallel)" prismaquant/tessera_joint_aura.py
(no matches)
$ grep -nE "partition|sample_parallel|shard_index|world" prismaquant/cost_streaming.py
(no matches)
```

Neither `tessera_joint_aura.py` nor `cost_streaming.py` contains a partition, shard, rank or fanout code path. The boundary capture runs one box, one process, one layer at a time.

## Why this is the wrong shape

The capture's own banner names the axis that should fan out:

```
[aura 16:24:17] boundary capture: calib (512, 512) in 512 partition(s) across 45 layers
```

512 calibration partitions. Per layer, each partition is an independent forward over a slice of calibration data whose outputs are written as independent boundary entries (`boundary-<layer>-<partition>-at-<n>.pt`, one file each, 16.8 MB). They do not communicate. This is the data-parallel axis, and it is already materialised as separate files.

The fleet policy is explicit that work must reach PrismaBuild as tiny independently verifiable and retryable quanta through its fanout and campaign interfaces, with PrismaBuild owning sharding, placement and balancing. A 2.4-hour single-box action that pins one GB10 and leaves an identical idle one next to it is the opposite of that, and it also makes the residency window unmanageable: one action holds a 3.60 TB plan's worth of ranges for its whole life (PB #632).

## What exists and is not wired

The repository already carries a sample-parallel contract:

- `prismaquant/sample_parallel_probe_contract.py`
- `prismaquant/sample_parallel_probe_merge.py`
- `prismaquant/cluster_campaign_contract.py`

The probe can fan out over samples and merge. The joint AURA boundary capture cannot. Note the recorded debt that the sample-parallel probe was coupled to a *campaign census* inside the retired Gridbook lane (D34), so the first question is whether that contract is reachable at all today or whether its driver went with the lane.

## What to decide

[...]

Issue #737, opened 2026-09-18
Source prefetch is pinned to 1 worker / 2 cache slots in the sealed plan while 97.4 GB of budget and 7 slots are free: 31% of the run's main thread waits on settle_prefetched_layers


RobTand:
## The observation

The live GLM-5.3-Flash run `ad8803aa` spends **31% of its main thread** (28 of 90 py-spy samples, 2026-09-18 17:39Z) blocked in:

```
wait (threading.py:355)
result (concurrent/futures/_base.py:451)
settle_prefetched_layers (prismaquant/streaming_model.py:992)
```

That is the forward pass waiting for the next layer's weights, with the GPU idle.

## Why it cannot go faster

The container's own log:

```
[cost-streaming] layer cache budget=97.4 GB (free=99.4 GB, headroom=2.0 GB, dynamic_reserve=2.0 GB, explicit)
[cost-streaming] prefetch auto: workers=1 (explicit, capped by max_cache_slots), cache_slots=7,
                 max_cache_slots=2, memory_slots=7, est_layer=13.8 GB, min_avail=2.0 GB (explicit)
```

The machine offers 97.4 GB of cache budget and 7 memory slots. The plan caps it to two slots and one reader:

```json
"source_prefetch": {
  "prefetch_workers": 1,
  "max_cache_slots": 2,
  "prefetch_lookahead": 1,
  "prefetch_min_available_gb": 2.0,
  "require_prefetched_residency": true
}
```

With `max_cache_slots: 2` there is exactly one layer of read-ahead, and `prefetch_workers` is then capped to 1 by that. One thread pulls 13.8 GB per layer while nothing else overlaps it. These values are in the **sealed** plan JSON (`_source_prefetch`, `tessera_joint_aura.py:1536-1554`), so no environment override reaches them and the running action cannot be changed.

## What to decide

1. Whether the depth should be derived from the measured budget rather than pinned: with 97.4 GB of budget and a 13.8 GB layer, ~7 slots and 2-4 readers fit, which is what `cache_slots=7 / memory_slots=7` already computed before the cap overrode it.
2. If it stays pinned, the plan for the next submission needs different numbers; that is a plan edit, not a code change, and it is Rob's call.

Deriving the depth from the free budget is the principle-2 shape (no constant where the objective gives a number); pinning it at 2 is a constant nobody has re-measured since the plan was sealed.

## Related

- #735 (the boundary-entry hashing that accounts for another 22% of the same thread).
[...]

Issue #725, opened 2026-09-18
joint prepare: 7.4 h for 198k cells — profile budget and ranked fixes (hashing chunk size, double parse, serial identity binding)


RobTand:
## 1. Time budget

Wall for phase `prepare` (results.json `phases`): **26,677.3 s (7.41 h)**, 36,423 units, 197,990 measured cells. cProfile thread total 33,648.7 s — ~26% of thread time overlapped (worker threads). Aggregate read rate from io_after − io_before: **233 MB/s** over 7.41 h.

| Bucket | From pstats | cum s | ~wall share |
|---|---|---|---|
| Activation/window archive loads (readinto 4,540 + torch serialization `_load` 4,105 + unpickler 2,870 + torch.tensor 947) | serialization.py:1976, perturbed_x_cache.py:579 | ~11–12k | ~40% |
| Hashing (openssl_sha256 tottime 3,943 + HASH.update 2,589) | builtins | 6,533 | ~24% |
| container.parse 396,008 calls | container.py:279 | 2,698 | ~10% |
| Identity binding (bind_checkpoint_unit_identity 36,423 × 71 ms) | tessera_campaign.py:2261 | 2,595 | ~10% |
| Wire-read fence + digest (_read_verified_wire_blob) | tessera_joint_aura.py:86 | 1,688 | ~6% |
| Render-tensor digests (_cb_cache_tensor_identity 234,413 calls) | production_weight_cache.py:3800 | 2,088 | ~8% |
| Pure waiting (settle_prefetched_layers 45 × 49 s) | streaming_model.py:972 | 2,217 | ~8% (overlaps loads) |

Buckets are not additive: loads/hashing run on worker pools (61,267 of verify_cached_unit's 197,990 calls arrive from futures thread.run; `file_hash_workers: 4`, `source_prefetch_workers: 1`, `prefetch_lookahead: 1`, `max_cache_slots: 2` in results.json) while binding/qualification runs on the main thread. cProfile cannot see libc/page-cache kernel time, GIL-blocked background time, or GPU-side dequant beyond its Python frame (`reconstruct_unit` 1,644 s cum).

[...]

PR #756, opened 2026-09-19
Joint-AURA throughput: zero-copy digest, fenced pair bind, prefetch derivation, partition quanta (#725 #737 #738)

RobTand:
Shared-layer receipt-identical: chunked 8MiB memoryview digest (same hex); verify_anchor_render binds reader fenced (blob,digest) dropping 3rd hash; recommend_source_prefetch() derivation + underuse note (no admission change); plan_boundary_partition_ranges()+verifier naming 512 quanta (schedules nothing; PB owns execution).

Profile-before-change confirmed triple-hash/double-parse cross boundary (PQ-side only removable), producer duplicate digest (0.105s→0.068s/33MiB), sealed prefetch cap, no fanout path. No defaults/stages/formats/pins; digests byte-identical so no KL/bpp re-baseline.

Verification: pbrun --here sparky admitted d7f88ec3514e CAS: 196 passed 4 skipped + doc guards 19 passed. Filed #753 (x86 pq-cpu312 Tessera provenance). Tessera single-parse/hash + #737 reseal + #738 campaign-row design reported as gaps. Report /tmp/opencode/pq-perf.md.

For GLM: accept/reject.


RobTand:
Closing per Rob's direction: this conflicts with #754 (the resumable, parallel head walk — in flight now), which restructures the same walk/read path in tessera_joint_aura.py that this PR's #725 pair-binding and #737 prefetch derivation also rework. #754 supersedes.

Salvage guidance for the rebase after #754 lands: (a) the zero-copy host-tensor digest (#725's _cb_cache_tensor_identity change) is function-independent of the walk and looked sound — rebase it as its own PR with the bit-exactness evidence; (b) verify_anchor_render's fenced pair bind similarly; (c) #737's prefetch derivation and #738's partition quanta must be reconciled against #754's new walk structure before re-PR — the derivation input set changes. Your validation receipts (196 passed + doc guards on sparky) are noted and were not the issue; the conflict is structural.

PR #766, opened 2026-09-19
Bump the Tessera serving-runtime pin to the post-#560 lineage (contract v32) — GATED, do not merge before the tessera PRs land

RobTand:
Closes #760. Closes #699.

## What moves

The Tessera serving-runtime pin (`prismaquant.tessera_serving_runtime_pin.v2`)
moves from the 2026-09-15 pin to the post-#560 lineage, in the one reviewed
change the pin discipline requires — pin JSON, module constants, dev-pin
answer, legal-domain pins, reader widen, tests and docs together:

| | old (v29) | new (v32) |
|---|---|---|
| commit | `4c384e6049dca3eeaf503bb2c9cd1cd2778978d1` | `e4a3a7d4516de61785e3f8738ccb64b86cf9e8cf` |
| contract_sha256 | `db9ca4c0c457ee7105cf6c533c3c583cc00c9344584418bd5c052bce233299b3` | `14acb1f78b2da32272f077e4ae69cb3486846b03912ffa825b677568b84d1780` |
| contract_version | 29 | 32 |
| lane schema | v10 | v10 (unchanged) |
| version | 0.1.0 | 0.1.0 (`version_is_release` stays advisory/false) |

The pinned commit is the head of **tessera #562 (D2/D2b), stacked on #560**
(`flash/506-routed-full-domain-rates-20260918`), i.e. the lineage tip that
carries the *admission* facts; #563 (docs fix-forward, pin-coupled) edits
only the contract changelog and projects nothing any gate reads. Digest
bound by the one command against the canonical remote:
`git cat-file -p e4a3a7d45:src/tessera/serving/runtime_contract.json | sha256sum`.

## The v29 → v32 delta (the answer diff is the review)

| dimension | v29 | v32 |
|---|---|---|
| cells | 14 | 6 (8 withdrawn, master-side v30/v31: tessera #538/#551 + A4 retirement) |
| withdrawn | — | all 4 `TESSERA_BF16_K1` dense cells (both platforms — the only `recorded`-evidence dense cells and gfx1201's only cells), all 4 `TESSERA_E4M3_K1` dense cells (resident+streamed) |
| `device_qualified` cells | 14 | 6, all `sm_121`, all `requires_plugin: "tessera"`, all `backed_with_serve_flag` |
| routed_moe evidence | e4m3 pair `recorded` (q1024); e2m1 pair `not_recorded` (q896) | **byte-identical both** — no status flips |
[...]

PR #800, opened 2026-09-19
Stage A writes into the plan's sealed output_root (#799)

### LINKED ISSUE #799: Stage A payload passed the wrong --output-root and the receipt had no default
Stage A v4 (066e00a48c41) died at 10.8s: adjoint_identity_refused, plan output_root .../complete-512-seed237... is not --output-root .../joint-panel. The capture's identity guard demands the plan's sealed root and writes its receipt to <plan_root>/layer-quanta/adjoint/adjoint-capture.json. The dispatcher passed records-dir's parent instead, and its --adjoint-receipt defaulted to None, so even a successful capture would never unblock the quanta. Fix: _plan_output_root() reads the plan's output_root for the payload arg and for the receipt default.

RobTand:
Closes #799. Payload root + receipt default + fixture plan file; targeted 1/1 green (7 passed).


RobTand:
Merge rationale (GLM-5.3 coordinator, 2026-09-19): run 35453507497 on head 667ae606 fails 52 tests, all pre-existing on main and none in this diff's families — 12 mount-bound test_joint_layer_quanta (owned by open #804's skips) + 40 contract-census expectations (owned by open #766's lineage). This PR's own two families are green on CI for the first time: test_dispatch_joint_quanta passes via the autouse portable-spec fixture, and test_allocator_* stays green post-#802. Strict subset; merging per the #802 standard.

PR #804, opened 2026-09-19
Skip mount-dependent campaign tests where the sealed campaign is absent (#803)

RobTand:
Closes #803

Two commits, one per test file, applying the repo's established skip convention (test_glm_campaign_streaming.py:188: "skip -- never fail -- where neither exists (GitHub CI has no /mnt/shared)") to the tests that hard-read /mnt/shared:

- tests/test_joint_layer_quanta.py: 12 real-plan producer tests skip when the sealed hostcap32 plan is absent (landed with #785).
- tests/test_dispatch_joint_quanta.py: 6 submission-building tests skip when the pinned spec-hostcap32-ram-dev.json is absent (landed with the #782 dispatcher merges). test_stale_stage_a_receipt_refuses refuses before the read and needs no guard.

No production code touched; where the mount exists (PB shards, both Sparks) every test still runs in full. Acceptance: this PR's own CI run goes green on these 18 (the skip path validates on GitHub CI by construction), and PB confirms the full run where the mount exists.

Found while merging #785 (merger lane, same session — principle 14: fixer filing and fixing).

---

**Reworked by the coordinator (GLM-5.3, 2026-09-19):** the `tests/test_dispatch_joint_quanta.py` commit is dropped from this branch. Those six tests are synthetic-fixture tests, not sealed-data tests — #800's autouse portable-spec fixture makes them RUN on GitHub CI with a synthetic spec instead of skipping, which keeps the pinned submission shape exercised everywhere ("drift breaks placement"). Only the genuinely mount-bound `tests/test_joint_layer_quanta.py` skips (the real sealed hostcap32 plan cannot exist on CI) are kept. Land after #800.

PR #802, opened 2026-09-19
Allocator v5 fixtures: re-address packaged scopes across structures (#801)

RobTand:
Closes #801

## The break

Nine tests of the Tessera lane-admission family are red on main @ cbad7bd4f9, all failing with `require_producer_formats` refusing `TESSERA_E4M3_K1_R1024` out of `allocator.main()`. Root cause is fixture-only: since d0be123a2a (contract v23 / lane schema v10) cell coverage is structure-specific — the packaged dense cells publish `TESSERA_E2M1_K2` only and `TESSERA_E4M3_K1` only as routed_moe — and the two `_v5_contract` fixtures synthesized their routed-MoE population by relabelling the dense cells, which left the population with no E4M3 cell in either structure. Admission then (correctly) refused the rung the tests select.

## The fix

- `tests/conftest.py`: new `project_lane_cells_onto_structures(payload, structures)` beside `down_convert_lane_table`. It re-addresses the packaged (platform, family, regime) scopes onto exactly the requested structures: kept cells stay as published, every published cell is cloned into a requested structure its scope does not already cover, and cells of non-requested structures are dropped after cloning. A clone carries its family's own admission facts — rungs, activation contract, route status, plugin requirement, launch set — and only `structure`, a pure lookup discriminator (`cell_matches_serving_context`), is synthesized. Clones skip overlapping residencies under the same `TESSERA_SERVE_MODE` grammar `parse_v4_cell_contract` reads, keyed without the runtime axes because the calling fixtures then flatten every cell's runtime onto one fixture image.
- `tests/test_allocator_expert_projection.py::_v5_contract` and `tests/test_tessera_scope_endpoints.py::_v5_contract` build their population through the helper (both structures / dense-only for the `with_experts=False` variant, which keeps dense E4M3 coverage after dropping routed_moe cells).

No admission predicate, producer-eligibility rule, or reader behavior changes; the ship gates stay fail-closed and now see a fixture population coherent with the packaged table.

## Evidence (PrismaBuild, sparky, tessera pin 79ddd4c60 verified in each receipt)

| run | red (main) | green (this branch) |
|---|---|---|
| `test_allocator_expert_projection.py` + `test_allocator_tessera_priced_inputs.py` | 5 failed, 20 passed | **25 passed** |
| `test_tessera_scope_endpoints.py` + `test_tessera_stack_driver_integration.py` + `test_tessera_stack_population_bridge.py` | 4 failed, 29 passed, 1 skipped | **33 passed, 1 skipped** |

[...]

PR #807, opened 2026-09-19
Stage A threads the plan's historical_encoder_reuse (#806)

### LINKED ISSUE #806: Stage A never passed the plan's historical_encoder_reuse to the anchor loader
Stage A v6 (89d3a0eb012a) refused at 85 s — deep inside the container, past CUDA, mid-dispatch: the anchor checkpoint's recorded encoder_source_sha256 0833671b… is allowlisted in the plan (recorded 2026-09-17 with evidence), but the stage-A caller at joint_cost_stage_a.py:441 omitted the historical_encoder_reuse kwarg, so the loader saw no policy and refused a reuse the plan explicitly names. The single-run path threads it (tessera_joint_aura.py:2777). Fix: pass config.get('historical_encoder_reuse'); pinned by a seam test asserting the plan's block reaches the loader verbatim and a plan without it reaches None.

RobTand:
Closes #806. One-line threading fix mirroring the single-run path + seam pin; targeted 1/1 green (25 passed).


RobTand:
Merge rationale (GLM-5.3, 2026-09-19, corrected): run 35456203557's failing set is exactly the two pre-existing families — test_joint_layer_quanta (mount-bound, owned by open #804) + the 10-file contract-census block (owned by open #766) — byte-for-byte the same family list as #800's run 35453507497 on the same base. This diff touches prismaquant/joint_cost_stage_a.py and its seam test only; PB targeted 25/25 green. Strict subset per the #802 standard.

PR #820, opened 2026-09-19
Stage A: explicitly-recorded source_prefetch override beside the frozen plan (#819)

### LINKED ISSUE #819: Stage A needs an explicitly-recorded source_prefetch override beside the frozen plan (IO-side #809 pattern)
## The observation

The live distributed campaign's stage A (PB action `74f12044f7748404…`, claimed, still running at filing) is IO-starved, not GPU-bound: at layer ~24/45 the attempt log shows a few GPU batches and then minutes of halt, repeatedly. Its read path is the streaming model's source prefetch, whose budget the sealed plan pins at `source_prefetch = {prefetch_workers: 1, max_cache_slots: 2, prefetch_lookahead: 1}` — one worker, latency-bound on many-file collections, reading source extents over NFS-RDMA from the dl380-hosted tiers (the GB10s have no local NVMe stage). That is the #737 single-worker pin again, now on stage A, and it violates the repo's GPU-bound rule for exactly the reason `recommend_source_prefetch` (PQ #737) exists.

The plan cannot be re-sealed to fix it: this campaign's plan is frozen (the prepared completion binds its digest; a new plan means a 7+h re-prepare). The IO-side #809 pattern is the sanctioned shape for this: an **explicit input beside the sealed plan** — two sealed sources, the deviation recorded, never silent, no default.

## What changes

- `prismaquant/joint_cost_stage_a.py` gains `--prefetch-override <path-to-json>` (plus a `PRISMAQUANT_STAGE_A_PREFETCH_OVERRIDE` env spelling for direct invocations; CLI and env naming different files refuses, the #809 rule). The document carries a `reason` and a `source_prefetch` block validated by the plan's own field grammar (`_source_prefetch`'s completeness check — same six fields, same positivity/lookahead rules, residency still required).
- Given an override, the model build threads the override's budget **instead of** the plan's, for that run only. Plan bytes, plan/prepared digests, and the adjoint receipt are untouched — the quanta's bindings do not move.
- The deviation is stamped into the run's provenance, both files: `results.json` and `counters.json` carry `prefetch_override: {plan_sealed: …, run_used: …, reason, path, sha256, source}` (or `null` when no override ran), and the attempt log prints the same block at startup. An override without a non-empty `reason` refuses at load.
- `tools/dispatch_joint_quanta.py` gains `--stage-a-prefetch-override <path>`, threaded into `stage_a_argv` as the payload's `--prefetch-override` (the payload argv is the channel that crosses the container boundary — the launcher forwards no ambient action env), so the coordinator can dispatch a v10 stage A with `workers≈8 / slots≈7 / lookahead≈4` if the live v9 dies. The campaign state's `stage-a-submitted` event records the override path.

## Why this and not re-sealing or widening the default

The budget stays an explicit, per-run, recorded deviation until the next plan seal adopts measured numbers (`recommend_source_prefetch` exists for that seal). Nothing about the default changes: no override given → the plan's block verbatim, byte-identical behavior.

Filed by the subagent fixing this under Rob's direction ("running quite poorly, barely being fed, this should be parallelized"); the fix is mine to make — implementation on the branch, tests through PrismaBuild.

RobTand:
Closes #819.

The live campaign's stage A (PB `74f12044f774…`) is IO-starved: the plan's
sealed `source_prefetch` budget is #737's single-worker pin
(`{workers 1, slots 2, lookahead 1}`), and one worker is latency-bound on
many-file collections over NFS-RDMA from the dl380-hosted tiers — a few GPU
batches, then minutes of halt, at layer ~24/45. The plan is frozen for this
campaign (the prepared binds its digest; a re-seal is a 7+h re-prepare), so
this lands the IO-side #809 seam instead:

- **The contract.** `python3 -m prismaquant.joint_adjoint_capture
  --prefetch-override <path>` (or `PRISMAQUANT_STAGE_A_PREFETCH_OVERRIDE`
  for direct invocations; two explicit sources that disagree refuse) names a
  document `{schema, reason, source_prefetch}` whose budget block passes the
  plan's own completeness check — `_source_prefetch`'s exact field grammar:
  same six fields, same positivity/finite/lookahead rules, prefetched
  residency still required — plus a mandatory non-empty `reason`.
- **Semantics.** Given an override, the capture's model build threads the
  override's budget instead of the plan's, **for that run only**. Plan bytes,
  the plan/prepared digests and the adjoint receipt are untouched, so the
  quanta's receipt bindings do not move. No override given: the plan's block
  verbatim, byte-identical behavior.
- **Never silent.** `results.json` and `counters.json` carry
  `prefetch_override: {plan_sealed, run_used, reason, path, sha256, source}`
  (or `null`), and the attempt log prints the same block at startup.
- **Dispatcher.** `tools/dispatch_joint_quanta.py
  --stage-a-prefetch-override <path>` threads the payload's
  `--prefetch-override` — the channel that crosses the container boundary
  (`tessera_campaign_container` forwards no ambient action environment into
  the payload) — and records the path in the campaign state's
  `stage-a-submitted` event, so the deviation is the dispatcher's provenance
  too. Nothing else in the pinned argv moves.

Design doc: §13 addendum in `docs/design/distributed_campaign_2026-09-19.md`.

Tests (targeted, PrismaBuild, tag `sparklina`):
`test_joint_cost_quantum_runtime.py` — the override replaces the plan's
budget at the model-build seam and stamps the deviation into both provenance
files (a full stubbed run reads `results.json`/`counters.json` back from
disk); the document grammar refuses through the shared `_source_prefetch`
words; CLI/env disagreement refuses; the CLI flag threads (dropping it would
revert silently — the thing this exists to make impossible); no override =
plan verbatim + null stamp. `test_dispatch_joint_quanta.py` — the stage-A
argv is unchanged without the flag, carries the payload flag with it, and
`main` records it in the state event.

A v10 stage A can now be dispatched with
`workers≈8 / slots≈7 / lookahead≈4` if the live v9 dies, without touching
the sealed plan's bytes.


RobTand:
Merge rationale (GLM-5.3, 2026-09-19, fast standard per Rob): targeted PB receipts green (builder's seam pins: override threads, grammar-validated, deviation stamped); fast checks green (import surface, linked issue). The full-suite job runs as post-merge verification — main is green and this diff's families (stage-A seam, dispatcher argv, their tests) are covered by the targeted run. Strict-subset standard.

Issue #880, opened 2026-09-21
Stage A boundary readback requires PB produced-output staging (strict tier policy)


RobTand:
Live failure: 8ca8952cc651… died reading its own just-written boundary-0 entries (`staged-not-serving`, perturbed_x_cache:632) under strict tier policy. Root decision 2026-09-21: stage them through PB produced-output; NO own-session exemption. Plan: pq-stagea-produced-boundaries-plan.json (API traced at 8a682535c309; gaps GAP-1 retire/restage durable accounting, GAP-2 origin placement). This branch lands the integration gate only: failing-before ade921a851294 (adapter seam ImportError on pristine main; pinned-PB provenance green). Adapter + behavioural scenarios follow root's GAP rulings.

PR #881, opened 2026-09-21
Stage A produced-boundary chain: plan + integration gate (https://github.com/RobTand/prismaquant/issues/880)

RobTand:
## Stage A reads its own boundary entries, through PrismaBuild

An admitted Stage A action writes 512 exact boundary entries per boundary and
then cannot read them: the strict allowed-tier reader resolves the process
residency map, which is the map of the run's **sealed inputs**, and an output
this action just produced is not in it. `staged-not-serving`, correctly —
there is no own-session exemption from the physical tier policy.

The fix does not touch the input resolver. The entries go through
PrismaBuild's produced-output lifecycle, and the read is handed a
**supplemental, explicitly namespaced** resolver as an argument. A read-only
or foreign attached generation stays an ordinary input-map entry and cannot
declare an owner prewrite.

### The bounded cycle

Writer → `require_prewrite` before the first byte → publish deferred to the
first read → a real sealed mover claimed and run by a real fleet in another
process → a bounded wait on PrismaBuild's own receipt → composed reader
context → the **unchanged** strict prefetch → byte-equal tensors → window
exit **retires** the stage copy → the same unchanged references read again
through `produced_output.ensure_batch_materialized`, under a PB-sealed
successor mover, with the durable origin charge unchanged.

- The publication unit is the existing 64-entry read window. Per-entry movers
  are rejected on PB's own arithmetic: `stage_tokens_for_bytes` rounds **per
  mover** up to a whole GiB token, so 64 movers of ~16 MiB cost 64 tokens
  where the one group that already is the read window costs
  `ceil(group_bytes / GiB)`. The smaller last group is priced on its own range.
- A window owns its group identities from the moment it opens. Resolving them
  again at exit lost credits whenever entries were disposed **inside** a live
  window — the ordinary Stage A pattern. Demonstrated red before green.
- A refused retirement is a **result**: recorded with its reason, re-driven
  through the same PB retire, and after bounded attempts left standing with
  its credits retained. Never reported as retired, never refunded.
- A group's durable charge is reclaimed when its **last** origin file is gone
  **and** its stage copy is retired.
- Cleanup never masks the compute failure that caused it.

### Geometry is derived, never constant

`build_boundary_template` takes the **effective** artifact max (the plan's
sealed `max_artifact_bytes` or the run's override) as the durable origin
class maximum, and derives the tier window from the **actual maximum group**
rather than from the retained origin peak. No budget figure is written into
this code.

### Submit side

The deployed `pbrun --produced-output-template` **does** exist and is used:
it ingests the document as a declared input, seals its declaration into
request params, carries it onto the queue row and derives the tier window
demand from it. `tools/dispatch_joint_quanta.py` gains
`--stage-a-produced-output-template`, forwarded as that pbrun envelope option
(never a payload flag), and its refusal is now conditional on a `pbrun --help`
probe rather than unconditional. An earlier revision of this description said
the flag did not exist; that was true of an older runtime generation and is
wrong of the deployed one.

### Provenance

Every `prismabuild.*` module loads through `staged_lease.sdk_submodule`, from
the **same sealed generation** as the reader SDK. Proven, not asserted: the
test interpreter's installed distribution has no `produced_output` module at
all. The candidate is the immutable digest-pinned bundle at
`42f2cfb874077da66c27401e718d3636903a8a40`.

### Open, and deliberately not worked around

Produced-output egress can **decharge** a batch's stage token instead of
releasing it: a retirement whose staged entries are still shared with an
in-flight copy reports `tokens_released=0, tokens_decharged=1,
entries_deleted=0, shared_with=['in-flight-copy']`, and the token is then
neither free nor held by anyone, permanently. Reproduced 1 run in 6 at one
token; `tools/audit_produced_window_tokens.py` is the instrumentation. Root
traced the cause to PrismaBuild (an egress eviction that does not exclude its
own mover from the claimed-path check) and the repair is assigned outside
this lane. The rollover fixture therefore mints two tokens although
steady-state occupancy is one, and its **one-token acceptance is pending a
re-run against the repaired candidate** — it is not closed here.

### Evidence

All through published PrismaBuild at `--priority -10`,
`--threads-per-shard 1`, tag `x86` (dl380g10). No local pytest, no GPU work,
no whole-model run.

- Gate set 4/4 green: boundary chain 15, artifact budget 13, docs staleness 6,
  architecture doc 13.
- Adjacent callers: 8 files, one real failure reproduced before repair
  (`test_stage_a_prefetch_override_replaces_plan_budget_and_stamps`, an
  offline stub inheriting an admitted launch context), then 30 passed.
- `_produced_group_for` carries a before/after cProfile at the production
  panel's shape: **0.537 s → 19.4 µs per window**, and beside it 1.046×
  against an inline dict control, which is overhead versus an idealized
  control and not the delta. Narrow CPU metadata only.
- The live cycle has **not** completed and is barred until root's PB sealing
  fix deploys.


RobTand:
Plan revision 2 (Astra corrections applied; adapter coding halted): (1) keep-until-reverse retired — same-action retire+restage REQUIRED, gated on GAP-1 durable-accounting semantics (commit_batch over identical origins after retire_batch: double-charge? second mover publishable? spent-fence independence?); (2) origins STAY at RUN/layer-quanta/adjoint via template output_prefix containment (instance_dir is metadata; GAP-2 withdrawn); (3) produced fragments live under output_fragment_root + batch_namespace — the input resolver never sees them; owner_action_key/material namespace must be threaded explicitly at the prefetch seam; (4) template declared via the dispatcher's existing PB --produced-output-template flag pre-admission; no sealed-plan stanza; (5) 8 windows per boundary (512/64) with framing ceilings in tokens; (6) behavioral RED replaces the import gate: 24846ba97e51, 3 passed 1 skipped (strict staged-not-serving on pristine main; SDK acquire not exercised — refusal precedes acquire; tamper case is a length check, not an isolated digest proof); (7) timestamps now datetime.now(utc).

RobTand:
Plan revision 3 (Astra write-ordering constraint incorporated): boundary-0's full 512 entries are written BEFORE the first forward read, so current+next READ-window funding deadlocks at write time; the 4 GiB assumption is void. Batch stays 64 (8 batches/boundary). The live production frontier = the full boundary being read + the written-ahead overlap of the next boundary — provisional math from actual ceilings: ~18.09 MiB avg serialized entry, ~1.13 GiB per 64-entry batch, ~9.04 GiB per boundary, ~18.09 GiB peak staged frontier (framing ceilings to be added from actual writer receipts at template declaration). Open for PB/GAP-1 resolution: either the existing API separates DURABLE WRITE ADMISSION (commit_batch durable classes under durable_maxima) from ON-DEMAND STAGING (publish movers under the tier window) — as read today, commit_batch acquires tier tokens and transfers whole to the mover, so commit implies staged residency and the frontier bound governs — or the scoped working window covers precisely that frontier. Durable origin survives eviction; restage-for-reverse unchanged and required; plan/calibration/window count untouched; no new cache; adapter still halted.

RobTand:
Plan revision 4 (Astra native source correction): require_prewrite reserves DURABLE CLASS quota only -- physical funding happens at publish_prepaid_batch -> stage_output_intent/fund -> commit_batch. The separation exists: write all 512 boundary-0 origins under prewrite with zero tier tokens; demand materialization ONLY at the actual read frontier (immediate publish during the initial writes would stall at batch 3 under two-window credit). Peak staged becomes current+next read windows ~= 2.26 GiB + framing (revision-3's 18.09 GiB frontier governed only under the refuted commit-implies-staging reading). Remaining gaps for the native final contract audit: GAP-A logical final commit for never-materialized durable outputs; GAP-B same-origin repeat materialization after retire (new batch_id over the same origins; retired batch stays logically terminal; dup commit returns the terminal; spent fence permanently unfundable; durable-class double-charge arithmetic still to pin). No 400 GB staged, no window raise; adapter still halted.

RobTand:
Plan REWRITTEN (v2-final, not appended) around the accepted Astra direction: PB-origin-only commit + idempotent materialization requests with durable predecessor/sequence successor identity; distinct origin vs materialization accounting (origins charged once until proven deletion; retire frees only the copy; fences stay non-refundable); origins stay under RUN/layer-quanta/adjoint; fragments read via the EXISTING SDK owner_action_key/material-namespace surface. PQ owns only the adapter, the dispatcher-side versioned template declaration, and the prefetch namespace threading (that threading targets SDK surface present today and may proceed now); everything else waits for the PB executor's public API. Acceptance: the recorded behavioral RED (24846ba97e51), the same-origin double-materialization lifecycle regression, and the eight-batch/two-window frontier demonstration, all through PB -10.

RobTand:
Scope hold recorded (extension only; campaign not paused). Inventory returned in the plan artifact: PQ branch head 88fb0ab0ad0, behavioral RED 24846ba97e51, characterization 5a53022292c2 (limits labeled), and the first-publication PB public signature (validate_template/bind_instance/require_prewrite/commit_batch/publish_prepaid_batch/retire_batch/reclaim_origin/recover_batches + funding fences + the existing SDK owner/material namespace surface). Working hypothesis for root's capacity review: first-publication + hold-on-SSD through reverse may suffice (~416 GiB staged SSD per actual census, origins on the RUN volume, RAM window bounded current+next, no HDD waiver); extension growth frozen until root decides.

